PR #2, opened 2014-06-10
feature: add option arg raw for unescape_uri, get_uri_args

doujiang24:
do not decode '+' to ' ' when option raw is true

according to rfc1738, '+' is reserved characters

may be we can add the option arg, just like: urldecode, rawurldeocde in PHP :)


zhuizhuhaomeng:
If there is a real need，we will reopen this PR.

PR #3, opened 2015-02-19
optimize: resty.core.regex: change regex cache key to eliminate string concatenations

guanlan:


agentzh:
@guanlan Thanks for this! I'll have a look shortly.

PR #4, opened 2015-03-14
add no-padding option to base64 encoding

yangshuxin:


lib/resty/core/base64.lua
```diff
@@ -42,12 +43,25 @@ ngx.encode_base64 = function (s)
...
+        local dst = get_string_buf(dlen)
+        local r_dlen = C.ngx_http_lua_ffi_encode_base64(s, slen, dst, 1)
+        -- if dlen ~= r_dlen then ngx.say ("discrepancy in len") end
+        return ffi_string(dst, r_dlen)
+    else
+        local dlen = base64_encoded_length(slen, false)
+        local dst = get_string_buf(dlen)
+        local r_dlen = C.ngx_http_lua_ffi_encode_base64(s, slen, dst, 0)
+        -- if dlen ~= r_dlen then ngx.say ("discrepancy in len") end
+        return ffi_string(dst, r_dlen)
```
agentzh: I think we can avoid the code duplication here? I don't see why we need a branch here in the first place ;) We can just pass down the `no_padding` argument to those low level functions like `base64_encoded_length` and `C.ngx_http_lua_ffi_encode_base64`, right?

t/encode-base64.t
```diff
@@ -178,3 +178,51 @@ qr/\[TRACE   \d+ content_by_lua\(nginx\.conf:\d+\):3 loop\]/
...
+My4xNA
+--- error_log eval
+qr/\[TRACE   \d+ content_by_lua\(nginx\.conf:\d+\):3 loop\]/
+--- no_error_log
+[error]
+ -- NYI:
+
+
+
+=== TEST 8: set base64 (number) without padding (explictly specified)
```
agentzh: This test case tests the padding case. The test title is wrong it seems?

agentzh:
Applied a slightly modified version of your patch. Thanks for your work!

PR #5, opened 2015-10-21
a simple implementation of semaphore in ngx_lua from kugou

cuiweixie:
implement basical sempahore operation wait and post,using luajit ffi,and c.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -25,6 +25,7 @@ Table of Contents
     * [resty.core.misc](#restycoremisc)
     * [resty.core.time](#restycoretime)
     * [resty.core.worker](#restycoreworker)
+    * [resty.core.semaphore](#restycoresemaphore)
 * [Caveat](#caveat)
 * [TODO](#todo)
 * [Author](#author)
@@ -182,6 +183,12 @@ API Implemented
 
 * [ngx.worker.exiting](https://github.com/openresty/lua-nginx-module#ngxworkerexiting)
 * [ngx.worker.pid](https://github.com/openresty/lua-nginx-module#ngxworkerpid)
+
+[Back to TOC](#table-of-contents)
+
+* [ngx.semaphore.new](https://github.com/openresty/lua-nginx-module#ngxsemaphorenew)
+* [ngx.semaphore.wait](https://github.com/openresty/lua-nginx-module#ngxsemaphorewait)
+* [ngx.semaphore.post](https://github.com/openresty/lua-nginx-module#ngxsemaphorepost)
 
 [Back to TOC](#table-of-contents)
 
--- a/README.markdown
+++ b/README.markdown
@@ -25,7 +25,7 @@ Table of Contents
     * [resty.core.misc](#restycoremisc)
     * [resty.core.time](#restycoretime)
     * [resty.core.worker](#restycoreworker)
-    * [resty.core.semaphore](#restycoresemaphore)
+    * [ngx.semaphore](./lib/ngx/semaphore.md)
 * [Caveat](#caveat)
 * [TODO](#todo)
 * [Author](#author)
@@ -183,12 +183,6 @@ API Implemented
 
 * [ngx.worker.exiting](https://github.com/openresty/lua-nginx-module#ngxworkerexiting)
 * [ngx.worker.pid](https://github.com/openresty/lua-nginx-module#ngxworkerpid)
-
-[Back to TOC](#table-of-contents)
-
-* [ngx.semaphore.new](https://github.com/openresty/lua-nginx-module#ngxsemaphorenew)
-* [ngx.semaphore.wait](https://github.com/openresty/lua-nginx-module#ngxsemaphorewait)
-* [ngx.semaphore.post](https://github.com/openresty/lua-nginx-module#ngxsemaphorepost)
 
 [Back to TOC](#table-of-contents)
 
--- a/lib/ngx/semaphore.lua
+++ b/lib/ngx/semaphore.lua
@@ -1,4 +1,7 @@
-
+-- Copyright (C) Yichun Zhang (agentzh)
+-- Copyright (C) cuiweixie
+-- I hereby assign copyright in this code to the lua-nginx-module project,
+-- to be licensed under the same terms as the rest of the code
 local ffi = require 'ffi'
 local base = require "resty.core.base"
 
--- a/lib/ngx/semaphore.lua
+++ b/lib/ngx/semaphore.lua
@@ -32,23 +32,29 @@ if not pcall(ffi.typeof,"ngx_http_lua_semaphore_t") then
 end
 
 ffi.cdef[[
-    int ngx_http_lua_ffi_semaphore_new(ngx_http_request_t *r, 
+    int ngx_http_lua_ffi_semaphore_new(ngx_http_request_t *r,
         ngx_http_lua_semaphore_t **psem, int n, char *errstr, size_t *errlen);
+
     int ngx_http_lua_ffi_semaphore_post(ngx_http_lua_semaphore_t *sem,
         int n, char **errstr);
+
     int ngx_http_lua_ffi_semaphore_count(ngx_http_lua_semaphore_t *sem);
+
     int ngx_http_lua_ffi_semaphore_wait(ngx_http_request_t *r,
         ngx_http_lua_semaphore_t *sem, int wait_ms, char *errstr, size_t *errlen);
+
     void ngx_http_lua_ffi_semaphore_gc(ngx_http_lua_semaphore_t *sem);
 ]]
 
 
-local _M = {}
-
 local psem = ffi_new("ngx_http_lua_semaphore_t *[1]")
 
-function _M.new(n)
 
+local _M = {}
+local mt = { __index = _M }
+
+
+function _M.new(n)
     local r = getfenv(0).__ngx_req
     if not r then
         return nil, "no request found"
@@ -68,7 +74,7 @@ function _M.new(n)
 
     ffi_gc(sem, C.ngx_http_lua_ffi_semaphore_gc)
 
-    return setmetatable({ sem = sem }, _M)
+    return setmetatable({ sem = sem }, mt)
 end
 
 
@@ -131,6 +137,7 @@ function _M.post(self, n)
     return true
 end
 
+
 function _M.count(self)
     if type(self) ~= "table" or type(self.sem) ~= "cdata" then
         return nil, "semaphore not inited"
@@ -139,10 +146,8 @@ function _M.count(self)
     local cdata_sem = self.sem
     local ret = C.ngx_http_lua_ffi_semaphore_count(cdata_sem)
 
-    return ret      
+    return ret
 end
-
-_M.__index = _M
 
 
 return _M
--- a/lib/ngx/semaphore.lua
+++ b/lib/ngx/semaphore.lua
@@ -32,18 +32,18 @@ if not pcall(ffi.typeof, "ngx_http_lua_semaphore_t") then
 end
 
 ffi.cdef[[
-    int ngx_http_lua_ffi_semaphore_new(ngx_http_request_t *r,
-        ngx_http_lua_semaphore_t **psem, int n, char *errstr, size_t *errlen);
+int ngx_http_lua_ffi_semaphore_new(ngx_http_request_t *r,
+    ngx_http_lua_semaphore_t **psem, int n, char *errstr, size_t *errlen);
 
-    int ngx_http_lua_ffi_semaphore_post(ngx_http_lua_semaphore_t *sem,
-        int n, char **errstr);
+int ngx_http_lua_ffi_semaphore_post(ngx_http_lua_semaphore_t *sem,
+    int n, char **errstr);
 
-    int ngx_http_lua_ffi_semaphore_count(ngx_http_lua_semaphore_t *sem);
+int ngx_http_lua_ffi_semaphore_count(ngx_http_lua_semaphore_t *sem);
 
-    int ngx_http_lua_ffi_semaphore_wait(ngx_http_request_t *r,
-        ngx_http_lua_semaphore_t *sem, int wait_ms, char *errstr, size_t *errlen);
+int ngx_http_lua_ffi_semaphore_wait(ngx_http_request_t *r,
+    ngx_http_lua_semaphore_t *sem, int wait_ms, char *errstr, size_t *errlen);
 
-    void ngx_http_lua_ffi_semaphore_gc(ngx_http_lua_semaphore_t *sem);
+void ngx_http_lua_ffi_semaphore_gc(ngx_http_lua_semaphore_t *sem);
 ]]
 
 
--- a/lib/ngx/semaphore.lua
+++ b/lib/ngx/semaphore.lua
@@ -32,10 +32,11 @@ if not pcall(ffi.typeof,"ngx_http_lua_semaphore_t") then
 end
 
 ffi.cdef[[
-    int ngx_http_lua_ffi_semaphore_new(ngx_http_lua_semaphore_t **psem,
-        int n, char **errstr);
+    int ngx_http_lua_ffi_semaphore_new(ngx_http_request_t *r, 
+        ngx_http_lua_semaphore_t **psem, int n, char *errstr, size_t *errlen);
     int ngx_http_lua_ffi_semaphore_post(ngx_http_lua_semaphore_t *sem,
         int n, char **errstr);
+    int ngx_http_lua_ffi_semaphore_count(ngx_http_lua_semaphore_t *sem);
     int ngx_http_lua_ffi_semaphore_wait(ngx_http_request_t *r,
         ngx_http_lua_semaphore_t *sem, int wait_ms, char *errstr, size_t *errlen);
     void ngx_http_lua_ffi_semaphore_gc(ngx_http_lua_semaphore_t *sem);
@@ -44,14 +45,23 @@ ffi.cdef[[
 
 local _M = {}
 
+local psem = ffi_new("ngx_http_lua_semaphore_t *[1]")
 
 function _M.new(n)
-    local psem = ffi_new("ngx_http_lua_semaphore_t *[1]")
 
-    local ret = C.ngx_http_lua_ffi_semaphore_new(psem, n, errmsg)
+    local r = getfenv(0).__ngx_req
+    if not r then
+        return nil, "no request found"
+    end
+
+    local err = get_string_buf(ERR_BUF_SIZE)
+    local errlen = get_size_ptr()
+    errlen[0] = ERR_BUF_SIZE
+
+    local ret = C.ngx_http_lua_ffi_semaphore_new(r, psem, n, err, errlen)
 
     if ret == FFI_ERROR then
-        return nil, ffi_str(errmsg[0])
+        return nil, ffi_str(err, errlen[0])
     end
 
     local sem = psem[0]
@@ -102,9 +112,13 @@ function _M.wait(self, time)
 end
 
 
-function _M.post(self)
+function _M.post(self, n)
     if type(self) ~= "table" or type(self.sem) ~= "cdata" then
         return nil, "semaphore not inited"
+    end
+
+    if not n then
+        n = 1
     end
 
     local cdata_sem = self.sem
@@ -117,6 +131,16 @@ function _M.post(self)
     return true
 end
 
+function _M.count(self)
+    if type(self) ~= "table" or type(self.sem) ~= "cdata" then
+        return nil, "semaphore not inited"
+    end
+
+    local cdata_sem = self.sem
+    local ret = C.ngx_http_lua_ffi_semaphore_count(cdata_sem)
+
+    return ret      
+end
 
 _M.__index = _M
 
--- a/lib/ngx/semaphore.lua
+++ b/lib/ngx/semaphore.lua
@@ -32,11 +32,10 @@ if not pcall(ffi.typeof, "ngx_http_lua_semaphore_t") then
 end
 
 ffi.cdef[[
-int ngx_http_lua_ffi_semaphore_new(ngx_http_request_t *r,
-    ngx_http_lua_semaphore_t **psem, int n, char *errstr, size_t *errlen);
+int ngx_http_lua_ffi_semaphore_new(ngx_http_lua_semaphore_t **psem,
+    int n, char *errstr, size_t *errlen);
 
-int ngx_http_lua_ffi_semaphore_post(ngx_http_lua_semaphore_t *sem,
-    int n, char **errstr);
+int ngx_http_lua_ffi_semaphore_post(ngx_http_lua_semaphore_t *sem, int n);
 
 int ngx_http_lua_ffi_semaphore_count(ngx_http_lua_semaphore_t *sem);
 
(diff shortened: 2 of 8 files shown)


lib/resty/core/semaphore.lua
```diff
@@ -0,0 +1,116 @@
...
+
+
+function _M.wait(sem, time)
+    if sem == nil then
+        return nil, "param is nil"
+    end
+
+    time = tonumber(time) or 0
+
+    local cdata_sem  = sem.sem
```
agentzh: Style: 2 spaces before the `=` operator?

lib/resty/core/semaphore.lua
```diff
@@ -0,0 +1,116 @@
+
```
agentzh: I suggest using the `ngx.semaphore` module name instead of `resty.core.semaphore`.

t/semaphore.t
```diff
@@ -0,0 +1,1431 @@
...
+          { "/sub_sem_post"},
+        }
+        ngx.say(res1.status)
+        ngx.say(res1.body)
+        ngx.say(res2.status)
+        ngx.say(res2.body)
+        ';
+    }
+     location /sub_sem_wait {
+        proxy_pass $scheme://127.0.0.1:$server_port/sem_wait;
```
agentzh: It's fine to use subrequests but not fine with proxy_pass and backend HTTP communication. This test case would fail when multiple nginx workers are configured (which is allowed in this test scaffold) since semaphores are a per-worker thing.

t/semaphore.t
```diff
@@ -0,0 +1,1431 @@
...
+
+
+=== TEST 2: basic semaphore wait post
+--- http_config eval: $::HttpConfig
+--- config
+    location /test {
+        content_by_lua '
+            require "resty.core.semaphore"
+            local sem  = ngx.semaphore.new(0)
+            local ok,err=sem:wait(1)
```
agentzh: Style: please be careful about spaces. We need exactly one space before and after infix operators :)

t/semaphore.t
```diff
@@ -0,0 +1,1431 @@
...
+       ';
+    }
+--- no_check_leak
+--- request
+GET /test
+--- response_body
+API disabled in the context of init_worker_by_lua*
+--- no_error_log
+[error]
+
```
agentzh: Style: please use the [reindex](https://raw.githubusercontent.com/agentzh/old-openresty/master/bin/reindex) script to reformat your `.t` files so that it's consistently formatted. Thanks.

t/semaphore.t
cuiweixie: the worker is limit to 1

agentzh: @cuiweixie I know the default worker setting is 1. What I'm saying is that this test case is making things unnecessarily confusing and complex. Please keep in mind some advanced people check out test cases for more detailed documentation. We definitely should not encourage strange usage like this.

I think you can just use (0-delay) timers for this testing purpose where timer handlers have their own request contexts. It's also simpler than subrequests. Being simpler means being easier to debug things when tests fail. Even if you want to test subrequests in particular, you don't have to use `proxy_pass` to do HTTP communication in between.

cuiweixie: I think the way using proxy_pass is best, since is not allowed here, i will change to subrequest. The timer way donnot has two irrelevant request. As far as I concerned, simulating the scene of semaphore being used is important.

README.markdown
```diff
@@ -25,6 +25,7 @@ Table of Contents
     * [resty.core.misc](#restycoremisc)
     * [resty.core.time](#restycoretime)
     * [resty.core.worker](#restycoreworker)
+    * [resty.core.semaphore](#restycoresemaphore)
```
agentzh: This does not make sense since we do not have a `resty.core.semaphore` module any more. Now it's `ngx.semaphore`.

lib/ngx/semaphore.lua
```diff
@@ -0,0 +1,124 @@
...
+local get_size_ptr = base.get_size_ptr
+local setmetatable = setmetatable
+local co_yield = coroutine._yield
+local ERR_BUF_SIZE = 128
+
+
+local errmsg = base.get_errmsg_ptr()
+
+
+if not pcall(ffi.typeof,"ngx_http_lua_semaphore_t") then
```
agentzh: Style: a space needed after `,`.

agentzh:
The title of this PR has a grammatical mistake: "a simple implement of ..." -> "a simple implementation of ...".

lib/ngx/semaphore.lua
```diff
@@ -0,0 +1,124 @@
...
+        ngx_http_lua_semaphore_t *sem, int wait_ms, char *errstr, size_t *errlen);
+    void ngx_http_lua_ffi_semaphore_gc(ngx_http_lua_semaphore_t *sem);
+]]
+
+
+local _M = {}
+
+
+function _M.new(n)
+    local psem = ffi_new("ngx_http_lua_semaphore_t *[1]")
```
agentzh: This `psem` cdata should be lifted to the toplevel scope so that we don't have to allocate and free it on hot code paths. This is similar to the existing `errmsg` variable in the toplevel scope.

lib/ngx/semaphore.lua
```diff
@@ -0,0 +1,124 @@
...
+
+    if ret == FFI_DECLINED then
+        return nil, "busy"
+    end
+
+    return co_yield()
+end
+
+
+function _M.post(self)
```
agentzh: This Lua API is less capable than the underlying C API. How about directly exposing the `n` argument to the caller? It can be faster in cases where the caller has a known number of resources to post.

t/semaphore.t
```diff
@@ -0,0 +1,1068 @@
+# vim:set ft= ts=4 sw=4 et fdm=marker:
+use lib 'lib';
+use Test::Nginx::Socket::Lua;
+use Cwd qw(cwd);
+
+#worker_connections(10140);
+master_process_enabled(1);
```
agentzh: Any particular reason to enable the master process in this test file?

t/semaphore.t
```diff
@@ -0,0 +1,1068 @@
...
+    lua_package_path "$pwd/lib/?.lua;;";
+    init_by_lua '
+            local g = _G
+            local semaphore = require "ngx.semaphore"
+            local sem, err = semaphore.new(0)
+            g.err = err
+    ';
+_EOC_
+
+our $HttpConfigIntWorkerBy = <<_EOC_;
```
agentzh: I think `IntWorker` should be `InitWorker`. Typo?

t/semaphore.t
```diff
@@ -0,0 +1,1068 @@
...
+    lua_package_path "$pwd/lib/?.lua;;";
+    init_worker_by_lua '
+            local semaphore = require "ngx.semaphore"
```
agentzh: Style: bad indentation. Let's stick with 4-space indentation.

t/semaphore.t
```diff
@@ -0,0 +1,1068 @@
...
+
+#my $lua_default_lib = "/usr/local/openresty/lualib/?.lua";
+our $HttpConfig = <<_EOC_;
+    lua_package_path "$pwd/lib/?.lua;;";
+_EOC_
+
+our $HttpConfigInitByLua= <<_EOC_;
+    lua_package_path "$pwd/lib/?.lua;;";
+    init_by_lua '
+            local g = _G
```
agentzh: Style: let's stick with 4-space indentation.

t/semaphore.t
```diff
@@ -0,0 +1,1068 @@
...
+
+no_long_string();
+#no_diff();
+
+#my $lua_default_lib = "/usr/local/openresty/lualib/?.lua";
+our $HttpConfig = <<_EOC_;
+    lua_package_path "$pwd/lib/?.lua;;";
+_EOC_
+
+our $HttpConfigInitByLua= <<_EOC_;
```
agentzh: Style: we need a space before the binary operator `=`.

Also, this Perl package variable is used only once in this test file. So better inline the content of this Perl variable in the user test block directly.

t/semaphore.t
agentzh: This `$HttpConfigIntWorker` Perl package variable is referenced only once in this test file. Better inline its content in the test block directly.

t/semaphore.t
agentzh: Use of global variable environment table is discouraged (see https://github.com/openresty/lua-nginx-module#lua-variable-scope ). Better use proper Lua modules or `package.loaded` for this.

t/semaphore.t
```diff
@@ -0,0 +1,1068 @@
...
+run_tests();
+
+__DATA__
+
+=== TEST 1: basic semaphore wait post
+--- http_config eval: $::HttpConfig
+--- config
+    location = /test {
+        access_log off;
+        content_by_lua '
```
agentzh: Let's use the `*_by_lua_block` directives throughout these newly-added tests.

t/semaphore.t
```diff
@@ -0,0 +1,1068 @@
...
+        local res1, res2 = ngx.location.capture_multi{
```
agentzh: We need one more indentation level here.

t/semaphore.t
```diff
@@ -0,0 +1,1068 @@
...
+          { "/sem_wait"},
+          { "/sem_post"},
+        }
+        ngx.say(res1.status)
+        ngx.say(res1.body)
+        ngx.say(res2.status)
+        ngx.say(res2.body)
+        ';
+    }
+    location /sem_wait {
```
agentzh: Style: a blank line needed before this line.

t/semaphore.t
```diff
@@ -0,0 +1,1068 @@
...
+        content_by_lua '
+            local semaphore = require "ngx.semaphore"
+            local g = getmetatable(_G).__index
```
agentzh: This is too convoluted and hacky. Use of the `package.loaded` table should be much better here.

t/semaphore.t
```diff
@@ -0,0 +1,1068 @@
...
+--- request
+GET /test
+--- response_body
+false timeout
+--- no_error_log
+[error]
+
+
+
+=== TEST 3: basic semaphore wait not allow in body_filter_by_lua
```
agentzh: Grammar: "not allow in" -> "not allowed in". Please fix the test titles in subsequent test blocks as well.

t/semaphore.t
```diff
@@ -0,0 +1,1068 @@
...
+--- error_log
+API disabled in the context of header_filter_by_lua*
+
+
+
+=== TEST 5: basic semaphore wait not allow in init_worker_by_lua
+--- http_config eval: $::HttpConfigIntWorkerBy
+--- config
+    location /test {
+       content_by_lua '
```
agentzh: This complicated `content_by_lua` directive is useless for the test case. Better replace it with something trivial like `echo ok`.

t/semaphore.t
```diff
@@ -0,0 +1,1068 @@
...
+--- error_log
+API disabled in the context of log_by_lua*
+
+
+
+=== TEST 8: basic semaphore new not allow in init_by_lua
+--- http_config eval: $::HttpConfigInitByLua
+--- config
+    location /test {
+        content_by_lua '
```
agentzh: This `content_by_lua` directive is useless for this test block and should be replaced with something trivial like `echo ok`. This is a horrible distraction for the reader.

t/semaphore.t
```diff
@@ -0,0 +1,1068 @@
...
+GET /test
+--- response_body
+try magics
+--- error_log
+API disabled in the context of log_by_lua*
+
+
+
+=== TEST 8: basic semaphore new not allow in init_by_lua
+--- http_config eval: $::HttpConfigInitByLua
```
agentzh: Don't use a Perl variable reference here since its content is the most interesting thing for this test case. Don't hide it.

t/semaphore.t
```diff
@@ -0,0 +1,1068 @@
...
+--- config
+    location /test {
+        content_by_lua '
+            local g = getmetatable(_G).__index
+            ngx.say(g.err)
+        ';
+    }
+--- no_check_leak
```
agentzh: What's the reason behind disabling the memory leak check mode for this (and most of the subsequent) test block? Is it just a copy&paste legacy?

t/semaphore.t
```diff
@@ -0,0 +1,1068 @@
...
+--- config
+    location /test {
+       content_by_lua '
+            local semaphore = require "ngx.semaphore"
+            local sem = semaphore.new(0)
+            local g = getmetatable(_G).__index
+            local ok, err = sem:post()
+            g.test = sem
+            local function func(premature, g)
+                local sem1 = g.test
```
agentzh: This is unnecessarily convoluted. Let's just use the power of closure, upvalues, for the references?

t/semaphore.t
```diff
@@ -0,0 +1,1068 @@
...
+                local ok, err = sem1:wait(10)
+                if not ok then
+                    ngx.log(ngx.ERR, err)
+                end
+            end
+            ngx.timer.at(0,func,g)
+            ngx.sleep(2)
```
agentzh: I don't think we have to wait as long as 2 seconds in this test case. The running time of tests is important since the tests are supposed to run over and over again.

t/semaphore.t
```diff
@@ -0,0 +1,1068 @@
...
+[error]
+
+
+
+=== TEST 12: semaphore post in header_filter_by_lua
+--- http_config eval: $::HttpConfig
+--- config
+    location /test{
+        content_by_lua '
+            local res1,res2 = ngx.location.capture_multi{
```
agentzh: Style: a space needed after the comma.

t/semaphore.t
```diff
@@ -0,0 +1,1068 @@
...
+                    ngx.log(ngx.ERR, err)
+                end
+                g.test = sem
+            end
+            local sem = g.test
+            local ok, err = sem:post()
+            if not ok then
+                ngx.log(ngx.ERR, err)
+            end
+    ';
```
agentzh: Style: bad indentation for the subsequent lines. Be careful ;)

t/semaphore.t
```diff
@@ -0,0 +1,1068 @@
...
+
+
+=== TEST 13: test semaphore gc
+--- http_config eval: $::HttpConfig
+--- config
+    location /test {
+       content_by_lua '
+            local semaphore = require "ngx.semaphore"
+            local sem, err = semaphore.new(0)
+            local g = getmetatable(_G).__index
```
agentzh: We don't need any global mechanism for this particular test case. Let's just use locals.

t/semaphore.t
```diff
@@ -0,0 +1,1068 @@
...
+ok
+--- no_error_log
+[error]
+
+
+
+=== TEST 12: semaphore post in header_filter_by_lua
+--- http_config eval: $::HttpConfig
+--- config
+    location /test{
```
agentzh: Style: need a space before `{`. Alas.

t/semaphore.t
```diff
@@ -0,0 +1,1068 @@
...
+
+
+
+=== TEST 15: semaphore post in log_by_lua
+--- http_config eval: $::HttpConfig
+--- config
+        location /test {
+        log_by_lua '
+            local semaphore = require "ngx.semaphore"
+            local g = getmetatable(_G).__index
```
agentzh: Seems like we don't really need the global mechanism here.

t/semaphore.t
```diff
@@ -0,0 +1,1068 @@
...
+            ngx.exit(200)
+    ';
+    }
+--- no_check_leak
+--- request
+GET /test
+--- response_body
+post
+--- no_error_log
+[error]
```
agentzh: The checks in this test case is too weak. This case can pass even when the whole `log_by_lua` thing is not run at all. Let's add some outputs along the OK code paths as well and check them.

t/semaphore.t
```diff
@@ -0,0 +1,1068 @@
...
+                    ngx.log(ngx.ERR, err)
+                end
+                g.test = sem
+            end
+            local sem = g.test
+            local ok, err = sem:post()
+            if not ok then
+                ngx.log(ngx.ERR, err)
+            end
+    ';
```
agentzh: Bad indentation. Since you seem to be copying and pasting all over the place, better look over other test cases too.

t/semaphore.t
```diff
@@ -0,0 +1,1068 @@
...
+--- config
+    location /test {
+       content_by_lua '
+            local semaphore = require "ngx.semaphore"
+            local sem = semaphore.new(0)
+            if not sem then
+                error("create failed")
+            end
+            local function func(sem)
+                sem:wait(10)
```
agentzh: Add some output in this function to ensure this code path is actually run. The current checks are just too weak. Tests are code and can always have bugs. So better be strict and skeptical. The same applies to some other test cases in this test file.

agentzh: Add some output in this function to ensure this code path is actually run. The current checks are just too weak. Tests are code and can always have bugs. So better be strict and skeptical. The same applies to some other test cases in this test file.

t/semaphore.t
```diff
@@ -0,0 +1,1068 @@
...
+            end
+            local co = ngx.thread.spawn(func,sem)
+            ngx.thread.kill(co)
```
agentzh: This call can fail. Always check the return values. Such checks are also a way to test the test itself.

t/semaphore.t
```diff
@@ -0,0 +1,1068 @@
...
+--- request
+GET /test
+--- response_body
+test
+--- error_log
+ngx_http_lua_semaphore_cleanup
+
+
+
+=== TEST 19: a light thread that is going to exit is waitting a semaphore
```
agentzh: Typo: "waitting" -> "waiting"
Grammar: "is waiting a semaphore" -> "is waiting on a semaphore".

Please fix other places from copy&paste as well.

t/semaphore.t
```diff
@@ -0,0 +1,1068 @@
...
+
+=== TEST 20: test thread wait or post a sem in one request
+--- http_config eval: $::HttpConfig
+--- config
+    location = /test {
+        access_log off;
+        content_by_lua '
+            local semaphore = require "ngx.semaphore"
+
+            local g = getmetatable(_G).__index
```
agentzh: We don't need the global environment table here at all. There is just one request and just one Lua handler.

t/semaphore.t
```diff
@@ -0,0 +1,1068 @@
...
+                ngx.exit(500)
+            end
+            g.test_next = sem2
+
+            g.log = {}
+            g.log_i = 0
+            g.wait_id = 0
+            g.wakeup_id = 0
+            --0:wait 1:post 2:post_all
+            g.test_op = {0,1,1,0}
```
agentzh: Let's avoid such magical numbers. In Lua, strings are always intern'd and thus as cheap as "symbols" in other programming languages. So just use strings here to avoid confusions. Tests should be simple to stay away from bugs in themselves.

t/semaphore.t
```diff
@@ -0,0 +1,1068 @@
...
+            local function func(op,index,g)
+                local res
+                if op == 0 then
+                    local sem = g.test
+                    local sem2 = g.test_next
```
agentzh: Why not just use `sem` and `sem2` from the outer scope? We don't really need the `g.test` and `g.test_next` intermediates here at all, which just add confusion. Alas.

t/semaphore.t
```diff
@@ -0,0 +1,1068 @@
...
+                    sem2:post()
+                    g.wait_id = g.wait_id +1
+                    local wait_id = g.wait_id
+                    local ok, err = sem:wait(10)
+                    g.wakeup_id = g.wakeup_id + 1
+                    g.log_i = g.log_i + 1
+                    local wake_id = g.wakeup_id
+                    g.log[g.log_i] = {state=ok, err=err,op=0,wait_id = wait_id,wake_id = wake_id}
```
agentzh: Style: we need spaces around `=`.

t/semaphore.t
```diff
@@ -0,0 +1,1068 @@
...
+                    local sem = g.test
+                    local sem2 = g.test_next
+                    sem2:post()
+                    local ok, err = sem:post()
+                    g.log_i = g.log_i + 1
+                    g.log[g.log_i] = {state=ok, err=err,op=1}
+                end
+            end
+            local co_array = {}
+            for i=1,#g.test_op do
```
agentzh: Style: we need spaces around binary operators like `=`. We also need a space after `,`. Please fix other places with similar issues.

t/semaphore.t
```diff
@@ -0,0 +1,1068 @@
...
+wait
+true
+1
+1
+wait
+true
+2
+2
+--- no_error_log
+[error]
```
agentzh: Let's make the test block more intuitive. It's wrong to write test cases so hard to comprehend.

t/semaphore.t
```diff
@@ -0,0 +1,1068 @@
...
+
+    ';
+    }
+
+--- request
+GET /test
+--- response_body
+ok
+--- no_error_log
+[error]
```
agentzh: Again, we should check if the code paths have actually been exercised since tests may have bugs in themselves. We need much stronger checks here instead of just checking the "ok" output and any "[error]" messages.

t/semaphore.t
```diff
@@ -0,0 +1,1068 @@
...
+wait
+true
+2
+2
+--- no_error_log
+[error]
+
+
+
+=== TEST 21: benchmark
```
agentzh: What's the indentation of this block? Benchmarking? What exactly are you trying to benchmark? Let's be explicit here so that the reader (like me) does not have to guess ;)

t/semaphore.t
```diff
@@ -0,0 +1,1068 @@
...
+            local ok, err = sem:wait(1)
+            local g = _G
+            g.test = err
+    ';
+
+_EOC_
+
+run_tests();
+
+__DATA__
```
agentzh: Sorry if I miss it, but have you tested *multiple* light threads *actually* waiting on the same semaphore? Most of the test cases just do one wait and one post.

t/semaphore.t
```diff
@@ -0,0 +1,1068 @@
...
+            end
+       ';
+}
+--- no_check_leak
+--- request
+GET /test
+--- response_body
+busy
+--- no_error_log
+[error]
```
agentzh: Have you run code coverage tools like LCOV to evaluate the actual code coverage of this semaphore test suite?

agentzh: Seems like we need test cases creating and collecting many different semaphores so that we can actually exercise the logic in `semaphore_mm`.

t/semaphore.t
cuiweixie: send the err is nessesary!

agentzh: @cuiweixie Will you just emit the error to error.log? Such error relay is really confusing IMHO :)

cuiweixie: ok

t/semaphore.t
cuiweixie: i just try to avoid the benchmark.

t/semaphore.t
cuiweixie: I runed gco after you mention this. The rate of coverage in ngx_http_lua_semaphore.c  is 91%. Is it OK? I don't know much about covrage testing.

agentzh: @cuiweixie The absolute number is not bad but I suggest looking at the coverage details to ensure that all the important code paths and branches are indeed covered. Even when a critical path takes just 0.1%, it still should be covered by some tests :)

Oh BTW, "runed" -> "ran".

agentzh:
Just a quick note: this PR requires the following PR for lua-nginx-module to be fully functional:

https://github.com/openresty/lua-nginx-module/pull/584

agentzh:
@cuiweixie @doujiang24 Please reivew my local changes to your branch. If it's okay, I'll merge this PR along side my own changes to master. Thanks!

https://gist.github.com/agentzh/9819ce4a1bb2d4491b14

cuiweixie:
@agentzh, thank you! I think it is ok!

doujiang24:
@agentzh Thanks a lot! It's ok for me too.

agentzh:
A slightly modified version of your patch has just been applied to git master. Thank you for your hard work!

PR #6, opened 2015-11-27
Cleanups on unneeded code, adding locals.

bungle:
Yes, just a round of cleanups here as well, removing dead code and adding missing locals.


pushrax:
Curious why the code was left commented – just made a PR to delete these lines entirely if it's desired https://github.com/openresty/lua-resty-core/pull/14

bungle:
@pushrax,

ffi_new and setmetatable:
I just commented them out. No particular reason.

print and tonumber:
if you uncomment those debug lines, then you can also uncomment these two locals, that's why.

PR #7, opened 2015-12-25
shdict check param

membphis:
throw an error when the exptime less than 0


agentzh:
@membphis Applied to master with some minor modifications. Thanks! BTW, we don't have to add tests to lua-resty-core since ngx_lua's test suite can be run with lua-resty-core enabled :)

PR #9, opened 2015-12-28
Add missing locals to ngx/balancer.lua

bungle:
Added some missing locals.


membphis:
learn something new from this PR. the local variable is more faster.

agentzh:
@membphis Yes, it is.

PR #10, opened 2015-12-29
Incr method exptime

splitice:
Added expttime to incr, the behaviour of this pull request matches that of my pending pull request for lua core

### DIFF
--- a/lib/resty/core/shdict.lua
+++ b/lib/resty/core/shdict.lua
@@ -26,7 +26,7 @@ ffi.cdef[[
         int get_stale, int *is_stale);
 
     int ngx_http_lua_ffi_shdict_incr(void *zone, const unsigned char *key,
-        size_t key_len, double *value, char **err);
+        size_t key_len, double *value, int exptime, char **err);
 
     int ngx_http_lua_ffi_shdict_store(void *zone, int op,
         const unsigned char *key, size_t key_len, int value_type,
@@ -311,7 +311,7 @@ local function shdict_get_stale(zone, key)
 end
 
 
-local function shdict_incr(zone, key, value)
+local function shdict_incr(zone, key, value, exptime)
     zone = check_zone(zone)
 
     if key == nil then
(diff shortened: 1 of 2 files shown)


agentzh:
@splitice Thanks for your contribution! I'll look into this :)

doujiang24:
@splitice I think it better to let the second option arg to be `init` (the init value when the key is not exists), like: incr(key, init). You can take a look at: https://github.com/openresty/lua-nginx-module/pull/579

May be we can add `expire` command?
https://github.com/openresty/lua-nginx-module/pull/586

splitice:
option arg - I dont disagree that its possibly a good idea. I am however not sure if I could port that to the C version of the module. I'll wat for @agentzh's input first though.

expire command - personally I've found in lots of work that I have done the requirement to increment and "bump" up the expire time to be extremely common, especially in counters or limits.

An expire command is probably a good idea, I do however like it in the incr, it also makes handling possible race conditions easier :)

splitice:
Oh wait, thats a pull request. :+1: 

No code requirement from me :)

doujiang24:
2016-01-04 12:25 GMT+08:00 Mathew Heard <notifications@github.com>:

> option arg - I dont disagree that its possibly a good idea. I am however
> not sure if I could port that to the C version of the module. I'll wat for
> @agentzh <https://github.com/agentzh>'s input first though.
>
> expire command - personally I've found in lots of work that I have done
> the requirement to increment and "bump" up the expire time to be extremely
> common, especially in counters or limits.
>
Yeah, same as incr(key, `init`)

> An expire command is probably a good idea, I do however like it in the
> incr, it also makes handling possible race conditions easier :)
>
Yeah, I agree, it's much easier to set expire in incr.
If we have to choose one, I personally like incr(key, `init`) than
incr(key, `expt`).  ( I personally like redis style )
Or it can be incr(key, `init`, `expt`) or incr(key, `expt`, `init`) ?

agentzh:
I'm fine with `incr(key, value, init, expire)`. And I also want to have a separate `expire(time)` method :)

splitice:
personally if you went the init path I would prefer an options style array, at that point I feel the method is getting a bit to complex with optional fields, that also leaves room for more expansion. 

```incr(key,value,options:{expire:,init:})```?

agentzh:
@splitice I understand that concern. Options tables are generally more expensive than passed-by-position arguments though :)

splitice:
True, And I certainly agree since I use incr in some performance sensitive paths.

One other point, I would probably put init fourth. Popularity wise I would expect it to be less used.

agentzh:
@splitice Personally I use `init` much more often ;) But I agree it can be more consistent with the `set` method's parameter order.

splitice:
One more quick fix applied, the default value is inconsistent between resty core and the shared dict. Defaulted it to one for compatibility.

PR #12, opened 2016-01-04
feature: added new Lua modules ngx.ssl and ngx.ocsp for ssl_certificate_by_lua*

agentzh:
This pull request depends on the following lua-nginx-module pull request:

https://github.com/openresty/lua-nginx-module/pull/637

Documentation:

https://github.com/openresty/lua-resty-core/blob/ssl-cert-by-lua-2/lib/ngx/ssl.md

https://github.com/openresty/lua-resty-core/blob/ssl-cert-by-lua-2/lib/ngx/ocsp.md

Feedback welcome!


lziest:
Soon there would be OCSP 'must-staple'. One issue with our dynamic ocsp stapling is it is not ok to not sending ocsp response for the first time. Web admin needs to be aware of this. The program may need to recognize such certificates and keep a ocsp response around. A quick solution is to start a separate timer constantly refreshing ocsp responses for such certificates.

That is not a problem right now. But we could add some comments in the example code, warning that 'must-staple' certificate cannot be used this way.

agentzh:
@lziest I think the code example in the ngx.ocsp's documentation always sends an OCSP stapling even on the first request? See

https://github.com/openresty/lua-resty-core/blob/ssl-cert-by-lua-2/lib/ngx/ocsp.md#synopsis

lziest:
cool, I thought we still use the old sample. That's good! LGTM

agentzh:
Okay, I've just merged this PR into master. Further issues should go to separate tickets :)

PR #17, opened 2016-01-12
Fix 'ssl.server_name' documentation

rev112:
Looks like the syntax section wasn't modified after copy-pasting.


agentzh:
Nice catch. Thank you!

rev112:
giphy

PR #18, opened 2016-01-14
Adding missing link in table of contents for ngx.ssl

thefosk:
I noticed this new function, but had a hard time finding it in the doc because it was missing from the table of contents.


agentzh:
Thanks!

PR #19, opened 2016-01-22
Unmatched capture are set to false in capture table

jdesgats:
This PR reflect the changes made in lua-nginx-module by https://github.com/openresty/lua-nginx-module/pull/653.

Now unmatched capture groups in `ngx.re.match` and `ngx.re.gmatch` are
set to `false` rather than nil. This prevents to create table with
holes that don't play nice with `ipairs` or `#` operators.

For consistency, unmatched trailing captures (for instance in
`ngx.re.match("hello", "(hello)(.+)?")`) as well as named captures are
also set to `false`.


agentzh:
@jdesgats Applied a modified version of your patch to git master. Thanks for your contribution!

PR #21, opened 2016-01-28
add test case for balancer_by_lua_file with lua_code_cache off

doujiang24:
This is test case for https://github.com/openresty/lua-nginx-module/pull/663


agentzh:
@doujiang24 I've added a test case to the ngx_lua test suite. Better make the fix live with its test :)

Issue #22, opened 2016-02-01
Behavior limitation in ngx.re.(g)sub


p0pr0ck5:
I believe I found a limitation in the resty.core implemention of ngx.re.(g)sub. Examine the following (very contrived) example, without loading resty.core:

```
content_by_lua '
      local lookup = function(m)
          -- note we are returning a number type here
          return 5
      end

      local newstr, n, err = ngx.re.sub("hello, 1234", "([0-9])[0-9]", lookup, "oij")
      ngx.say(newstr)

';
```

This performs as expected:

```
$ curl localhost/re-test
hello, 55
```

However, when we enable resty.core, this fails with the following:

```
2016/02/01 08:49:52 [debug] 4659#0: pcre JIT compiling result: 1
2016/02/01 08:49:52 [debug] 4659#0: *1 [lua] content_by_lua(nginx.conf:271):3: replace(): m[0] is 12
2016/02/01 08:49:52 [debug] 4659#0: *1 lua resume returned 2
2016/02/01 08:49:52 [error] 4659#0: *1 lua entry thread aborted: runtime error: /usr/local/openresty/lualib/resty/core/regex.lua:641: attempt to get length of local 'bit' (a number value)
stack traceback:
coroutine 0:
	/usr/local/openresty/lualib/resty/core/regex.lua: in function 'gsub'
	content_by_lua(nginx.conf:271):7: in function <content_by_lua(nginx.conf:271):1>, client: 127.0.0.1, server: localhost, request: "GET /re-test HTTP/1.1", host: "localhost"
```

Obviously, we can see why in regex.lua this fails:

```
587 local function re_sub_func_helper(subj, regex, replace, opts, global)
[...snip...]
638         local res = collect_captures(compiled, rc, subj, flags)
639 
640         local bit = replace(res)
641         local bit_len = #bit
```

When `replace` returns a number value, we see our issue.

There are other locations where the `#` operator is called, that could lead a similar issue. For example:

```
347 local function re_match_helper(subj, regex, opts, ctx, want_caps, res, nth)
[...snip...]
374         rc = C.ngx_http_lua_ffi_exec_regex(compiled, flags, subj, #subj, pos)
```

[...]

PR #23, opened 2016-02-03
Caste various regex params to string to avoid thread abort

p0pr0ck5:
This commit updates regex.lua to force the use of string types
in situations where the '#' operator is called. This resolves
potential thread abort situations where non-string types are used
as parameters to ngx.re.(g)sub, ngx.re.(g)match and ngx.re.find
as noted in issue #22.

### DIFF
--- a/lib/resty/core/regex.lua
+++ b/lib/resty/core/regex.lua
@@ -353,6 +353,11 @@ end
 
 
 local function re_match_helper(subj, regex, opts, ctx, want_caps, res, nth)
+    -- we need to caste these to a string to avoid a potential thread abort
+    -- see http://bit.ly/1SH0oz6
+    regex = tostring(regex)
+    subj  = tostring(subj)
+
     local compiled, compile_once, flags = re_match_compile(regex, opts)
     if compiled == nil then
         -- compiled_once holds the error string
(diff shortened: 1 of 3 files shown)


lib/resty/core/regex.lua
```diff
@@ -353,6 +353,11 @@ end
 
 
 local function re_match_helper(subj, regex, opts, ctx, want_caps, res, nth)
+    -- we need to caste these to a string to avoid a potential thread abort
+    -- see http://bit.ly/1SH0oz6
+    regex = tostring(regex)
```
agentzh: I think regexes should always be of string type. I cannot think of real use cases of passing numbers or booleans as the pattern. These cases are almost always undesired. Thoughts?

agentzh:
@p0pr0ck5 I've applied a modified version of your patch to git master. Basically I think the cast on the regex parameter is superfluous. Let me know if you think otherwise. Thanks a lot!

p0pr0ck5:
I agree, I can't think of a valid use case. I added it in for the sake of completeness. Agreed that removing it as superfluous is acceptable. Glad it's in, looking forward to this version of resty core getting bundled. Thanks!

PR #25, opened 2016-02-04
Balancer by lua stream

splitice:
Commits in this branch corrospond to the changes for Balancer by LUA in the Stream Subsystem.

I have renambe the balancer.lua to balancer_http, and created the new implementation as balancer_stream.lua.

I will create a balancer.lua compatibility wrapper once the subsystem check is implemented.

Feel free to adjust the API to match your desired layout. 

https://github.com/openresty/stream-lua-nginx-module/pull/7


lib/ngx/balancer_stream.lua
```diff
@@ -0,0 +1,82 @@
+-- Copyright (C) Yichun Zhang (agentzh)
```
agentzh: I think `ngx.balancer.stream` is a better module name than `ngx.balancer_stream` :)

splitice: Or perhaps ngx.stream.balancer and ngx.http.balancer ?

That way the namespace expands with other APIs ported.

agentzh: Also, I hope we can still use the `ngx.balancer` module in the stream context. We could do a simple dispatch to `ngx.balancer.stream` in the top-level scope of `ngx.balancer`.

splitice: Yes, thats my intention. Once you, I or someone else implements the subsystem check.

agentzh: @splitice It's already too late to introduce the `ngx.http` and ngx.stream` namespace. Also it fragment the API namespace unnecessarily. The hope is that most Lua modules can work automatically in both contexts.

agentzh:
@splitice Just a side note: I think the `ngx.semaphore` API can be ported over to the stream subsystem in a similar way :)

lib/ngx/balancer_stream.lua
splitice: Well I'll let you lay this out this however you like. I'll defer to your experience.

splitice:
Currently I have no use for the semaphore module so wont be doing it myself at this stage.

I ported this because of the lack of variable support in the stream subsystem preventing me from doing something as simple as ```proxy_pass $map_variable;```. 

If you have any comments on any of the PR's I've made feel free to let me know today. I am spending the day porting over a couple more of our small patches to the stream subsystem so its not out of my way.

lib/resty/core/shdict.lua
```diff
@@ -26,7 +26,7 @@ ffi.cdef[[
         int get_stale, int *is_stale);
 
     int ngx_http_lua_ffi_shdict_incr(void *zone, const unsigned char *key,
-        size_t key_len, double *value, char **err);
+        size_t key_len, double *value, int exptime, char **err);
```
agentzh: I'm afraid these changes are irrelevant and should be removed out of this PR.

agentzh:
@splitice Will you add a test file for using `ngx.balancer` in the stream subsystem here?

splitice:
Probably best you do. I can do a theoretical port but dont have the time to get the environment passing.

splitice:
Also might be best to check if the balancer_by_lua_* tests all pass first :)

agentzh:
@splitice Please check out this tutorial for how to run the tests on your side:

It should be easy :)

t/shdict.t
```diff
@@ -915,3 +915,27 @@ qr/\[TRACE   \d+ content_by_lua\(nginx\.conf:\d+\):7 loop\]/
  -- NYI:
 stitch
 
+
+
+
+=== TEST 27: incr key expire
```
agentzh: This change is unrelated to this PR, I'm afraid.

PR #30, opened 2016-03-08
New Feature: added timeout support for balancer_by_lua*

toruneko:
the set_timeout API for balancer_by_lua*

Related: https://github.com/openresty/lua-nginx-module/pull/700

Related: https://groups.google.com/forum/#!topic/openresty/oFN8TfUUgq4/discussion


lib/ngx/balancer.lua
```diff
@@ -101,4 +104,23 @@ function _M.get_last_failure()
 end
 
 
+function _M.set_timeout(connect_timeout, send_timeout, read_timeout)
```
agentzh: Maybe it's more flexible to provide separate `set_xxx_timeout` methods? Sometimes we may just want to adjust a single timeout entry, for example.

toruneko: You right. In our business case, it's adjust read timeout only. 
Because of change timeout will allocating new memory for per request, so I write a method who can set a series variable of timeout. 
I will fix it when we resolve the memory problem.

t/balancer.t
```diff
@@ -595,3 +595,169 @@ ok
...
+--- config
+    location = /t {
+        proxy_pass http://backend;
+    }
+    location = /back {
+        echo "hello proxy timeout!";
+    }
+--- request
+    GET /t
+--- response_body:
```
agentzh: This is a syntax error. And you'll see the following error while running your `.t` file:

```
Extra lines not allowed in 'response_body' section at t/balancer.t
```

So you haven't run your new tests yourself?

t/balancer.t
```diff
@@ -595,3 +595,169 @@ ok
...
+        balancer_by_lua_block {
+            print("hello from balancer by lua!")
+            local b = require "ngx.balancer"
+            b.set_timeout(1000, 1000, 1000)
+            assert(b.set_current_peer("127.0.0.1", tonumber(ngx.var.server_port)))
+        }
+    }
+--- config
+    location = /t {
+        proxy_pass http://backend;
```
agentzh: This line is wrong. It should be `proxy_pass http://backend/back;`. Apparently you never run this test case yourself. Alas.

t/balancer.t
```diff
@@ -595,3 +595,169 @@ ok
...
+=== TEST 15: set connect timeout zero got 504
+--- http_config
+    lua_package_path "$TEST_NGINX_CWD/lib/?.lua;;";
+
+    upstream backend {
+        server 0.0.0.1;
+        balancer_by_lua_block {
+            print("hello from balancer by lua!")
+            local b = require "ngx.balancer"
+            b.set_timeout(0, 1000, 1000)
```
agentzh: Hmm, zero timeout values are never useful (except for testing). We should test nonzero values like 1 (ms) instead.

t/balancer.t
```diff
@@ -595,3 +595,169 @@ ok
...
+            ngx.log(ngx.ALERT, "unexpected success")
+        }
+    }
+
+--- request
+GET /t
+--- response_body
+ok
+--- error_log eval
+qr/\[error\] .*? log_by_lua.*? failed to call: API disabled in the current context/
```
agentzh: The expected error message is wrong. The error should be "failed to call: no upstream found" instead.

t/balancer.t
```diff
@@ -595,3 +595,169 @@ ok
...
+    }
+
+--- config
+    location = /t {
+        proxy_pass http://backend;
+    }
+--- request
+GET /t
+--- response_body_like: 502 Bad Gateway
+--- error_code: 502
```
agentzh: The status code should be 500 instead of 502.

agentzh:
Closing this in favor of #46.

toruneko:
agentzh, I am so sorry  that I did not know how to use cpan test, and modified existing test case but never running it. it's my fault，i should be more rigorous. sorry...

agentzh:
@toruneko Well, we have a full chapter in the "Programming OpenResty" book with detailed explanation on the testing toolchain:

Please let me know if you have any further questions.

PR #32, opened 2016-03-25
new test case: add test case for balancer module

xbsura:
add a test case for balaner module, as the set_current_peer may make ngx.var.upstream_addr not correct

### DIFF
--- a/t/balancer.t
+++ b/t/balancer.t
@@ -590,3 +590,44 @@ ok
 qr/\[error\] .*? log_by_lua.*? failed to call: API disabled in the current context/
 --- no_error_log
 [alert]
+
+
+=== TEST 14: test ngx.var.upstream_addr after using more than one set_current_peer
+--- wait: 0.2
+--- http_config
+    lua_package_path "$TEST_NGINX_CWD/lib/?.lua;;";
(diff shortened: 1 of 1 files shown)


t/balancer.t
```diff
@@ -590,3 +590,44 @@ ok
...
+            log_by_lua_block {
+                ngx.log(ngx.INFO, "ngx.var.upstream_addr is " .. ngx.var.upstream_addr)
+            }
+    }
+
+--- request
+GET /t
+--- response_body_like: 502 Bad Gateway
+--- error_code: 502
+--- error_log eval
```
agentzh: We do not need the `eval` filter here since the pattern is just a plain string.

t/balancer.t
```diff
@@ -590,3 +590,44 @@ ok
 qr/\[error\] .*? log_by_lua.*? failed to call: API disabled in the current context/
 --- no_error_log
 [alert]
+
```
agentzh: Will you use the `reindex` script to re-format the test file you have edited? Please see

https://github.com/agentzh/old-openresty/blob/master/bin/reindex

agentzh:
Applied with modifications. Thanks!

PR #33, opened 2016-04-02
Feature/worker id and count

membphis:


membphis:
syn the new api to resty-core library.

agentzh:
@membphis Applied your patch to git master with minor fixes. Thanks for your contribution!

PR #34, opened 2016-04-28
Support for passphrase protected key

guanglinlv:
read a private key from a BIO using the specify passphrase.

related [lua-nginx-module#750](https://github.com/openresty/lua-nginx-module/pull/750)


lib/ngx/ssl.lua
```diff
@@ -162,10 +163,11 @@ function _M.cert_pem_to_der(pem)
 end
 
 
-function _M.priv_key_pem_to_der(pem)
+function _M.priv_key_pem_to_der(pem, passphrase)
     local outbuf = get_string_buf(#pem)
 
-    local sz = C.ngx_http_lua_ffi_priv_key_pem_to_der(pem, #pem, outbuf, errmsg)
+    local sz = C.ngx_http_lua_ffi_priv_key_pem_to_der(pem, #pem, passphrase,
```
doujiang24: unnecessary space in the end

lib/ngx/ssl.lua
```diff
@@ -162,10 +163,11 @@ function _M.cert_pem_to_der(pem)
 end
 
 
-function _M.priv_key_pem_to_der(pem)
+function _M.priv_key_pem_to_der(pem, passphrase)
     local outbuf = get_string_buf(#pem)
 
-    local sz = C.ngx_http_lua_ffi_priv_key_pem_to_der(pem, #pem, outbuf, errmsg)
+    local sz = C.ngx_http_lua_ffi_priv_key_pem_to_der(pem, #pem, passphrase, 
+                                                        outbuf, errmsg)
```
doujiang24: `outbuf` should align with the `pem`

doujiang24:
@guanglinlv nice job!

agentzh:
Hmm, I don't think I like the API. We definitely don't want to add a passphrase argument to every Lua API functions involved with PEM-formatted private keys. I think you can just implement this as a 3rd-party Lua library that directly calls OpenSSL C functions via LuaJIT FFI and just strips the password from the private key provided? Frankly I don't see a reason to include this Lua API in the ngx_lua core :)

guanglinlv:
@agentzh , yeah,  i have to take care about some security reason in my case. all of the PEM-formated private key is protected by passphrase and can't strip out it  in my company. so, `priv_key_pem_to_der` can't work with that successfully. then i found it out and post the PR into [`lua-nginx-module`](https://github.com/openresty/lua-nginx-module/pull/750) and here.  for me or for the most of openresty users, we hope the official bundle will cover our case.  of course, i will try to fix the problem by implementing a 3rd-party library. thanks for your advice all  the same.

calind:
Hi guys and thanks for the contributions to these great projects.

I want to give you some arguments for including this patch in the main `ngx.ssl` module.

First, I think that passing a passphrase is better because avoids double call to `PEM_read_bio_PrivateKey` (one for encrypted key and one unwrapped key) which in turn goes from lua space to C space and back again. Of course the response can be cached and it's a micro-optimisation but for me it makes sense. Regarding the password argument, that can be optional, so it doesn't need to be carried all over the place.

Second, I think that only `strip_passphrase_in_priv_key` is too little for a module and personally I don't see a greater scoped module implementing this, but if you have one in mind we would gladly contribute to it.

zhuizhuhaomeng:
merged. @guanglinlv

PR #35, opened 2016-05-03
bugfix: we should prevent semaphore instance to be garbage collected when someone is waiting on it

doujiang24:


lib/ngx/semaphore.lua
```diff
@@ -110,7 +110,10 @@ function _M.wait(self, seconds)
         return nil, "timeout"
     end
 
-    return co_yield()
+    -- prevent the semaphore instance to be garbage collected
+    -- when someone is waiting on it
+    local ok, err = co_yield()
+    return ok, err
```
agentzh: `coroutine.yield()` never returns so we can just omit the `return` part to avoid triggering Lua's tail-call optimization (TCO).

agentzh: Okay, please ignore my previous comment. The return values of `coroutine.yield` still make a difference :)

agentzh:
Patch applied. Thanks a lot!

PR #40, opened 2016-05-14
skip ipv6 if not supported

chipitsine:
can you please merge it ? it the same as ipv6 magic for other tests


agentzh:
@chipitsine Applied. Thanks.

PR #44, opened 2016-05-31
Add jit_stack_size option

alubbe:
Complements https://github.com/openresty/lua-nginx-module/pull/782


lib/resty/core/regex.lua
```diff
@@ -162,6 +165,11 @@ local function get_max_regex_cache_size()
     return max_regex_cache_size
 end
 
+function _M.opt(option, value)
+  if option == "jit_stack_size" then
+    C.ngx_http_lua_ffi_set_jit_stack_size(value)
```
agentzh: Style: we use 4-space indentation in this project.

alubbe:
The function body now uses 4 spaces instead of 2

alubbe:
Regarding tests for resty-core, can we reuse the tests from lua-nginx-module's repo somehow or do you want me to copy them over to this PR?

agentzh:
@alubbe Yes, we can run ngx_lua's test suite using lua-resty-core by specifying the following environment beforehand:

```bash
export TEST_NGINX_INIT_BY_LUA="package.path = '$PWD/../lua-resty-core/lib/?.lua;$PWD/../lua-resty-lrucache/lib/?.lua;;' .. (package.path or '') require 'resty.core'"
```

assuming that lua-resty-core and lua-resty-lrucache are cloned in the same (parent) directory as ngx_lua.

alubbe:
Awesome, they pass locally.
So, this PR does not need additional tests, correct?
What is your opinion on the status of https://github.com/openresty/lua-nginx-module/pull/782?

agentzh:
@alubbe I think we still need some test cases to ensure that this new Lua API is indeed JIT compiled, as with the existing tests of this project.

lib/resty/core/regex.lua
```diff
@@ -135,6 +135,9 @@ ffi.cdef[[
                                              unsigned char *dst);
 
     uint32_t ngx_http_lua_ffi_max_regex_cache_size(void);
+
+    void ngx_http_lua_ffi_set_jit_stack_size(int size);
+
```
agentzh: Shall we remove this empty line?

lib/resty/core/regex.lua
```diff
@@ -162,6 +165,11 @@ local function get_max_regex_cache_size()
     return max_regex_cache_size
 end
```
agentzh: style: 2 blank lines are used to separate Lua function definitions.

lib/resty/core/regex.lua
```diff
@@ -162,6 +165,11 @@ local function get_max_regex_cache_size()
...
 
+function ngx.re.opt(option, value)
+    if option == "jit_stack_size" then
+        C.ngx_http_lua_ffi_set_jit_stack_size(value)
+    end
+end
```
agentzh: Ditto.

lib/resty/core/regex.lua
```diff
@@ -162,6 +165,11 @@ local function get_max_regex_cache_size()
     return max_regex_cache_size
 end
 
+function ngx.re.opt(option, value)
+    if option == "jit_stack_size" then
+        C.ngx_http_lua_ffi_set_jit_stack_size(value)
+    end
```
agentzh: I think we need to throw out an error when the option is unkown.

alubbe:
I've updated this PR to include the exact same functionality and error messages as the other PR.

agentzh:
@alubbe I think we still need a simple test to ensure LuaJIT's JIT compiler does not complain about any NYI when running this API function in a loop (see other tests for such examples). Thanks!

alubbe:
I've been thinking about this for a while, and, if possible, I'd simply like to re-run all find, match and sub tests with this option enabled. It would take care of NYI checks and other potential issues that we might have missed. Is there a simple way of doing this or do we need to duplicate every single test? If the latter, maybe we should add only one test per test file.

agentzh:
@alubbe I think you can just run typical tests. I don't mind duplications in the test suite since tricky tests often hide bugs of their own :) Simplicity takes preference in testing.

alubbe:
I have added two NYI tests. Presumely, they can't pass until https://github.com/openresty/lua-nginx-module/pull/782 is merged, right?

alubbe:
I've updated the travis file to use my other branch. The PCRE allocation fails, even though it works locally on my machine (I ran a minimal example, not the test suite). Could you help me figure out what's going on?

Also, my perl-foo was not strong enough to add a switch to use ngx.re.opt only for the last two tests. Or do you want to create a whole new file for these tests?

Beyond those two points, this PR should be finished - it's been working fine in production for several weeks now.

alubbe:
ping @agentzh 
if you're busy just let me know

t/re-find.t
```diff
@@ -261,3 +263,83 @@ NYI
...
 qr/\[TRACE   \d+ content_by_lua\(nginx\.conf:\d+\):4 loop\]/
 
+
+
+=== TEST 7: matched, no submatch, no jit compile, no regex cache, large jit_stack_size
+--- http_config eval: $::HttpConfig
+--- config
+    location = /re {
+        access_log off;
+        content_by_lua '
```
agentzh: New tests should always use the `*_by_lua_block {}` form :)

agentzh:
@alubbe I see the tests in Travis CI are failing. Maybe you should rebase your lua-resty-core and lua-nginx-module branches to the their latest masters?

alubbe:
I have just rebased, but the issues are still the same as described in https://github.com/openresty/lua-resty-core/pull/44#issuecomment-234528584 and https://github.com/openresty/lua-nginx-module/pull/782#issuecomment-260184185
Specifically, the `init_by_lua` handler fails in this test run (even though it works fine in 'real' code. If possible, I'd have it so that all test in that file run twice, with and without a changed jit_stack_size. But I cannot figure out how to do that.

agentzh:
@alubbe What I'm seeing in your lua-resty-core PR's travis CI report is actually a compilation error:

https://travis-ci.org/openresty/lua-resty-core/jobs/175504983

Please provide links to the travis CI report you're talking about. I'm not sure what you mean here.

agentzh:
@alubbe 

> If possible, I'd have it so that all test in that file run twice, with and without a changed jit_stack_size. But I cannot figure out how to do that.

Maybe you'll find the `repeat_each(N)` thing helpful? Though I'm not sure what exactly you mean here.

alubbe:
@agentzh only owners can rebuild, folk like me have to add ghost commits :)
Because I removed the ghost commit, the link on github is broken - you'll want to look at https://travis-ci.org/openresty/lua-resty-core/jobs/175508325

I would like to run all regex test as they are today, and then re-run them as they are, but with an additional ngx.re.opt. I could copy+paste all tests but that would be an additional maintenance burden when adding new regex tests. So what I am looking for is a way to repeat a test but to run a new init_ handler that sets a different jit stack size.

agentzh:
@alubbe There is a way by setting the `TEST_NGINX_INIT_BY_LUA` environment. I've been using it to run ngx_lua's test suite with lua-resty-core, as in

```bash
export TEST_NGINX_INIT_BY_LUA="debug.sethook(function () collectgarbage() end, 'l') jit.off() package.path = '$PWD/../lua-resty-core/lib/?.lua;$PWD/../lua-resty-lrucache/lib/?.lua;' .. (package.path or '') require 'resty.core' require('resty.core.base').set_string_buf_size(1) require('resty.core.regex').set_buf_grow_ratio(1)
```

Yeah, this example does much more than just using lua-resty-core though. But you can get the basic idea.

agentzh:
@alubbe Okay, I've just noticed that I had already mentioned the `TEST_NGINX_INIT_BY_LUA` trick earlier in this thread.

agentzh:
@alubbe Why am I seeing compilation errors in this PR's travis ci build report?

https://travis-ci.org/openresty/lua-resty-core/builds/175504982

lib/resty/core/regex.lua
```diff
@@ -154,6 +156,24 @@ local function get_max_regex_cache_size()
 end
 
 
+function ngx.re.opt(option, value)
+    if option == "jit_stack_size" then
+        local result = C.ngx_http_lua_set_jit_stack_size(value)
+
+        if result == -1 then
```
agentzh: I believe we also have named constants like `FFI_ERROR` and etc. Please avoid magical numbers here as well. Thanks!

lib/resty/core/regex.lua
```diff
@@ -154,6 +156,24 @@ local function get_max_regex_cache_size()
...
+            return error("Changing jit stack size is not allowed when " ..
+                         "regexs have already been compiled and cached")
+        elseif result == -2 then
```
agentzh: Ditto.

lib/resty/core/regex.lua
```diff
@@ -154,6 +156,24 @@ local function get_max_regex_cache_size()
 end
 
 
+function ngx.re.opt(option, value)
+    if option == "jit_stack_size" then
+        local result = C.ngx_http_lua_set_jit_stack_size(value)
+
```
agentzh: I think we should test good return value here first, since it's the most common case. Also, please use the `rc` variable name here instead of `result`.

t/re-find.t
```diff
@@ -32,6 +32,8 @@ our $HttpConfig = <<_EOC_;
         -- jit.opt.start("hotloop=1")
         -- jit.opt.start("loopunroll=1000000")
         -- jit.off()
+
+        ngx.re.opt("jit_stack_size", 128 * 1024)
```
agentzh: Let's avoid enabling this by default for our existing tests.

agentzh: I think we should add a new test file like `re-opt.t` for this new feature instead.

agentzh:
@alubbe This PR's new tests pass for me on my local linux.

agentzh:
@alubbe BTW, I think you can just enable Travis CI for your own forked repo.

lib/resty/core/regex.lua
```diff
@@ -126,6 +126,8 @@ ffi.cdef[[
                                              unsigned char *dst);
 
     uint32_t ngx_http_lua_ffi_max_regex_cache_size(void);
+
+    int ngx_http_lua_set_jit_stack_size(int size);
```
agentzh: This function name does not follow our public C API naming convention. We need the `ngx_http_lua_ffi_` prefix.

alubbe:
For test 2, how can I extend `HttpConfig` by adding another line?

agentzh:
@alubbe The eval filter is just treating the section value as Perl source code. So you can just use Perl variable interpolation like below:

--- http_config eval
qq!
$::HttpConfig
# a new line or new lines go here...
!

But for your `TEST 2` it won't work since you're duplicating `init_by_lua*` in the same scope which is disallowed by ngx_lua's config loader anyway.

Also, a duplicate `http_config` section just overrides the previous one. The right way is to avoid using `$::HttpConfig` altogether in that test case.

alubbe:
So I just hard-copy everything and add my one line?
Wouldn't that be hard to maintain / be surprising when you change `HttpConfig` but test 2 does not reflect your change?

alubbe:
Both PRs have now been separated, the Lua 5.1 API has been removed and the tests now live in a separate file in this PR.
@doujiang24 could you help me once more to get these tests working? My trial & error approach is failing me :(

agentzh:
I suggest you spend a little bit of your time to learn some very basic perl lexical syntax to avoid trials and errors altogether.

agentzh:
@alubbe Also try sitting down and spending some time understanding @doujiang24's previous changes for you. Try to understand why your version does not work while @doujiang24's works. Hey, it's not rocket science nor any kind of magic.

agentzh:
@alubbe Please just try to be a bit more patient :) It'll save a lot of your time and also our time :) Thanks!

alubbe:
@agentzh Please don't think me lazy - I was getting desparate. This is not only the first time I've written any C and Perl code, but also the first I've read any. I want to see this through, but it's extremely frustrating to work on something for hours that you know is simple and easy.
Anyway, I figured out that resty.core uses lua-level caching of the compiled regex, so the previous detection has to be overthrown.

Now, all my tests pass locally, but test 2 fails on travis. Is there anything different about the travis setup?

agentzh:
@alubbe No, I was not assuming you are lazy. Just giving your suggestions :)

Regarding travis ci, you can find all the detailed commands for Travis CI in the `.travis.yml` file of this git repo. Maybe it's because the travis ci's ubuntu's system pcre library is a bit old? Or maybe you are trying to allocate too big a stack which exceeds the travis ci VM's memory limit? I'm not sure though. Feel free to dig it up on your side :)

Thanks!

alubbe:
I have put together a docker file emulating the travis environment - I can now reproduce these errors.
Here is the valgrind output:
```
TEST 2: increase jit_stack_size
nginx: [error] init_by_lua error: init_by_lua:20: PCRE jit stack allocation failed
stack traceback:
	[C]: in function 'opt'
	init_by_lua:20: in main chunk
==77707== 1,024 bytes in 1 blocks are definitely lost in loss record 658 of 682
==77707==    at 0x4C2D110: memalign (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==77707==    by 0x4C2D227: posix_memalign (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==77707==    by 0x43BEB9: ngx_memalign (ngx_alloc.c:57)
==77707==    by 0x41AE80: ngx_create_pool (ngx_palloc.c:24)
==77707==    by 0x41910C: main (nginx.c:243)
==77707== 
{
   <insert_a_suppression_name_here>
   Memcheck:Leak
   match-leak-kinds: definite
   fun:memalign
   fun:posix_memalign
   fun:ngx_memalign
   fun:ngx_create_pool
   fun:main
}
```
Indeed, we can fix these errors by removing these two lines: https://github.com/openresty/lua-nginx-module/pull/782/files#diff-5a62a11d0a2e8e55e4c5a7feaac0edd1R1954 and https://github.com/openresty/lua-nginx-module/pull/782/files#diff-5a62a11d0a2e8e55e4c5a7feaac0edd1R1959

But that is probably not what we want - is it?

Even stranger is the fact that the regex compilation in test 1 does not fail, presumeably meaning that the regex was successfully compiled. This surprises me because I have copied over these tests from the lua5.1 api, where the compilation of test 1 fails and the setting of a new jit_stack_size in test 2 succeeds. Here, it is the reverse.

Is there a fundamental difference between resty.core and the 'old' api? Are there any changes to the memory structure?

Just as a reference, here is the CI output of these exact same 4 tests in lua5.1, all passing: https://travis-ci.org/openresty/lua-nginx-module/builds/184582058

alubbe:
I've also linked against PCRE-8.39 and tried setting the jit stack to small values like 16kb, but to no avail.

agentzh:
@alubbe The valgrind error message you gave above was the consequence of the allocation failure instead of the cause. So it is not really helpful.

I wonder if there is a bug in `ngx_http_lua_pcre_malloc_init` or `ngx_http_lua_pcre_malloc_done` or your usage of these 2 function. Not sure :)

alubbe:
I believe it might be the case that the travis build does not have access to pcre JIT at all. I can only reproduce the travis errors if I remove the "with-pcre-jit" flag. How could we check whether travis has it enabled?

alubbe:
Oh wow, for the first time in 7.5 months, both PRs are passing! 👍🎆 

It was indeed the pcre that travis comes with. I think this also means that this was the first time the resty-core test suite was run using pcre jit.

agentzh:
@alubbe Good catch. Fortunately, my local testing environment and my Amazon EC2 test cluster have both been running this project's test suite with PCRE JIT for years :) This Travis CI thing is relatively new for this project :)

agentzh:
@alubbe Thanks for the hard work!

@doujiang24 @membphis @moonbingbing @thibaultcha Do you guys think this PR and also its sister openresty/lua-nginx-module#782 are ready to merge? Many thanks!

.travis.yml
```diff
@@ -22,6 +22,7 @@ env:
     - LUAJIT_INC=$LUAJIT_PREFIX/include/luajit-2.1
     - LUA_INCLUDE_DIR=$LUAJIT_INC
     - LUA_CMODULE_DIR=/lib
+    - PCRE_VER=8.39
```
agentzh: Better use 8.33 to be consistent with lua-nginx-module.

lib/ngx/re.lua
```diff
@@ -193,4 +201,26 @@ function _M.split(subj, regex, opts, ctx, max, res)
 end
 
 
+function _M.opt(option, value)
+    if option == "jit_stack_size" then
+        if not core_regex.regex_cache_is_empty() then
```
agentzh: Maybe it's better to avoid this constant table look up on hot code paths?

lib/resty/core/regex.lua
```diff
@@ -153,6 +154,14 @@ local function get_max_regex_cache_size()
     return max_regex_cache_size
 end
 
+function _M.regex_cache_is_empty()
+    if not regex_match_cache then
+        return true
+    else
+        return next(regex_match_cache.hasht) == nil
```
agentzh: Hmm, IIRC, this `next()` builtin cannot be JIT compiled, unfortunately.

t/re-find.t
```diff
@@ -260,4 +260,3 @@ matched: 234
 NYI
 --- error_log eval
 qr/\[TRACE   \d+ content_by_lua\(nginx\.conf:\d+\):4 loop\]/
-
```
agentzh: Better avoid unrelated changes in this PR.

lib/resty/core/regex.lua
agentzh: Also, directly indexing internal fields of `resty.lrucache` instances break encapsulation and poses risk of future incompatibilities.

agentzh: @alubbe I think it's better to just keep the track of regex caching ourselves via a top-level local flag variable.

thibaultcha: > Hmm, IIRC, this next() builtin cannot be JIT compiled, unfortunately.

Indeed. Source: http://wiki.luajit.org/NYI

agentzh: @alubbe The simplest thing for us to do is to simply set a flag when any of those `ngx.re.*` API functions that may cache a new compiled regex are called.

lib/resty/core/regex.lua
```diff
@@ -153,6 +154,14 @@ local function get_max_regex_cache_size()
     return max_regex_cache_size
 end
 
+function _M.regex_cache_is_empty()
+    if not regex_match_cache then
+        return true
+    else
```
thibaultcha: Isn't one of this module's guidelines to avoid such branches when the previous one is returning? 
```lua
if not regex_match_cache then
    return true
end

return next(regex_match_cache.hasht) == nil
```

lib/ngx/re.lua
```diff
@@ -193,4 +201,26 @@ function _M.split(subj, regex, opts, ctx, max, res)
...
+            return error("Changing jit stack size is not allowed when some " ..
+                         "regexs have already been compiled and cached")
+        end
+
+        local rc = C.ngx_http_lua_ffi_set_jit_stack_size(value)
+
+        if rc == FFI_OK then
+            return
+        elseif rc == FFI_ERROR then
```
thibaultcha: ditto: previous branch returning (if we care)

lib/ngx/re.lua
```diff
@@ -193,4 +201,26 @@ function _M.split(subj, regex, opts, ctx, max, res)
...
+            return error("PCRE jit stack allocation failed")
+        end
+
+        return error("Received unexpected return code: " .. rc)
+    end
+
+    return error("unrecognized option name")
```
thibaultcha: A nitpick, but all of the other error messages are capitalized, but they are not in other modules of this library, nor are they in ngx_lua iirc.

Better to stay consistent in our API, and maybe un-capitalize the other errors thrown by this function?

agentzh: @thibaultcha Agreed. Should be un-capitalized in error messages.

lib/ngx/re.lua
agentzh: Agreed.

lib/resty/core/regex.lua
agentzh: @alubbe Okay, I think it's better just to set the flag in our own wrapper as `lrucache_set()`. Right now `lrucache_set()` is pointed directly to `lrucache.set` and we can turn it into a wrapper around `lrucache.set`.

lib/ngx/re.md
```diff
@@ -131,6 +135,37 @@ local res, err = ngx_re.split("a,b", ",", nil, nil, nil, my_table)
 When the trailing `nil` is not enough for your purpose, you should
 clear the table yourself before feeding it into the `split` function.
 
+ngx.re.opt
+----------
+**syntax:** *ngx.re.opt(option, value)*
```
agentzh: This looks wrong since we never invoke this function this way :)

lib/resty/core/regex.lua
```diff
@@ -24,6 +24,7 @@ local ngx = ngx
 local type = type
 local tostring = tostring
 local error = error
+local next = next
```
agentzh: Let's remove this.

lib/resty/core/regex.lua
```diff
@@ -154,6 +155,19 @@ local function get_max_regex_cache_size()
 end
 
 
+local regex_cache_is_empty = true
+
```
agentzh: Style: two blank lines needed here.

lib/resty/core/regex.lua
```diff
@@ -154,6 +155,19 @@ local function get_max_regex_cache_size()
...
+function _M.is_regex_cache_empty()
+    return regex_cache_is_empty
+end
+
+
+local function lrucache_set_wrapper(...)
+    _M.regex_cache_is_empty = false
```
agentzh: Let's avoid this table lookup in `_M`. We should directly reference the upvalue.

lib/resty/core/regex.lua
```diff
@@ -154,6 +155,19 @@ local function get_max_regex_cache_size()
 end
 
 
+local regex_cache_is_empty = true
+
+function _M.is_regex_cache_empty()
```
agentzh: We don't need this global function. Just reference the top-level local variable wherever needed.

lib/resty/core/regex.lua
```diff
@@ -154,6 +155,19 @@ local function get_max_regex_cache_size()
 end
 
 
+local regex_cache_is_empty = true
```
agentzh: I don't quite like this variable name. How about `regex_cache_updated`?

lib/resty/core/regex.lua
agentzh: Also, I think this is wrong. You insert a Lua function into `_M`, remember?

alubbe:
This PR now handles missing pcre jit support.
Is it possible to write a test for this? I.e., to run a test once with JIT enabled and once with it off?

agentzh:
@alubbe No, it's not easy to write a test for this while having a lot of existing tests specifically for JIT :)

alubbe:
Great, then we have only two open points: 
a) the variable name
b) the way in which this variable's value is accessed from other files
@agentzh I've left my thoughts on your feedback, so let me know what you think.

Once that's cleared up, I can do a final rebase.

agentzh:
@alubbe Looking good to me now. @doujiang24 and @thibaultcha ?

lib/ngx/re.md
```diff
@@ -131,6 +135,37 @@ local res, err = ngx_re.split("a,b", ",", nil, nil, nil, my_table)
 When the trailing `nil` is not enough for your purpose, you should
 clear the table yourself before feeding it into the `split` function.
 
+opt
+----------
```
thibaultcha: Seems only 3 dashes are necessary here

lib/ngx/re.md
```diff
@@ -131,6 +135,37 @@ local res, err = ngx_re.split("a,b", ",", nil, nil, nil, my_table)
...
+**syntax:** *ngx_re.opt(option, value)*
+
+Allows changing of regex settings. Currently, it can only change the
+`jit_stack_size` of the PCRE engine. For example,
```
thibaultcha: nitpick: I believe a comma here is strange. Most of the documentation in OR will either state:
```
[...] like below
'''lua
-- some example
'''
```

Or:
```
For example:
'''lua
-- some example
'''
```

This is honestly more of a nitpick, not sure if relevant, feel free to ignore.

lib/ngx/re.md
```diff
@@ -131,6 +135,37 @@ local res, err = ngx_re.split("a,b", ",", nil, nil, nil, my_table)
...
+             -- instead, we get a match
+             ngx.print(from .. "-" .. to) -- prints '1-1563'
+         }
+     }
+ }
+```
+
+This method requires the PCRE library enabled in Nginx.
+
+This feature was first introduced in the `XXX` release.
```
thibaultcha: reminder: let's not forget to update this value once/before merging :)

t/re-opt.t
```diff
@@ -0,0 +1,175 @@
...
+--- response_body
+error: pcre_exec() failed: -27
+--- no_error_log
+[error]
+
+
+
+=== TEST 2: increase jit_stack_size
+--- http_config eval
+<<_EOC_;
```
thibaultcha: I believe most tests in this module (and elsewhere) use the `qq` operator as in:
```
--- http_config eval
qq{
    lua_package_path "...";
    init_by_lua_block {

    }
}
```

thibaultcha:
Looking good to me. The tests are passing on my side, everything looks good except for a few nitpicks that I'm not sure we want to address. Feel free to ignore :) 

@alubbe Good job!

thibaultcha:
Q: is it worth documenting the minimum PCRE JIT stack size defined [here](https://github.com/openresty/lua-nginx-module/pull/782/files#diff-5a62a11d0a2e8e55e4c5a7feaac0edd1R37)?

agentzh:
@thibaultcha Thanks for your review!

doujiang24:
LGTM:)

alubbe:
Resolved merge conflicts in travis.yml

alubbe:
Are there any remaining open points?

alubbe:
@agentzh What's the status here, is there anything left for me to do?

agentzh:
@alubbe Looks fine now. Sorry for the delay on my side. Been busy with founding OpenResty Inc. these weeks. I'll look into this ASAP :) Thanks!

alubbe:
No worries, I'm painfully aware of how much time that takes, I've set up and run a few companies. 
Is there any public info on how the foundation and the inc will be funded? Is the foundation going to accept sponsorships from companies using it in production and the inc will run a consultancy-like business offering support on a project / support contract basis?

agentzh:
@alubbe OpenResty Inc. mainly sells licenses of its own commercial products and enterprise solutions based on OpenResty. It's not a consulting company. OpenResty Software Foundation does accept donations from everyone though it's still being set up.

alubbe:
The tests are failing with an error message that I don't think is caused by this PR:
```
/home/travis/build/openresty/lua-nginx-module/src/ngx_http_lua_headers.c:721:13: error: implicit declaration of function 'luaL_getn' is invalid in C99 [-Werror,-Wimplicit-function-declaration]

        n = luaL_getn(L, 3);
```

bjoe2k4:
This is caused by recent changes in LuaJIT: https://github.com/LuaJIT/LuaJIT/commit/dc320ca70f2c5bb3977b82853bcee6dad2523d01

Fix is already there: https://github.com/openresty/lua-nginx-module/commit/844e842cb3d03c3b9b796809bb86102b538fab9e

agentzh:
@alubbe You need to rebase to the latest master. Please do not run `git merge` on your branch since we want to keep the git history as linear as possible. Thank you.

alubbe:
The merge was me trying out the github reconciliation feature. Glad to rebase!

agentzh:
@alubbe Hmm, it's sad that this branch cannot rebase to the latest master. Seeing conflicts.

agentzh:
Getting this:

```diff
agentzh@fedora64 ~/git/lua-resty-core (PR/44)$ g rb master
First, rewinding head to replay your work on top of it...
Applying: Add ngx.re.opt with a "jit_stack_size" option
Applying: Add error messages
Applying: Disallow the changing of jit_stack_size once regex cache is populated
Applying: Update error message to be the same as non-ffi
Applying: Add NYI tests
Using index info to reconstruct a base tree...
M	t/re-find.t
.git/rebase-apply/patch:17: trailing whitespace.
            ngx.re.opt("jit_stack_size", 128 * 1024)
.git/rebase-apply/patch:57: trailing whitespace.
            ngx.re.opt("jit_stack_size", 128 * 1024)
.git/rebase-apply/patch:88: new blank line at EOF.
+
warning: 3 lines add whitespace errors.
Falling back to patching base and 3-way merge...
Auto-merging t/re-find.t
CONFLICT (content): Merge conflict in t/re-find.t
error: Failed to merge in the changes.
Patch failed at 0005 Add NYI tests
The copy of the patch that failed is found in: .git/rebase-apply/patch

When you have resolved this problem, run "git rebase --continue".
If you prefer to skip this patch, run "git rebase --skip" instead.
To check out the original branch and stop rebasing, run "git rebase --abort".
```

Will you fix your git history in this branch. We need a linear history for PRs.

alubbe:
It broke because of the new commit removing the empty lines from the test files - rebased!

agentzh:
@alubbe Merged with the following extra patch. Thanks!

```diff
diff --git a/.travis.yml b/.travis.yml
index 5c70fc8..f99be4c 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,7 +56,7 @@ install:
   - git clone https://github.com/openresty/openresty.git ../openresty
   - git clone https://github.com/openresty/nginx-devel-utils.git
   - git clone https://github.com/simpl/ngx_devel_kit.git ../ndk-nginx-module
-  - git clone -b pcre_jit_stack_size https://github.com/alubbe/lua-nginx-module.git ../lua-nginx-module
+  - git clone https://github.com/openresty/lua-nginx-module.git ../lua-nginx-module
   - git clone https://github.com/openresty/no-pool-nginx.git ../no-pool-nginx
   - git clone https://github.com/openresty/echo-nginx-module.git ../echo-nginx-module
   - git clone https://github.com/openresty/lua-resty-lrucache.git
diff --git a/lib/ngx/re.lua b/lib/ngx/re.lua
index 93e29c9..28e9ed0 100644
--- a/lib/ngx/re.lua
+++ b/lib/ngx/re.lua
@@ -36,8 +36,8 @@ local split_ctx = new_tab(0, 1)
 
 
 ffi.cdef[[
-    int ngx_http_lua_ffi_set_jit_stack_size(int size, unsigned char *errstr,
-                                            size_t *errstr_size);
+int ngx_http_lua_ffi_set_jit_stack_size(int size, unsigned char *errstr,
+    size_t *errstr_size);
 ]]
 
 
@@ -235,8 +235,8 @@ end
 function _M.opt(option, value)
     if option == "jit_stack_size" then
         if not is_regex_cache_empty() then
-            return error("changing jit stack size is not allowed when some " ..
-                         "regexs have already been compiled and cached")
+            error("changing jit stack size is not allowed when some " ..
+                  "regexs have already been compiled and cached")
         end
 
         local errbuf = get_string_buf(MAX_ERR_MSG_LEN)
@@ -247,13 +247,12 @@ function _M.opt(option, value)
 
         if rc == FFI_OK then
             return
-
-        else
-            return error(ffi_str(errbuf, sizep[0]))
         end
+
+        error(ffi_str(errbuf, sizep[0]))
     end
 
-    return error("unrecognized option name")
+    error("unrecognized option name")
 end
 
 
diff --git a/lib/ngx/re.md b/lib/ngx/re.md
index 153daba..c1df8ed 100644
--- a/lib/ngx/re.md
+++ b/lib/ngx/re.md
@@ -176,7 +176,7 @@ The `jit_stack_size` cannot be set to a value lower than PCRE's default of 32K.
 
 This method requires the PCRE library enabled in Nginx.
 
-This feature was first introduced in the `XXX` release.
+This feature was first introduced in the `v0.1.12` release.
 
 [Back to TOC](#table-of-contents)
 
diff --git a/t/re-find.t b/t/re-find.t
index b896949..772ec3b 100644
--- a/t/re-find.t
+++ b/t/re-find.t
@@ -260,4 +260,3 @@ matched: 234
 NYI
 --- error_log eval
 qr/\[TRACE   \d+ content_by_lua\(nginx\.conf:\d+\):4 loop\]/
-
diff --git a/t/re-opt.t b/t/re-opt.t
index d2bda46..18f66a3 100644
--- a/t/re-opt.t
+++ b/t/re-opt.t
@@ -69,12 +69,13 @@ __DATA__
 error: pcre_exec() failed: -27
 --- no_error_log
 [error]
+--- timeout: 10
 
 
 
 === TEST 2: increase jit_stack_size
 --- http_config eval
-<<_EOC_;
+qq{
     lua_package_path "$::pwd/lib/?.lua;../lua-resty-lrucache/lib/?.lua;;";
     init_by_lua_block {
         -- local verbose = true
@@ -97,7 +98,7 @@ error: pcre_exec() failed: -27
         local ngx_re = require "ngx.re"
         ngx_re.opt("jit_stack_size", 128 * 1024)
     }
-_EOC_
+}
 --- config
     location /re {
         content_by_lua_block {
@@ -124,6 +125,7 @@ from: 1
 to: 1563
 --- no_error_log
 [error]
+--- timeout: 10
 
 
 
@@ -153,6 +155,7 @@ qr/changing jit stack size is not allowed when some regexs have already been com
 
 --- grep_error_log_out eval
 ["", "changing jit stack size is not allowed when some regexs have already been compiled and cached\n"]
+--- timeout: 10
 
 
 
@@ -169,7 +172,8 @@ qr/changing jit stack size is not allowed when some regexs have already been com
     }
 --- request
     GET /re
---- response_body
-unrecognized option name
+--- response_body_like chomp
+unrecognized option name$
 --- no_error_log
 [error]
+--- timeout: 10

```

alubbe:
Perfect, it's been a long journey :D
Thanks again for all of your input and patience with me!

PR #47, opened 2016-07-04
feature: add an option init argument in shdict.incr

doujiang24:
C api updated in PR: https://github.com/openresty/lua-nginx-module/pull/579


agentzh:
Applied a modified version of your patch to master. Thanks for the patch!

PR #54, opened 2016-08-11
balancer_by_lua for streams

rshriram:
This PR is the counterpart to https://github.com/openresty/stream-lua-nginx-module/pull/30
Its based on @splitice's https://github.com/openresty/lua-resty-core/pull/25


agentzh:
@rshriram I think you should update `.travis.yml` in your PR to make the Travis CI tests pass.

rshriram:
@agentzh I have updated the Travis file. Temporarily it's pointing to my fork of stream lua module. I want the Travis tests to pass before fixing it. I have also temporarily disabled the timeout tests for stream module until the timeout related compilation issue can be resolved. 

Should this PR get accepted, when will it be available via standard OpenResty release? I would like to avoid shipping custom variants of OpenResty in my docker images.

agentzh:
@rshriram Just patch the `.travis.yml` file in your branch to get Travis CI passing.

Yes, they should go into the official OpenResty distribution eventually. But before that, we should do the following:

1. Merge the corresponding patches into the mainline ngx_stream_lua_module and lua-resty-core.
2. Bundle the ngx_stream_lua module into the OpenResty distribution.

Let's do the 1st item for now.

t/balancer-stream.t
```diff
@@ -0,0 +1,298 @@
...
+--- skip_nginx: 4: < 1.7.5
+--- stream_config
+    lua_package_path "$TEST_NGINX_CWD/lib/?.lua;;";
+
+    proxy_next_upstream_tries 2;
+
+    upstream backend {
+        server 0.0.0.1:80;
+        balancer_by_lua_block {
+            local b = require "ngx.balancer.stream"
```
agentzh: We should use `ngx.balancer` as the API while keeping the implementation in `ngx.balancer.stream`. So the user require `ngx.balancer` instead of `ngx.balancer.stream`. This encourages Lua code reuse on the user Lua land.

rshriram:
One option is to sprinkle a whole bunch of if blocks across the balancer.lua to check if it's in http or stream context (how do I do this?)
I am also getting module not found errors from base.lua. I put stream.lua under the ngx/core/balancer/  folder.

agentzh:
@rshriram No, you can do something like below in the top-level scope of `ngx.balancer`:

```lua
if ngx.config.subsystem == "stream" then
    return require "ngx.balancer.stream"
end
```

Just one `if` statement needed ;)

Also check existing test files and Lua modules in lua-resty-core for how to handle Lua module search paths properly.

rshriram:
Hi @agentzh 
 I have fixed the tests and added the if check in `balancer.lua` so that it automatically pulls in `ngx.balancer.stream` based on the context (as you suggested). 
Just FYI: the code will not pass the tests until the corresponding PR in the stream-lua-nginx-module repository is merged.
Also, I had to omit some tests, such as checks for upstream_addr, as those variables are not really available in the stream context.

PR #58, opened 2016-09-06
add balancer-set-more-tries-fix.t

kingluo:
See https://github.com/openresty/lua-nginx-module/pull/853

### DIFF
--- a/t/balancer-set-more-tries-fix.t
+++ b/t/balancer-set-more-tries-fix.t
@@ -1,0 +1,45 @@
+# vim:set ft= ts=4 sw=4 et fdm=marker:
+
+#use Test::Nginx::Socket 'no_plan';
+use Test::Nginx::Socket::Lua 'no_plan';
+use Cwd qw(cwd);
+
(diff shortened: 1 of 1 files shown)


t/balancer-set-more-tries-fix.t
```diff
@@ -0,0 +1,45 @@
...
+    }
+--- config
+    location = /t {
+        proxy_pass http://backend;
+    }
+--- request
+    GET /t
+--- error_code: 502
+--- no_error_log
+reduced tries due to limit
```
agentzh: I'm afraid this test case would enter an infinite loop with the fixed ngx_lua implementation. This is because `proxy_next_upstream_tries 0` essentially turns off any limit while your `balancer_by_lua_block` handler always add 30 more tries upon *every* try. Have you actually run this test case on your side?

agentzh:
@kingluo I've just fixed your test case in this PR and added it to the openresty/lua-nginx-module repo instead. Thanks!

PR #59, opened 2016-09-08
add testcase if would execeed the limit of proxy_next_upstream_tries

qlee001:
see https://github.com/openresty/lua-nginx-module/pull/866


agentzh:
Already merged into the openresty/lua-nginx-module repo instead. Thanks!

PR #66, opened 2016-10-22
feat: implement 'ngx_re.split()'

thibaultcha:
Hey @agentzh :)

Had some time on my hands so here we go:

Implement a new `ngx.re` module which implements the `split()` API
described at https://github.com/openresty/lua-nginx-module/issues/217

Let me know what you think of this in general before polishing, and if this
is good enough, I will write the documentation for it
(I'd rather not write it until at least the API is approved).

Best,

### DIFF
--- a/lib/ngx/re.lua
+++ b/lib/ngx/re.lua
@@ -1,0 +1,196 @@
+-- I hereby assign copyright in this code to the lua-resty-core project,
+-- to be licensed under the same terms as the rest of the code.
+
+
+local ffi = require 'ffi'
+local bit = require "bit"
+local base = require "resty.core.base"
+local core_regex = require "resty.core.regex"
+
+
+local C = ffi.C
+local sub = string.sub
+local type = type
+local band = bit.band
+local new_tab = base.new_tab
+local tostring = tostring
+local math_max = math.max
+local math_min = math.min
+local re_match_compile = core_regex.re_match_compile
+local destroy_compiled_regex = core_regex.destroy_compiled_regex
+
+
+local FLAG_DFA               = 0x02
+local PCRE_ERROR_NOMATCH     = -1
+local DEFAULT_SPLIT_RES_SIZE = 4
+
+
+local split_ctx = new_tab(0, 1)
+
+
+local _M = { version = base.version }
+
+
+local function re_split_helper(subj, compiled, compile_once, flags, ctx)
+    local rc
+    do
+        local pos = math_max(ctx.pos - 1, 0)
+
+        rc = C.ngx_http_lua_ffi_exec_regex(compiled, flags, subj, #subj, pos)
+    end
+
+    if rc == PCRE_ERROR_NOMATCH then
+        if not compile_once then
+            destroy_compiled_regex(compiled)
+        end
+        return nil, nil, nil
+    end
+
+    if rc < 0 then
+        if not compile_once then
+            destroy_compiled_regex(compiled)
+        end
+        return nil, nil, nil, "pcre_exec() failed: " .. rc
+    end
+
+    if rc == 0 then
+        if band(flags, FLAG_DFA) == 0 then
+            return nil, nil, nil, "capture size too small"
+        end
+
+        rc = 1
+    end
+
+    local caps = compiled.captures
+    local ncaps = compiled.ncaptures
+
+    local from = caps[0] + 1
+    local to = caps[1]
+
+    if from < 0 or to < 0 then
+        return nil, nil, nil
+    end
+
+    ctx.pos = to + 1
+
+    -- retrieve the first sub-match capture if any
+
+    if ncaps > 0 and rc > 1 then
+        return from, to, sub(subj, caps[2] + 1, caps[3])
+    end
+
+    return from, to
+end
+
+
+function _M.split(subj, regex, opts, ctx, max, res)
+    -- we need to cast this to strings to avoid exceptions when they are
+    -- something else.
+    -- needed because of further calls to string.sub in this function.
+    subj = tostring(subj)
+
+    if not ctx then
+        ctx = split_ctx
+        ctx.pos = 1 -- set or reset upvalue field
+
+    elseif not ctx.pos then
+        -- ctx provided by user but missing pos field
+        ctx.pos = 1
+    end
+
+    max = max or 0
+
+    if not res then
+        -- limit the initial arr_n size of res to a reasonable value
+        -- 0 < narr <= DEFAULT_SPLIT_RES_SIZE
+        local narr = DEFAULT_SPLIT_RES_SIZE
+        if max > 0 then
+            -- the user specified a valid max limiter if max > 0
+            narr = math_min(narr, max)
+        end
+
+        res = new_tab(narr, 0)
+
+    elseif type(res) ~= "table" then
+        return error("res is not a table", 2)
+    end
+
+    -- compile regex
+
+    local compiled, compile_once, flags = re_match_compile(regex, opts)
+    if compiled == nil then
+        -- compiled_once holds the error string
+        return nil, compile_once
+    end
+
+    local sub_idx = ctx.pos
+    local res_idx = 0
+
+    -- splitting: with and without a max limiter
+
+    if max > 0 then
+        local count = 1
+
+        while count < max do
+            local from, to, capture, err = re_split_helper(subj, compiled,
+                                                compile_once, flags, ctx)
+            if err then
+                return nil, err
+            end
+
+            if not from then
+                break
+            end
+
+            count = count + 1
+            res_idx = res_idx + 1
+            res[res_idx] = sub(subj, sub_idx, from - 1)
+
+            if capture then
+                res_idx = res_idx + 1
+                res[res_idx] = capture
+            end
+
+            sub_idx = to + 1
+        end
+
+        if count == max then
+            if not compile_once then
+                destroy_compiled_regex(compiled)
+            end
+        end
+
+    else
+        while true do
+            local from, to, capture, err = re_split_helper(subj, compiled,
+                                                compile_once, flags, ctx)
+            if err then
+                return nil, err
+            end
+
+            if not from then
+                break
+            end
+
+            res_idx = res_idx + 1
+            res[res_idx] = sub(subj, sub_idx, from - 1)
+
+            if capture then
+                res_idx = res_idx + 1
+                res[res_idx] = capture
+            end
+
+            sub_idx = to + 1
+        end
+    end
+
+    -- trailing nil for non-cleared res tables
+
+    res[res_idx + 1] = sub(subj, sub_idx, #subj)
+    res[res_idx + 2] = nil
+
+    return res
+end
+
+
+return _M
--- a/lib/ngx/re.md
+++ b/lib/ngx/re.md
@@ -1,0 +1,199 @@
+Name
+====
+
+ngx.re - Lua API for convenience utilities for `ngx.re`.
+
+Table of Contents
+=================
+
+* [Name](#name)
+* [Status](#status)
+* [Synopsis](#synopsis)
+* [Description](#description)
+* [Methods](#methods)
+    * [split](#split)
+* [Community](#community)
+    * [English Mailing List](#english-mailing-list)
+    * [Chinese Mailing List](#chinese-mailing-list)
+* [Bugs and Patches](#bugs-and-patches)
+* [Author](#author)
+* [Copyright and License](#copyright-and-license)
+* [See Also](#see-also)
+
+Status
+======
+
+This Lua module is currently considered experimental.
+
+Synopsis
+========
+
+```lua
+local ngx_re = require "ngx.re"
+
+-- split
+local res, err = ngx_re.split("a,b,c,d", ",")
+--> res is now {"a", "b", "c", "d"}
+```
+
+[Back to TOC](#table-of-contents)
+
+Description
+===========
+
+This Lua module provides a Lua API which implements convenience utilities for
+the `ngx.re` API.
+
+[Back to TOC](#table-of-contents)
+
+Methods
+=======
+
+All the methods of this module are static (or module-level). That is, you do
+not need an object (or instance) to call these methods.
+
+[Back to TOC](#table-of-contents)
+
+split
+-----
+**syntax:** *res, err = ngx_re.split(subject, regex, options?, ctx?, max?, res?)*
+
+Splits the `subject` string using the Perl compatible regular expression
+`regex` with the optional `options`.
+
+This function returns a Lua (array) table (with integer keys) containing the
+splitted values.
+
+In case of error, `nil` will be returned as well as a string describing the
+error.
+
+When `regex` contains a sub-match capturing group, and when such a match is
+found, the first submatch capture will be inserted in between each splitted
+value, like so:
+
+```lua
+local ngx_re = require "ngx.re"
+
+local res, err = ngx_re.split("a,b,c,d", "(,)")
+-- res is now {"a", ",", "b", ",", "c", ",", "d"}
+```
+
+The optional `ctx` table argument can be a Lua table holding an optional `pos`
+field. When the `pos` field in the `ctx` table argument is specified,
+`ngx_re.split` will start splitting the `subject` from that index:
+
+```lua
+local ngx_re = require "ngx.re"
+
+local res, err = ngx_re.split("a,b,c,d", ",", nil, {pos = 5})
+-- res is now {"c", "d"}
+```
+
+The optional `max` argument is a number that when specified, will prevent
+`ngx_re.split` from adding more than `max` matches to the `res` array:
+
+```lua
+local ngx_re = require "ngx.re"
+
+local res, err = ngx_re.split("a,b,c,d", ",", nil, nil, 3)
+-- res is now {"a", "b", "c,d"}
+```
+
+Specifying `max <= 0` disables this behavior, meaning that the number of
+results won't be limited.
+
+The optional 6th argument `res` can be a table that `ngx_re.split` will re-use
+to hold the results instead of creating a new one, which can improve
+performance in hot code paths. It is used like so:
+
+```lua
+local ngx_re = require "ngx.re"
+
+local my_table = {"hello world"}
+
+local res, err = ngx_re.split("a,b,c,d", ",", nil, nil, nil, my_table)
+-- res/my_table is now {"a", "b", "c", "d"}
+```
+
+When provided with a `res` table, `ngx_re.split` won't clear the table
+for performance reasons, but will rather insert a trailing `nil` value
+when the split is completed:
+
+```lua
+local ngx_re = require "ngx.re"
+
+local my_table = {"W", "X", "Y", "Z"}
+
+local res, err = ngx_re.split("a,b", ",", nil, nil, nil, my_table)
+-- res/my_table is now {"a", "b", nil, "Z"}
+```
+
+When the trailing `nil` is not enough for your purpose, you should
+clear the table yourself before feeding it into the `split` function.
+
+[Back to TOC](#table-of-contents)
+
+Community
+=========
+
+[Back to TOC](#table-of-contents)
+
+English Mailing List
+--------------------
+
+The [openresty-en](https://groups.google.com/group/openresty-en) mailing list
+is for English speakers.
+
+[Back to TOC](#table-of-contents)
+
+Chinese Mailing List
+--------------------
+
+The [openresty](https://groups.google.com/group/openresty) mailing list is for
+Chinese speakers.
+
+[Back to TOC](#table-of-contents)
+
+Bugs and Patches
+================
+
+Please report bugs or submit patches by
+
+1. creating a ticket on the [GitHub Issue Tracker](https://github.com/openresty/lua-resty-core/issues),
+1. or posting to the [OpenResty community](#community).
+
+[Back to TOC](#table-of-contents)
+
+Author
+======
+
+Thibault Charbonnier - ([@thibaultcha](https://github.com/thibaultcha))
+
+[Back to TOC](#table-of-contents)
+
+Copyright and License
+=====================
+
+This module is licensed under the BSD license.
+
+Copyright (C) 2016, by Yichun "agentzh" Zhang, CloudFlare Inc.
+
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without modification, are permitted provided that the following conditions are met:
+
+* Redistributions of source code must retain the above copyright notice, this list of conditions and the following disclaimer.
+
+* Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the following disclaimer in the documentation and/or other materials provided with the distribution.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+[Back to TOC](#table-of-contents)
+
+See Also
+========
+* library [lua-resty-core](https://github.com/openresty/lua-resty-core)
+* the ngx_lua module: https://github.com/openresty/lua-nginx-module
+* OpenResty: http://openresty.org
+
+[Back to TOC](#table-of-contents)
--- a/lib/resty/core/regex.lua
+++ b/lib/resty/core/regex.lua
@@ -286,6 +286,9 @@ local function destroy_compiled_regex(compiled)
 end
 
 
+_M.destroy_compiled_regex = destroy_compiled_regex
+
+
 local function re_match_compile(regex, opts)
     local flags = 0
     local pcre_opts = 0
@@ -341,6 +344,9 @@ local function re_match_compile(regex, opts)
 
     return compiled, compile_once, flags
 end
+
+
+_M.re_match_compile = re_match_compile
 
 
 local function re_match_helper(subj, regex, opts, ctx, want_caps, res, nth)
--- a/t/re-split.t
+++ b/t/re-split.t
@@ -1,0 +1,707 @@
+# vim:set ft= ts=4 sw=4 et fdm=marker:
+use lib 'lib';
+use Test::Nginx::Socket::Lua;
+use Cwd qw(cwd);
+
+#worker_connections(1014);
+#master_process_enabled(1);
+#log_level('warn');
+
+repeat_each(2);
+
+plan tests => repeat_each() * blocks() * 4 + (2 * repeat_each());
+
+my $pwd = cwd();
+
+our $HttpConfig = <<_EOC_;
+    lua_package_path "$pwd/lib/?.lua;../lua-resty-lrucache/lib/?.lua;;";
+    init_by_lua_block {
+        -- local verbose = true
+        local verbose = false
+        local outfile = "$Test::Nginx::Util::ErrLogFile"
+        -- local outfile = "/tmp/v.log"
+        if verbose then
+            local dump = require "jit.dump"
+            dump.on(nil, outfile)
+        else
+            local v = require "jit.v"
+            v.on(outfile)
+        end
+
+        require "resty.core"
+        -- jit.opt.start("hotloop=1")
+        -- jit.opt.start("loopunroll=1000000")
+        -- jit.off()
+    }
+_EOC_
+
+no_diff();
+no_long_string();
+check_accum_error_log();
+run_tests();
+
+__DATA__
+
+=== TEST 1: split matches, no submatch, no jit compile, no regex cache
+--- http_config eval: $::HttpConfig
+--- config
+    location = /re {
+        content_by_lua_block {
+            local ngx_re = require "ngx.re"
+
+            local res, err = ngx_re.split("a,b,c,d", ",")
+            if err then
+                ngx.log(ngx.ERR, "failed: ", err)
+                return
+            end
+
+            for i = 1, #res do
+                ngx.say(res[i])
+            end
+        }
+    }
+--- request
+GET /re
+--- response_body
+a
+b
+c
+d
+--- error_log eval
+qr/\[TRACE   \d+/
+--- no_error_log
+[error]
+
+
+
+=== TEST 2: split matches, no submatch, no jit compile, no regex cache
+--- http_config eval: $::HttpConfig
+--- config
+    location = /re {
+        content_by_lua_block {
+            local ngx_re = require "ngx.re"
+
+            local res, err = ngx_re.split("a;,b;,c;,d;e", ";,")
+            if err then
+                ngx.log(ngx.ERR, "failed: ", err)
+                return
+            end
+
+            for i = 1, #res do
+                ngx.say(res[i])
+            end
+        }
+    }
+--- request
+GET /re
+--- response_body
+a
+b
+c
+d;e
+--- error_log eval
+qr/\[TRACE   \d+/
+--- no_error_log
+[error]
+
+
+
+=== TEST 3: split matches, no submatch, jit compile, regex cache
+--- http_config eval: $::HttpConfig
+--- config
+    location = /re {
+        content_by_lua_block {
+            local ngx_re = require "ngx.re"
+
+            local res, err = ngx_re.split("a,b,c,d", ",", "jo")
+            if err then
+                ngx.log(ngx.ERR, "failed: ", err)
+                return
+            end
+
+            for i = 1, #res do
+                ngx.say(res[i])
+            end
+        }
+    }
+--- request
+GET /re
+--- response_body
+a
+b
+c
+d
+--- error_log eval
+qr/\[TRACE   \d+/
+--- no_error_log
+[error]
+
+
+
+=== TEST 4: split matches + submatch (matching)
+--- http_config eval: $::HttpConfig
+--- config
+    location = /re {
+        content_by_lua_block {
+            local ngx_re = require "ngx.re"
+
+            local res, err = ngx_re.split("a;,b;,c;,d,e", "(;),")
+            if err then
+                ngx.log(ngx.ERR, "failed: ", err)
+                return
+            end
+
+            for i = 1, #res do
+                ngx.say(res[i])
+            end
+        }
+    }
+--- request
+GET /re
+--- response_body
+a
+;
+b
+;
+c
+;
+d,e
+--- error_log eval
+qr/\[TRACE   \d+/
+--- no_error_log
+[error]
+
+
+
+=== TEST 5: split matches + submatch (not matching)
+--- http_config eval: $::HttpConfig
+--- config
+    location = /re {
+        content_by_lua_block {
+            local ngx_re = require "ngx.re"
+
+            local res, err = ngx_re.split("a,b,c,d,e", "(;)|,")
+            if err then
+                ngx.log(ngx.ERR, "failed: ", err)
+                return
+            end
+
+            for i = 1, #res do
+                ngx.say(res[i])
+            end
+        }
+    }
+--- request
+GET /re
+--- response_body
+a
+b
+c
+d
+e
+--- error_log eval
+qr/\[TRACE   \d+/
+--- no_error_log
+[error]
+
+
+
+=== TEST 6: split matches + max limiter
+--- http_config eval: $::HttpConfig
+--- config
+    location = /re {
+        content_by_lua_block {
+            local ngx_re = require "ngx.re"
+
+            local res, err = ngx_re.split("a,b,c,d,e", ",", nil, nil, 3)
+            if err then
+                ngx.log(ngx.ERR, "failed: ", err)
+                return
+            end
+
+            for i = 1, #res do
+                ngx.say(res[i])
+            end
+        }
+    }
+--- request
+GET /re
+--- response_body
+a
+b
+c,d,e
+--- error_log eval
+qr/\[TRACE   \d+/
+--- no_error_log
+[error]
+
+
+
+=== TEST 7: split matches + submatch + max limiter
+--- http_config eval: $::HttpConfig
+--- config
+    location = /re {
+        content_by_lua_block {
+            local ngx_re = require "ngx.re"
+
+            local res, err = ngx_re.split("a,b,c,d,e", "(,)", nil, nil, 3)
+            if err then
+                ngx.log(ngx.ERR, "failed: ", err)
+                return
+            end
+
+            for i = 1, #res do
+                ngx.say(res[i])
+            end
+        }
+    }
+--- request
+GET /re
+--- response_body
+a
+,
+b
+,
+c,d,e
+--- error_log eval
+qr/\[TRACE   \d+/
+--- no_error_log
+[error]
+
+
+
+=== TEST 8: split matches + max limiter set to 0
+--- http_config eval: $::HttpConfig
+--- config
+    location = /re {
+        content_by_lua_block {
+            local ngx_re = require "ngx.re"
+
+            local res, err = ngx_re.split("a,b,c,d,e", ",", nil, nil, 0)
+            if err then
+                ngx.log(ngx.ERR, "failed: ", err)
+                return
+            end
+
+            for i = 1, #res do
+                ngx.say(res[i])
+            end
+        }
+    }
+--- request
+GET /re
+--- response_body
+a
+b
+c
+d
+e
+--- error_log eval
+qr/\[TRACE   \d+/
+--- no_error_log
+[error]
+
+
+
+=== TEST 9: split matches + max limiter set to a negative value
+--- http_config eval: $::HttpConfig
+--- config
+    location = /re {
+        content_by_lua_block {
+            local ngx_re = require "ngx.re"
+
+            local res, err = ngx_re.split("a,b,c,d,e", ",", nil, nil, -1)
+            if err then
+                ngx.log(ngx.ERR, "failed: ", err)
+                return
+            end
+
+            for i = 1, #res do
+                ngx.say(res[i])
+            end
+        }
+    }
+--- request
+GET /re
+--- response_body
+a
+b
+c
+d
+e
+--- error_log eval
+qr/\[TRACE   \d+/
+--- no_error_log
+[error]
+
+
+
+=== TEST 10: split matches + max limiter set to 1
+--- http_config eval: $::HttpConfig
+--- config
+    location = /re {
+        content_by_lua_block {
+            local ngx_re = require "ngx.re"
+
+            local res, err = ngx_re.split("a,b,c,d,e", ",", nil, nil, 1)
+            if err then
+                ngx.log(ngx.ERR, "failed: ", err)
+                return
+            end
+
+            for i = 1, #res do
+                ngx.say(res[i])
+            end
+        }
+    }
+--- request
+GET /re
+--- response_body
+a,b,c,d,e
+--- error_log eval
+qr/\[TRACE   \d+/
+--- no_error_log
+[error]
+
+
+
+=== TEST 11: split matches, provided res table
+--- http_config eval: $::HttpConfig
+--- config
+    location = /re {
+        content_by_lua_block {
+            local ngx_re = require "ngx.re"
+
+            local my_table = {}
+
+            local res, err = ngx_re.split("a,b,c,d,e", ",", nil, nil, nil, my_table)
+            if err then
+                ngx.log(ngx.ERR, "failed: ", err)
+                return
+            end
+
+            for i = 1, #res do
+                ngx.say(res[i])
+            end
+        }
+    }
+--- request
+GET /re
+--- response_body
+a
+b
+c
+d
+e
+--- error_log eval
+qr/\[TRACE   \d+/
+--- no_error_log
+[error]
+
+
+
+=== TEST 12: split matches, provided res table (non-cleared)
+--- http_config eval: $::HttpConfig
+--- config
+    location = /re {
+        content_by_lua_block {
+            local ngx_re = require "ngx.re"
+
+            local my_table = {}
+
+            for i = 1, 10 do
+                my_table[i] = i.." hello world"
+            end
+
+            local res, err = ngx_re.split("a,b,c,d,e", ",", nil, nil, nil, my_table)
+            if err then
+                ngx.log(ngx.ERR, "failed: ", err)
+                return
+            end
+
+            for i in ipairs(my_table) do
+                ngx.say(res[i])
+            end
+        }
+    }
+--- request
+GET /re
+--- response_body
+a
+b
+c
+d
+e
+--- error_log eval
+qr/\[TRACE   \d+/
+--- no_error_log
+[error]
+
+
+
+=== TEST 13: split matches, provided res table + max limiter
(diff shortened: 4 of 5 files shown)


lib/resty/core/regex.lua
```diff
@@ -343,7 +343,7 @@ local function re_match_compile(regex, opts)
 end
 
 
-local function re_match_helper(subj, regex, opts, ctx, want_caps, res, nth)
+local function re_match_helper(subj, regex, opts, ctx, want_caps, want_both, res, nth)
```
agentzh: The `want_both` parameter name does not look very clear to me. Will you come up with a better name? Also, it's better to make it the last parameter instead of adding into the middle of the parameter list.

t/re-split.t
```diff
@@ -0,0 +1,496 @@
...
+
+repeat_each(2);
+
+plan tests => repeat_each() * blocks() * 4 + repeat_each();
+
+my $pwd = cwd();
+
+our $HttpConfig = <<_EOC_;
+    lua_package_path "$pwd/lib/?.lua;../lua-resty-lrucache/lib/?.lua;;";
+    init_by_lua '
```
agentzh: For new tests, we should use `init_by_lua_block` instead.

agentzh:
@thibaultcha I see you add an `nth` parameter to `ngx.re.split` as compared to my original API proposal. What does it do?

lib/ngx/re.lua
```diff
@@ -0,0 +1,77 @@
...
+    end
+
+    local idx = ctx.pos
+    local n = 0
+
+    while true do
+        local from, to, err, caps = re_match_helper(subj, regex, opts, ctx, nil, true)
+        if err then
+            return nil, err
+        elseif not from then
```
agentzh: The previous branch always returns, so we should not use `elseif` or `if` here for the 2nd branch.

lib/ngx/re.lua
```diff
@@ -0,0 +1,77 @@
...
+            break
+        end
+
+        if #caps > 0 then
+            if not nth then
+                nth = 1
+            end
+
+            if caps[nth] then
+                res[#res + 1] = caps[nth]
```
agentzh: `#` on tables is expensive (like `O(n)`), better track the index yourself in a local Lua variable.

agentzh:
@doujiang24 @yangshuxin @membphis @moonbingbing Will you have a look at this PR as well? Thanks!

agentzh:
@thibaultcha BTW, thank you very much for looking into this!

lib/ngx/re.lua
thibaultcha: Yeah hehe, I was expecting this :p I was rather waiting for an API review first before diving into those :)
I'll update.

agentzh:
@thibaultcha I'm just not feeling comfortable with the `nth` argument. Other than that, I'm fine with the API design.

thibaultcha:
> I see you add an nth parameter to ngx.re.split as compared to my original API proposal. What does it do?

When a regex provides multiple sub-match capturing groups, the `nth` parameter will allow us to select which group we should extract from, as per this use-case:

https://github.com/openresty/lua-resty-core/pull/66/files#diff-6a76a2e5b6a2e266e1644852ae39bb5cR183

tbh I'm not sure if this is a relevant addition, aka if there is such a use-case for it. I'm fine with removing it if not.

agentzh:
@thibaultcha I think it's good enough to assume `nth == 1`, just like in perl :)

The API is already quite complicated, let's make it a bit simpler :)

thibaultcha:
Patch updated with the requested changes :)

thibaultcha:
(and with the documentation for `ngx_re.split`)

lib/resty/core/regex.lua
```diff
@@ -343,15 +343,15 @@ local function re_match_compile(regex, opts)
 end
 
 
-local function re_match_helper(subj, regex, opts, ctx, want_caps, res, nth)
+local function re_match_helper(subj, regex, opts, ctx, only_caps, res, nth, want_caps)
```
thibaultcha: I took the liberty of renaming in order to carry the meaning of "wanting *only* captures" vs "wanting captures *too*"

lib/ngx/re.lua
```diff
@@ -0,0 +1,74 @@
...
+    elseif type(res) ~= "table" then
+        return error("res is not a table", 2)
+    end
+
+    local sub_idx = ctx.pos
+    local res_idx = #res + 1
+    local count = 0
+
+    while true do
+        local from, to, err, caps = re_match_helper(subj, regex, opts, ctx, nil, nil, nil, true)
```
agentzh: After some thinking, I'm afraid it's not a good idea to call `re_match_helper` here at all. We should not get all the submatch captures when at most one capture will be needed.

I suggest we call lower level primitives to first find out how many submatches are specified in the regexes. And never pay the price of allocating a temporary Lua table as the captures table. Also, we should avoid creating new Lua strings here unless absolutely necessary.

lib/ngx/re.lua
```diff
@@ -0,0 +1,74 @@
...
+    if not ctx then
+        ctx = new_tab(0, 1)
+    end
+
+    if not ctx.pos then
+        ctx.pos = 1
+    end
+
+    if not res then
+        res = new_tab(0, 0)
```
agentzh: For performance reasons, we should use nonzero initial sizes to reduce table auto-growth.

lib/ngx/re.lua
```diff
@@ -0,0 +1,74 @@
...
+
+        if max and count == max then
+            break
+        end
+
+        if caps[1] then
+          res[res_idx] = caps[1]
+          res_idx = res_idx + 1
+        end
+
```
agentzh: Better set a trailing `nil` here just in case the user-supplied `res` table has not yet cleared.

lib/ngx/re.lua
```diff
@@ -0,0 +1,74 @@
+-- Copyright (C) Thibault Charbonnier (thibaultcha)
```
agentzh: Because you are granting copyright to the authors of this library, please remove this copyright notice. Thank you!

lib/ngx/re.lua
```diff
@@ -0,0 +1,74 @@
...
+    end
+
+    if not res then
+        res = new_tab(0, 0)
+    elseif type(res) ~= "table" then
+        return error("res is not a table", 2)
+    end
+
+    local sub_idx = ctx.pos
+    local res_idx = #res + 1
```
agentzh: This is wrong since the user-supplied `res` table might not be cleared. Also, I think we should just initialize `res_idx` to zero. And everytime we want to append a new element, we always *first* increment the `res_idx` variable.

lib/ngx/re.lua
thibaultcha: Sorry, I took example on the `ssl.lua` file. Will do.

lib/ngx/re.lua
thibaultcha: I thought it would be the user's responsibility to clear the table? One of the tests is even currently asserting this behavior. I'll clear it.

thibaultcha: `table.clear` with a fallback implementation should be ok, right?

agentzh: @thibaultcha The users may intentionally avoid clearing the table since clearing can be expensive for relatively large Lua tables. We shouldn't call `table.clear` here ourselves but we should be prepare for the worst case.

lib/ngx/re.lua
thibaultcha: Are you suggesting that I re-implement a subset of `re_match_helper`, including compilation and captures retrieval? (Which would lead to duplicated code which I bet is not so much of an issue here, but I just want to make sure before investing a substantial amount of time into this).

thibaultcha:
@agentzh Patch updated, let me know how you feel about it. I wasn't sure what to export from `core.regex` and what to duplicate, so I ended up only exporting the `compile` and `destroy_compile` helpers, and move error constants and FFI cdefs into `ngx.re` itself.

lib/ngx/re.lua
```diff
@@ -0,0 +1,154 @@
...
+            res[res_idx + 2] = nil -- trailing nil for non-cleared res tables
+            break
+        end
+
+        res_idx = res_idx + 1
+        count = count + 1
+
+        res[res_idx] = sub(subj, sub_idx, from - 1)
+
+        if max and count == max then
```
thibaultcha: Could/should this be replaced with `if count == max then` since `max` can be `nil` but `count` won't?

lib/ngx/re.lua
```diff
@@ -0,0 +1,154 @@
...
+
+
+function _M.split(subj, regex, opts, ctx, max, res)
+    -- we need to cast this to strings to avoid exceptions when they are
+    -- something else.
+    -- needed because of further calls to string.sub in this function.
+    subj = tostring(subj)
+
+    if not ctx then
+        ctx = new_tab(0, 1)
```
doujiang24: @thibaultcha How about use `ctx = { pos = 1 }` here?

lib/ngx/re.lua
```diff
@@ -0,0 +1,154 @@
...
+
+    while true do
+        local from, to, capture, err = re_split_helper(subj, regex, opts, ctx)
+        if err then
+            return nil, err
+        end
+
+        if not from then
+            res[res_idx + 1] = sub(subj, sub_idx, #subj)
+            res[res_idx + 2] = nil -- trailing nil for non-cleared res tables
```
doujiang24: better to move this line to after while loop, since there also have a `break` at line 139 :)

lib/ngx/re.lua
doujiang24: Agreed :)

lib/ngx/re.lua
```diff
@@ -0,0 +1,154 @@
...
+
+        res[res_idx] = sub(subj, sub_idx, from - 1)
+
+        if max and count == max then
+            break
+        end
+
+        if capture then
+          res_idx = res_idx + 1
+          res[res_idx] = capture
```
doujiang24: wrong indent?

lib/ngx/re.lua
```diff
@@ -0,0 +1,154 @@
...
+    if not ctx then
+        ctx = new_tab(0, 1)
+    end
+
+    if not ctx.pos then
+        ctx.pos = 1
+    end
+
+    if not res then
+        res = new_tab(1, 0)
```
doujiang24: better add a blank line after this line.
May be we can use max here? like: 
```lua
max = tonumber(max)
res = new_tab(max and max <= 64 and max or 1, 0)
```

lib/ngx/re.lua
doujiang24: I'm confused here. I personally think it's better to simply ignore capture here.

lib/ngx/re.lua
thibaultcha: Doing so would be a mistake, since instead of:
```lua
local res, err = ngx_re.split("a,b,c,d,e", ",", nil, nil, 3)
-- res is {"a", "b", "c"}
```

We would have:
```lua
local res, err = ngx_re.split("a,b,c,d,e", ",", nil, nil, 3)
-- res is {"a", "b", "c", "d,e"}
```

lib/ngx/re.lua
thibaultcha: Good idea on using `max` is available.

lib/ngx/re.lua
thibaultcha: Eventually, but doing so would duplicate the number of initialization points in the code for `pos`:
```lua
if not ctx then
    ctx = { pos = 1 }
elseif not ctx.pos then
    ctx.pos = 1
end
```

Whether as of now we have:
```lua
if not ctx then
    ctx = new_tab(0, 1)
end

if not ctx.pos then
    ctx.pos = 1 -- only initialization
end
```

Ultimately I am fine with either, I just think the second (current) implementation is cleaner :)

lib/ngx/re.lua
thibaultcha: My bad, I think you are referring to the trailing `nil` only. In that case, yeah! I'll also add a test for this, good catch!

lib/ngx/re.lua
thibaultcha: This is to follow @agentzh 's specification of the API from [this comment](https://github.com/openresty/lua-nginx-module/issues/217#issuecomment-231198630), which I find quite useful. Or maybe the implementation itself is confusing you? Not sure.

thibaultcha:
Patch updated with the latest comments!

lib/ngx/re.lua
```diff
@@ -0,0 +1,159 @@
...
+local destroy_compiled_regex = core_regex.destroy_compiled_regex
+
+
+local PCRE_ERROR_NOMATCH = -1
+
+
+local FLAG_DFA           = 0x02
+
+
+ffi.cdef [[
```
agentzh: Definitions via `ffi.cdef` are global. So if you already require `resty.core.regex` then all the C symbols declared there should already be available.

lib/ngx/re.lua
agentzh: I think instead of creating our own `ctx` table upon every `ngx.re.split` call without a user `ctx` argument (which is common) here, we should share the `ctx` table on the upper level. This is safe since this `split` function never ever yield so there is no possibility for conflicts.

lib/ngx/re.lua
```diff
@@ -0,0 +1,159 @@
...
+        ctx = new_tab(0, 1)
+    end
+
+    if not ctx.pos then
+        ctx.pos = 1
+    end
+
+    if not res then
+        max = tonumber(max)
+        res = new_tab(max and max <= 64 and max or 1, 0)
```
agentzh: Better give the magical number 64 a name. IMHO, 64 is too large in the wild, better be 4.

lib/ngx/re.lua
```diff
@@ -0,0 +1,163 @@
...
+        ctx.pos = 1
+    end
+
+    max = tonumber(max)
+    if not max or max < 0 then
+        max = 0
+    end
+
+    if not res then
+        res = new_tab(max > 0 and max <= 64 and max or 1, 0)
```
agentzh: Better give the magical number 64 a name.

Also, 64 is too large for common cases in reality, how about 4?

lib/ngx/re.lua
```diff
@@ -0,0 +1,163 @@
+-- I hereby assign copyright in this code to the lua-resty-core project,
+-- to be licensed under the same terms as the rest of the code.
+
+local ffi = require 'ffi'
+local bit = require "bit"
+local base = require "resty.core.base"
+local core_regex = require "resty.core.regex"
```
agentzh: I'm not feeling very comfortable with loading the whole `resty.core.regex` module just for `ngx.re`. Since requiring `resty.core.regex` has the side effect of overriding most of the standard `ngx.re` API functions.

I think a better way is to put the main implementations of `resty.core.regex` here, in `ngx.re`, and export functions for `resty.core.regex` to use.

I understand that it is a big change for this PR, maybe we can do this refactoring in the future.

lib/ngx/re.lua
```diff
@@ -0,0 +1,163 @@
...
+    int ngx_http_lua_ffi_exec_regex(ngx_http_lua_regex_t *re, int flags,
+        const unsigned char *s, size_t len, int pos);
+]]
+
+
+local _M = { version = base.version }
+
+
+local function re_split_helper(subj, regex, opts, ctx)
+    local compiled, compile_once, flags = re_match_compile(regex, opts)
```
agentzh: This is unnecessarily inefficient to compile the same regex multiple times (or query the compiled regex cache when `o` opt is specified) in a single `split` call. We should cache the results of `re_match_compile` outside the `while` loop in the `split` function.

lib/ngx/re.lua
thibaultcha: right, I forgot that...

lib/ngx/re.lua
thibaultcha: I'll try to do it in this PR if it's ok.

lib/ngx/re.lua
thibaultcha: I was actually thinking the same this morning!

thibaultcha:
Question: Jade H. Silva's patch was returning the subject string when `max == 0`. Should we do it? Currently, `max <= 0` will disable the maximum-array-size behavior entirely.

lib/ngx/re.lua
thibaultcha: Sounds good. For now should I re-implement `re_match_compile` (it'll be quite the copy-pasta, that one is huge) and `destroy_compiled_regex` in `ngx.re` then?

agentzh: @thibaultcha I'm fine with loading `resty.core.regex` in `ngx.re` right now.

agentzh:
@thibaultcha When `max <= 0`, it should mean "unlimited".

I tend to follow the behavior of perl's `split` builtin here.

lib/ngx/re.lua
```diff
@@ -0,0 +1,163 @@
...
+        return nil, nil, nil, compile_once
+    end
+
+    -- exec the compiled regex
+
+    local rc
+    do
+        local pos = ctx.pos - 1
+        if pos < 0 then
+            pos = 0
```
thibaultcha: LuaJIT's wiki recommends using branch-free algorithms, such as using `math.min` and `math.max` (http://wiki.luajit.org/Numerical-Computing-Performance-Guide). However, I see such code in several places in `resty.core`, hence, should we stick to it or follow Mike Pall's suggestion? I guess this is some sort of general question I'm asking myself after seeing this got no specific comments.

agentzh: @thibaultcha I think we should follow Mike's advice to use `math.min` and `math.max` instead. Yeah, my existing code should eventually get updated as well :)

thibaultcha:
I just updated the patch with all the latest comments and a few optimizations, to the exception of the `resty.core.regex` module loading thing.

What do we think of the patch as it is now? Sorry for all the back and forth, I hope it's not too much.

lib/ngx/re.lua
```diff
@@ -0,0 +1,158 @@
...
+
+
+function _M.split(subj, regex, opts, ctx, max, res)
+    -- we need to cast this to strings to avoid exceptions when they are
+    -- something else.
+    -- needed because of further calls to string.sub in this function.
+    subj = tostring(subj)
+
+    if not ctx or not ctx.pos then
+        ctx = split_ctx
```
agentzh: I think it's safer to only use `split_ctx` when the user `ctx` is nil.

lib/ngx/re.lua
```diff
@@ -0,0 +1,158 @@
...
+
+
+local FLAG_DFA           = 0x02
+
+
+local PCRE_ERROR_NOMATCH = -1
+
+
+local split_ctx = new_tab(0, 1)
+local max_split_res_arr_n = 4
```
agentzh: This name looks wrong to me. Maybe it should be `default_split_result_size`?

agentzh: BTW, constant names should be in all upper case, like `PCRE_ERROR_NOMATCH` and `FLAG_DFA` above.

lib/ngx/re.lua
```diff
@@ -0,0 +1,158 @@
...
+    if not ctx or not ctx.pos then
+        ctx = split_ctx
+        ctx.pos = 1
+    end
+
+    max = math_max(tonumber(max) or 0, 0)
+
+    if not res then
+        -- limit the initial arr_n size of res to a reasonable value
+        res = new_tab(max <= max_split_res_arr_n and max or 1, 0)
```
agentzh: This looks wrong to me. Maybe you mean `max <= DEFAULT_SPLIT_RESULT_SIZE and max or DEFAULT_SPLIT_RESULT_SIZE` here instead?

t/re-split.t
```diff
@@ -0,0 +1,678 @@
...
+                ngx.say(res[i])
+            end
+        }
+    }
+--- request
+GET /re
+--- response_body
+a
+b
+c
```
agentzh: This is JavaScript's `split` behavior. I tend to follow Perl's split semantics here:

```console
$ perl -e 'print join "\n", split /,/, "a,b,c,d,e", 3'
a
b
c,d,e
```

t/re-split.t
```diff
@@ -0,0 +1,678 @@
...
+        }
+    }
+--- request
+GET /re
+--- response_body
+a
+,
+b
+,
+c
```
agentzh: Again, let's follow Perl's semantics:

```console
$ perl -e 'print join "\n", split /(,)/, "a,b,c,d,e", 3'
a
,
b
,
c,d,e
```

lib/ngx/re.lua
agentzh: Well, maybe just `DEFAULT_SPLIT_RES_SIZE`?

agentzh:
@thibaultcha Thanks for your hard work on this!

lib/ngx/re.lua
thibaultcha: Yes, I wasn't entirely sure if the new constant you wanted should be used as an upper-limit to `narr` only or a default value as well. I thought a default value as well would make more sense too, so I feel the same way.

t/re-split.t
thibaultcha: I think this test is correct? It has no `max` argument.

lib/ngx/re.lua
agentzh: @thibaultcha Looking good to me.

t/re-split.t
agentzh: Sorry, I meant this test case instead of TEST 1.

thibaultcha:
I updated it with the proper `narr` feeding for the `res` table and the Perl behavior you described in the test cases.

lib/ngx/re.lua
```diff
@@ -0,0 +1,174 @@
...
+        ctx = split_ctx
+        ctx.pos = 1 -- set or reset upvalue field
+    end
+
+    if not ctx.pos then
+        -- ctx provided by user but missing pos field
+        ctx.pos = 1
+    end
+
+    max = math_max(tonumber(max) or 0, 0)
```
agentzh: I think user-supplied `max` has to be a number and we can save the `tonumber` part.

lib/ngx/re.lua
```diff
@@ -0,0 +1,174 @@
...
+        -- 0 < narr <= DEFAULT_SPLIT_RES_SIZE
+
+        local narr = DEFAULT_SPLIT_RES_SIZE
+        if max > 0 then
+            -- max was specified by the user
+            narr = math_min(max, DEFAULT_SPLIT_RES_SIZE)
+        end
+
+        res = new_tab(narr, 0)
+    elseif type(res) ~= "table" then
```
agentzh: Style: better add a blank line before the line `elseif ...`.

thibaultcha:
@agentzh @doujiang24 Thanks for your detailed reviews and attention on this!

lib/ngx/re.lua
```diff
@@ -0,0 +1,174 @@
...
+    end
+
+    -- exec the compiled regex
+
+    local sub_idx = ctx.pos
+    local res_idx = 0
+    local count = 0
+
+    while true do
+        local from, to, capture, err = re_split_helper(subj, compiled, compile_once, flags, ctx)
```
agentzh: Style: this line has exceeded 80 columns.

lib/ngx/re.lua
```diff
@@ -0,0 +1,174 @@
...
+local destroy_compiled_regex = core_regex.destroy_compiled_regex
+
+
+local FLAG_DFA           = 0x02
+
+
+local PCRE_ERROR_NOMATCH = -1
+
+
+local DEFAULT_SPLIT_RES_SIZE = 4
```
agentzh: I think these constants can be grouped together without using 2 blank lines in between.

lib/ngx/re.md
```diff
@@ -0,0 +1,182 @@
...
+`regex` with the optional `options`.
+
+This function returns a Lua (array) table (with integer keys) containing the
+splitted values.
+
+In case of error, `nil` will be returned as well as a string describing the
+error.
+
+When `regex` contains a sub-match capturing group, and when such a match is
+found, the capture will be inserted in between each splitted value, like so:
```
agentzh: "the capture will be inserted" => "the *first* submatch capture  will be inserted".

thibaultcha:
Sorry :( patch updated :eyes:

agentzh:
@thibaultcha Looking good to me now. Will wait @doujiang24 for his feedback on this :)

Thanks for your fast updates!

lib/ngx/re.lua
```diff
@@ -0,0 +1,170 @@
...
+        res[res_idx] = sub(subj, sub_idx, from - 1)
+
+        if capture then
+            res_idx = res_idx + 1
+            res[res_idx] = capture
+        end
+
+        sub_idx = to + 1
+    end
+
```
thibaultcha: Indeed... I did just that :)

thibaultcha:
Oh, that made all the tests fail... oops, will check tomorrow.

thibaultcha:
Interesting, adding the `destroy_compiled_regex` after the loop and running the tests now crashes the worker with:
```
[TRACE   1 lrucache.lua:61 loop]
1#0: using the "kqueue" event method
2016/10/27 11:47:11 [notice] 41981#0: openresty/1.11.2.1
2016/10/27 11:47:11 [notice] 41981#0: built by clang 8.0.0 (clang-800.0.38)
2016/10/27 11:47:11 [notice] 41981#0: OS: Darwin 16.0.0
2016/10/27 11:47:11 [notice] 41981#0: hw.ncpu: 8
2016/10/27 11:47:11 [notice] 41981#0: net.inet.tcp.sendspace: 131072
2016/10/27 11:47:11 [notice] 41981#0: kern.ipc.somaxconn: 128
2016/10/27 11:47:11 [notice] 41981#0: getrlimit(RLIMIT_NOFILE): 4864:9223372036854775807
nginx(41982,0x7fffd954a3c0) malloc: *** error for object 0x7fd190c0f6a0: pointer being freed was not allocated
*** set a breakpoint in malloc_error_break to debug
```

agentzh:
@thibaultcha I think you should only call `destroy_compiled_regex` when `re_split_helper` does not call it already (i.e., when `re_split_helper` never finds a no match). Otherwise we'll surely have crashes due to double frees :)

agentzh:
@thibaultcha Explicitly calling `destroy_compiled_regex` is just an optimization since we already register it as a GC hook to the cdata object it's freeing.

thibaultcha:
Yeah makes total sense.

thibaultcha:
@doujiang24 I updated the patch to not unnecessarily re-run `re_split_helper` when `max` is reached, but rather went with using `math.huge`, which I think simplifies the function quite a lot: sanitization of the `max` argument is simpler to grasp, same for the `narr` part of created `res` tables, and the loop is easier to grasp and also saves us from evaluating a few expressions (such as the `step` variable and `while`condition). 

I also added a test for `max == 1`. 

@doujiang24 @agentzh I wanted to add a test to ensure we were freeing the compiled regex when `max` is reached. To do so, I wanted to add an `ngx.log` statement, only triggered when `ngx.config.debug` is enabled as per other lua-resty libs, but noticed no other component in resty.core has such tests (and the test suite does not necessarily expects an debug-enabled nginx). Should I still do it? 

Let me know what you think, thanks!

lib/ngx/re.lua
```diff
@@ -0,0 +1,169 @@
...
+    elseif not ctx.pos then
+        -- ctx provided by user but missing pos field
+        ctx.pos = 1
+    end
+
+    if not max or max <= 0 then
+        -- mimic perl's split behavior where max <= 0
+        -- translates to "unlimited"
+        max = math_huge
```
agentzh: I don't like the `math.huge` thing since in dual-number mode, the value in `max` may well be a 32-bit integer instead of a double. Mixing it with `math.huge` may defeat such opportunities of optimizations.

agentzh:
@thibaultcha I don't like adding a debug logging line just for testing. It will add an unnecessary branch to the hot code path. To test memory leaks, we can just use the valgrind test mode to run our tests. See

https://openresty.gitbooks.io/programming-openresty/content/testing/test-modes.html#_valgrind_mode

thibaultcha:
Heavily biased branches should be fine though? However I understand.

agentzh:
@thibaultcha If we are going for the best performance, we can duplicate Lua loop here by branching for the case with a `max` limit and the case without `max` limit.

agentzh:
@thibaultcha biased branching is much better than unbiased ones but still worse than no branch at all.

thibaultcha:
Sure thing.

thibaultcha:
@agentzh Here is a version of the patch that duplicates the loop with and without a `max` limit. Would that be the way we'd rather go?

lib/ngx/re.lua
```diff
@@ -0,0 +1,196 @@
...
+    if not ctx then
+        ctx = split_ctx
+        ctx.pos = 1 -- set or reset upvalue field
+
+    elseif not ctx.pos then
+        -- ctx provided by user but missing pos field
+        ctx.pos = 1
+    end
+
+    max = math_max(max or 0, 0)
```
agentzh: With your latest changes, I think it's fine to just use `max = max or 0` now?

agentzh:
@thibaultcha The separate loops look good to me. Because `split` is more often to be called without a limit, a separate simpler loop dedicated for this case should be worth it. Just my 2 cents :)

@doujiang24 What's your opinion on this?

lib/ngx/re.lua
thibaultcha: addressed

lib/ngx/re.md
```diff
@@ -0,0 +1,183 @@
...
+
+```lua
+local ngx_re = require "ngx.re"
+
+local my_table = {"hello world"}
+
+local res, err = ngx_re.split("a,b,c,d", ",", nil, nil, nil, my_table)
+-- res/my_table is now {"a", "b", "c", "d"}
+```
+
```
agentzh: Better add a note regarding the user responsibility of clearing this `res` table and our behavior of setting a trailing `nil` value as the guard?

thibaultcha: Done, with an example output as well.

doujiang24:
LGTM:)

lib/ngx/re.md
```diff
@@ -0,0 +1,196 @@
...
+Methods
+=======
+
+All the methods of this module are static (or module-level). That is, you do
+not need an object (or instance) to call these methods.
+
+[Back to TOC](#table-of-contents)
+
+split
+----------------
```
agentzh: Better remove those superfluous dashes :)

lib/ngx/re.md
```diff
@@ -0,0 +1,196 @@
...
+
+```lua
+local ngx_re = require "ngx.re"
+
+local my_table = {"W", "X", "Y", "Z"}
+
+local res, err = ngx_re.split("a,b", ",", nil, nil, nil, my_table)
+-- res/my_table is now {"a", "b", nil, "Z"}
+```
+
```
agentzh: Better add another paragraph here as well:

```markdown
When the trailing `nil` is not enough for your purpose, you should
clear the table yourself before feeding it into the `split` function.
```

t/re-split.t
```diff
@@ -0,0 +1,707 @@
...
+abcd
+--- error_log eval
+qr/\[TRACE   \d+/
+--- no_error_log
+[error]
+
+
+
+=== TEST 20: subject is not a string type
+--- http_config eval: $::HttpConfig
```
agentzh: Valgrind finds a memory invalid read issue while running this test case on my side (Linux x86_64):

```
t/re-split.t .. # I found ONLY: maybe you're debugging?
TEST 20: subject is not a string type
==44835== Invalid read of size 16
==44835==    at 0x4024080: ???
==44835==    by 0x905CE9F: ???
==44835==    by 0xFFEFF64AF: ???
==44835==  Address 0x905cea0 is 16 bytes inside a block of size 27 alloc'd
==44835==    at 0x4C28C50: malloc (in /usr/lib64/valgrind/vgpreload_memcheck-amd64-linux.so)
==44835==    by 0x4C2AB1E: realloc (in /usr/lib64/valgrind/vgpreload_memcheck-amd64-linux.so)
==44835==    by 0x54690BF: lj_mem_realloc (lj_gc.c:811)
==44835==    by 0x546AE40: lj_str_new (lj_str.c:188)
==44835==    by 0x5473ABD: lj_strfmt_int (lj_strfmt.c:356)
==44835==    by 0x5465AFA: lj_ff_tostring (in /opt/luajit21sysm/lib/libluajit-5.1.so.2.1.0)
==44835==    by 0x4D5B94: ngx_http_lua_run_thread (ngx_http_lua_util.c:1005)
==44835==    by 0x4D96F0: ngx_http_lua_content_by_chunk (ngx_http_lua_contentby.c:120)
==44835==    by 0x4D98C5: ngx_http_lua_content_handler_inline (ngx_http_lua_contentby.c:310)
==44835==    by 0x4D92AF: ngx_http_lua_content_handler (ngx_http_lua_contentby.c:222)
==44835==    by 0x45581D: ngx_http_core_content_phase (ngx_http_core_module.c:1379)
==44835==    by 0x4504E3: ngx_http_core_run_phases (ngx_http_core_module.c:856)
==44835==    by 0x4505F1: ngx_http_handler (ngx_http_core_module.c:839)
==44835==    by 0x45BC58: ngx_http_process_request (ngx_http_request.c:1916)
==44835==    by 0x45C034: ngx_http_process_request_headers (ngx_http_request.c:1348)
==44835==    by 0x45C349: ngx_http_process_request_line (ngx_http_request.c:1027)
==44835==    by 0x45C783: ngx_http_wait_request_handler (ngx_http_request.c:504)
==44835==    by 0x444A18: ngx_epoll_process_events (ngx_epoll_module.c:900)
==44835==    by 0x43B189: ngx_process_events_and_timers (ngx_event.c:242)
==44835==    by 0x443A6E: ngx_single_process_cycle (ngx_process_cycle.c:309)
==44835==    by 0x41D0B0: main (nginx.c:364)
==44835==
```

t/re-split.t
```diff
@@ -0,0 +1,707 @@
...
+d;e
+--- error_log eval
+qr/\[TRACE   \d+/
+--- no_error_log
+[error]
+
+
+
+=== TEST 3: split matches, no submatch, jit compile, regex cache
+--- http_config eval: $::HttpConfig
```
agentzh: Valgrind finds a memory invalid read issue while running this test case:

```
t/re-split.t .. # I found ONLY: maybe you're debugging?
TEST 3: split matches, no submatch, jit compile, regex cache
==44860== Invalid read of size 16
==44860==    at 0x402407B: ???
==44860==    by 0x9054E5F: ???
==44860==    by 0xFFEFF64AF: ???
==44860==  Address 0x9054e60 is 16 bytes inside a block of size 24 alloc'd
==44860==    at 0x4C28C50: malloc (in /usr/lib64/valgrind/vgpreload_memcheck-amd64-linux.so)
==44860==    by 0x4C2AB1E: realloc (in /usr/lib64/valgrind/vgpreload_memcheck-amd64-linux.so)
==44860==    by 0x54690BF: lj_mem_realloc (lj_gc.c:811)
==44860==    by 0x546AE40: lj_str_new (lj_str.c:188)
==44860==    by 0x547E392: lj_parse_keepstr (lj_parse.c:237)
==44860==    by 0x547B799: lex_string (lj_lex.c:280)
==44860==    by 0x547B799: lex_scan (lj_lex.c:364)
==44860==    by 0x547BA1C: lj_lex_next (lj_lex.c:454)
==44860==    by 0x5480046: parse_args (lj_parse.c:1896)
==44860==    by 0x54801E2: expr_primary (lj_parse.c:1963)
==44860==    by 0x547EC2B: expr_simple (lj_parse.c:2009)
==44860==    by 0x547EC2B: expr_unop (lj_parse.c:2075)
==44860==    by 0x547EC2B: expr_binop (lj_parse.c:2088)
==44860==    by 0x547FEF6: expr (lj_parse.c:2107)
==44860==    by 0x547FEF6: expr_list (lj_parse.c:1874)
==44860==    by 0x54811BF: parse_local (lj_parse.c:2259)
==44860==    by 0x54811BF: parse_stmt (lj_parse.c:2646)
==44860==    by 0x54811BF: parse_chunk (lj_parse.c:2682)
==44860==    by 0x547E5D4: parse_body (lj_parse.c:1851)
==44860==    by 0x547F232: expr_simple (lj_parse.c:2006)
==44860==    by 0x547F232: expr_unop (lj_parse.c:2075)
==44860==    by 0x547F232: expr_binop (lj_parse.c:2088)
==44860==    by 0x547FEF6: expr (lj_parse.c:2107)
==44860==    by 0x547FEF6: expr_list (lj_parse.c:1874)
==44860==    by 0x5480FC6: parse_return (lj_parse.c:2314)
==44860==    by 0x5480FC6: parse_stmt (lj_parse.c:2649)
==44860==    by 0x5480FC6: parse_chunk (lj_parse.c:2682)
==44860==    by 0x5481915: lj_parse (lj_parse.c:2715)
==44860==    by 0x54836FA: cpparser (lj_load.c:41)
==44860==    by 0x54654E6: lj_vm_cpcall (in /opt/luajit21sysm/lib/libluajit-5.1.so.2.1.0)
==44860==    by 0x548388C: lua_loadx (lj_load.c:58)
==44860==    by 0x4DB646: ngx_http_lua_clfactory_loadbuffer (ngx_http_lua_clfactory.c:731)
==44860==    by 0x4D8E60: ngx_http_lua_cache_loadbuffer (ngx_http_lua_cache.c:166)
==44860==    by 0x4D98B0: ngx_http_lua_content_handler_inline (ngx_http_lua_contentby.c:300)
==44860==    by 0x4D92AF: ngx_http_lua_content_handler (ngx_http_lua_contentby.c:222)
==44860==    by 0x45581D: ngx_http_core_content_phase (ngx_http_core_module.c:1379)
==44860==    by 0x4504E3: ngx_http_core_run_phases (ngx_http_core_module.c:856)
==44860==    by 0x4505F1: ngx_http_handler (ngx_http_core_module.c:839)
==44860==    by 0x45BC58: ngx_http_process_request (ngx_http_request.c:1916)
==44860==    by 0x45C034: ngx_http_process_request_headers (ngx_http_request.c:1348)
==44860==    by 0x45C349: ngx_http_process_request_line (ngx_http_request.c:1027)
==44860==    by 0x45C783: ngx_http_wait_request_handler (ngx_http_request.c:504)
==44860==    by 0x444A18: ngx_epoll_process_events (ngx_epoll_module.c:900)
==44860==    by 0x43B189: ngx_process_events_and_timers (ngx_event.c:242)
==44860==    by 0x443A6E: ngx_single_process_cycle (ngx_process_cycle.c:309)
==44860==    by 0x41D0B0: main (nginx.c:364)
==44860==
```

agentzh:
Oh, BTW, please also rebase to the latest git master branch. Thanks!

t/re-split.t
thibaultcha: Any reason why I wouldn't see those on Linux x86_64 too? I installed Valgrind from `apt-get` and added the contents of `lj.supp` to lua-resty-core's `valgrind.suppress`. No complaints from running the test in Valgrind mode. Does Valgrind needs special compilation options?

agentzh: @thibaultcha Because you need to disable the nginx memory pool *and* the built-in allocator of LuaJIT to maximize the strength of valgrind. Please read the book section I referenced earlier in this PR for more details on this: https://openresty.gitbooks.io/programming-openresty/content/testing/test-modes.html#_valgrind_mode

agentzh: @thibaultcha BTW, you do *not* need LuaJIT's `lj.supp` at all if you are using the right setup (including openresty's luajit branch).

thibaultcha: My OpenResty installation is compiled with:

```
./configure --prefix=/usr/local/openresty-valgrind \
  --with-luajit-xcflags='-DLUAJIT_USE_VALGRIND -DLUAJIT_USE_SYSMALLOC' \
  --with-debug \
  --with-openssl=$HOME/tmp/openssl-1.0.2j \
  -j4
```

And tests are run with:
```
PATH="/usr/local/openresty-valgrind/nginx/sbin:$PATH"  TEST_NGINX_USE_VALGRIND=1 prove t/re-split.t
```

Valgrind is not complaining about any of the tests.

```
PATH="/usr/local/openresty-valgrind/nginx/sbin:$PATH" nginx -V 
nginx version: openresty/1.11.2.1
built by gcc 4.8.4 (Ubuntu 4.8.4-2ubuntu1~14.04.3) 
built with OpenSSL 1.0.2j  26 Sep 2016
TLS SNI support enabled
configure arguments: --prefix=/usr/local/openresty-valgrind/nginx --with-debug --with-cc-opt='-DNGX_LUA_USE_ASSERT -DNGX_LUA_ABORT_AT_PANIC -O2' --add-module=../ngx_devel_kit-0.3.0 --add-module=../echo-nginx-module-0.60 --add-module=../xss-nginx-module-0.05 --add-module=../ngx_coolkit-0.2rc3 --add-module=../set-misc-nginx-module-0.31 --add-module=../form-input-nginx-module-0.12 --add-module=../encrypted-session-nginx-module-0.06 --add-module=../srcache-nginx-module-0.31 --add-module=../ngx_lua-0.10.6 --add-module=../ngx_lua_upstream-0.06 --add-module=../headers-more-nginx-module-0.31 --add-module=../array-var-nginx-module-0.05 --add-module=../memc-nginx-module-0.17 --add-module=../redis2-nginx-module-0.13 --add-module=../redis-nginx-module-0.3.7 --add-module=../rds-json-nginx-module-0.14 --add-module=../rds-csv-nginx-module-0.07 --with-ld-opt=-Wl,-rpath,/usr/local/openresty-valgrind/luajit/lib --with-openssl=/home/thibaultcha/tmp/openssl-1.0.2j --with-http_ssl_module
```

thibaultcha: Recompiled with no pool:
```
$ PATH="/usr/local/openresty-valgrind/nginx/sbin:$PATH" nginx -V                                    
nginx version: openresty/1.11.2.1 (no pool)
built by gcc 4.8.4 (Ubuntu 4.8.4-2ubuntu1~14.04.3) 
built with OpenSSL 1.0.2j  26 Sep 2016
TLS SNI support enabled
configure arguments: --prefix=/usr/local/openresty-valgrind/nginx --with-debug --with-cc-opt='-DNGX_LUA_USE_ASSERT -DNGX_LUA_ABORT_AT_PANIC -O2' --add-module=../ngx_devel_kit-0.3.0 --add-module=../echo-nginx-module-0.60 --add-module=../xss-nginx-module-0.05 --add-module=../ngx_coolkit-0.2rc3 --add-module=../set-misc-nginx-module-0.31 --add-module=../form-input-nginx-module-0.12 --add-module=../encrypted-session-nginx-module-0.06 --add-module=../srcache-nginx-module-0.31 --add-module=../ngx_lua-0.10.6 --add-module=../ngx_lua_upstream-0.06 --add-module=../headers-more-nginx-module-0.31 --add-module=../array-var-nginx-module-0.05 --add-module=../memc-nginx-module-0.17 --add-module=../redis2-nginx-module-0.13 --add-module=../redis-nginx-module-0.3.7 --add-module=../rds-json-nginx-module-0.14 --add-module=../rds-csv-nginx-module-0.07 --with-ld-opt=-Wl,-rpath,/usr/local/openresty-valgrind/luajit/lib --with-openssl=/home/thibaultcha/tmp/openssl-1.0.2j --with-http_ssl_module
```

But Valgrind still does not complain, sorry I can't seem to be of help here.

agentzh: @thibaultcha Okay, I've digged into the valgrind error for you :) It's worth recording my process here:

1. I run TEST 3 in t/re-split separately by adding a `--- ONLY` section to that test block.
2. Also, I set the environment `export TEST_NGINX_USE_VALGRIND='--db-attach=yes --quiet --vgdb-error=1'` so that valgrind stops at the first memory issue and lets us use gdb to inspect the nginx process state. Additionally, I add the `---timeout: 1000` section to that test block too so that the client will not time out too soon.
3. By running the `re-split.t` test file with `prove`, it indeed stops at the first error:

    ```
t/re-split.t .. # I found ONLY: maybe you're debugging?
TEST 3: split matches, no submatch, jit compile, regex cache
==85110==
==85110== Warning: --db-attach is a deprecated feature which will be
==85110==    removed in the next release. Use --vgdb-error=1 instead
==85110==
==85110==
==85110== TO DEBUG THIS PROCESS USING GDB: start GDB like this
==85110==   /path/to/gdb nginx
==85110== and then give GDB the following command
==85110==   target remote | /usr/lib64/valgrind/../../bin/vgdb --pid=85110
==85110== --pid is optional if only one valgrind process is running
==85110==
==85110== Invalid read of size 16
==85110==    at 0x402407B: ???
==85110==    by 0x9054EBF: ???
==85110==    by 0xFFEFF646F: ???
==85110==  Address 0x9054ec0 is 16 bytes inside a block of size 24 alloc'd
==85110==    at 0x4C28C50: malloc (in /usr/lib64/valgrind/vgpreload_memcheck-amd64-linux.so)
==85110==    by 0x4C2AB1E: realloc (in /usr/lib64/valgrind/vgpreload_memcheck-amd64-linux.so)
==85110==    by 0x54690BF: lj_mem_realloc (lj_gc.c:811)
==85110==    by 0x546AE40: lj_str_new (lj_str.c:188)
==85110==    by 0x547E392: lj_parse_keepstr (lj_parse.c:237)
==85110==    by 0x547B799: lex_string (lj_lex.c:280)
==85110==    by 0x547B799: lex_scan (lj_lex.c:364)
==85110==    by 0x547BA1C: lj_lex_next (lj_lex.c:454)
==85110==    by 0x5480046: parse_args (lj_parse.c:1896)
==85110==    by 0x54801E2: expr_primary (lj_parse.c:1963)
==85110==    by 0x547EC2B: expr_simple (lj_parse.c:2009)
==85110==    by 0x547EC2B: expr_unop (lj_parse.c:2075)
==85110==    by 0x547EC2B: expr_binop (lj_parse.c:2088)
==85110==    by 0x547FEF6: expr (lj_parse.c:2107)
==85110==    by 0x547FEF6: expr_list (lj_parse.c:1874)
==85110==    by 0x54811BF: parse_local (lj_parse.c:2259)
==85110==    by 0x54811BF: parse_stmt (lj_parse.c:2646)
==85110==    by 0x54811BF: parse_chunk (lj_parse.c:2682)
==85110==
==85110== (action on error) vgdb me ...
```
4. Following the instructions provided in the valgrind output above, we use gdb to connect to valgrind:
    ```
gdb work/nginx/sbin/nginx
(gdb) lbt
builtin#169
@/home/agentzh/git/lua-resty-core/lib/ngx/re.lua:39
@/home/agentzh/git/lua-resty-core/lib/ngx/re.lua:165
=content_by_lua(nginx.conf:71):4
```

    Okay, so the Lua backtrace is shown above for the first memory error. Nice. BTW, I have the following `~/.gdbinit` file:

    ```gdb
directory /home/agentzh/git/openresty-gdb-utils

py import sys
py sys.path.append("/home/agentzh/git/openresty-gdb-utils")

source ngx-lua.gdb
source luajit21.py
set python print-stack full
```

    You may need to edit the file/directory paths above for your local file system layout accordingly.

    It's worth mentioning that we can find out what `builtin#169` is with the following command:

    ```
$ luajit $HOME/git/nginx-devel-utils/ljff.lua 169
FastFunc ffi.meta.__call
```

     Just ensure `luajit` is of exactly the same version of the one your nginx is currently loading. This is because different LuaJIT versions may have incompatible builtin Lua function IDs. So we can see that it's FFI's meta __call builtin.

thibaultcha: I'm sorry but I compile OpenResty as such:
```
./configure --prefix=/usr/local/openresty-valgrind \
  --with-luajit-xcflags='-DLUAJIT_NUMMODE=2 -DLUAJIT_ENABLE_LUA52COMPAT -DLUAJIT_USE_VALGRIND -DLUAJIT_USE_SYSMALLOC -O0' \
  --with-no-pool-patch  \
  --with-openssl=$HOME/tmp/openssl-1.0.2j \
  --with-debug -j4
```

And still don't get any errors, I can't reproduce this on my side nor investigate it.

agentzh: @thibaultcha Okay, I've got to the bottom of it. And it seems to be an issue in my Fedora Linux's valgrind. After switching to my own builds of Valgrind, these memory invalid reads go away. Sorry for the false alarms.

thibaultcha: Okay, good to know! Well, I just rebased on master and addressed the README comments as well!

agentzh: @thibaultcha Cool, thanks!

agentzh:
@thibaultcha Merged with minor edits in your commit log message. Thanks!

thibaultcha:
Would you like a new PR addressing the loading of the `resty.core.regex` module in `ngx.re`?

agentzh:
@thibaultcha If you have the time, sure :)

lib/ngx/re.lua
```diff
@@ -0,0 +1,196 @@
...
+    if rc < 0 then
+        if not compile_once then
+            destroy_compiled_regex(compiled)
+        end
+        return nil, nil, nil, "pcre_exec() failed: " .. rc
+    end
+
+    if rc == 0 then
+        if band(flags, FLAG_DFA) == 0 then
+            return nil, nil, nil, "capture size too small"
```
agentzh: @spacewander Explicit cleanup is just an optimization (since it's already registered a GC hook) and it's not worth it on rare error handling code paths.

PR #69, opened 2016-10-30
change: ssl_session_fetch/store_by_lua_* only allowed in http context.

doujiang24:
update the test cases: move to http context.


agentzh:
Merged. Thanks!

PR #77, opened 2017-01-02
Implement ngx.shared.DICT.decr method

p0pr0ck5:
Corresponding with lua-nginx-module PR #943.

### DIFF
--- a/lib/resty/core/shdict.lua
+++ b/lib/resty/core/shdict.lua
@@ -25,9 +25,9 @@ ffi.cdef[[
         size_t *str_value_len, double *num_value, int *user_flags,
         int get_stale, int *is_stale, char **errmsg);
 
-    int ngx_http_lua_ffi_shdict_incr(void *zone, const unsigned char *key,
-        size_t key_len, double *value, char **err, int has_init, double init,
-        int *forcible);
+    int ngx_http_lua_ffi_shdict_incr(void *zone, int op,
+        const unsigned char *key, size_t key_len, double *value, char **err,
+        int has_init, double init, int *forcible);
 
     int ngx_http_lua_ffi_shdict_store(void *zone, int op,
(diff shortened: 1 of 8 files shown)


p0pr0ck5:
Added tests for the new decr(), but obviously this is failing because the upstream master doesn't have this ffi function in `ngx_http_lua_shdict.c`. I didn't want to update .travis.yml with a patch if it wasn't necessary; once the openresty change is merged these tests will pass.

t/shdict.t
```diff
@@ -999,3 +999,179 @@ number expected, got string
...
 [alert]
 [crit]
+
+
+
+=== TEST 31: decr int
+--- http_config eval: $::HttpConfig
+--- config
+    location = /t {
+        content_by_lua '
```
agentzh: Newly added test cases should use `*_by_lua_block`. Please fix other test cases as well. Thank you.

lordnynex: @p0pr0ck5 Added the above changes in https://github.com/p0pr0ck5/lua-resty-core/pull/1

p0pr0ck5: ^ merged into my branch

lib/resty/core/shdict.lua
```diff
@@ -380,6 +383,16 @@ local function shdict_incr(zone, key, value, init)
 end
 
 
+local function shdict_incr(zone, key, value, init)
+    return shdict_incr_helper(zone, 0x0001, key, value, init)
+end
+
+
+local function shdict_decr(zone, key, value, init)
+    return shdict_incr_helper(zone, 0x0002, key, value, init)
```
agentzh: Can we just reuse the existing C API and negate `value` here? If we want to reject decrementing a negative value, we should do the check just here. But frankly I don't think it's worth it for an extra branch.

PR #79, opened 2017-01-07
bugfix: will enter endless loop when regex is empty string.

doujiang24:
For now, when `regex` is empty string `""`, the `subject` will be split into chars, like so:
```lua

local res, err = ngx_re.split("abcd", "")
-- res is now {"a", "b", "c", "d"}
```

### DIFF
--- a/lib/ngx/re.lua
+++ b/lib/ngx/re.lua
@@ -115,20 +115,28 @@ function _M.split(subj, regex, opts, ctx, max, res)
         return error("res is not a table", 2)
     end
 
+    local len = #subj
+    if ctx.pos > len then
+        res[1] = nil
+        return res
+    end
+
     if regex == "" then
         local pos = ctx.pos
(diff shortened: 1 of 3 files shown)


agentzh:
@thibaultcha Mind to have a look at this PR? Thanks!

thibaultcha:
@agentzh Yes, I saw it yesterday but was flying back to SF and completely jet lagged and on mobile, thanks for the reminder.

@doujiang24 Good catch! I was thinking yesterday: dince we are duplicating a bit of the business logic, maybe a test with user provided `ctx.pos` and `max` arguments would be safer?

doujiang24:
@thibaultcha Thanks for the reminder, there indeed have a bug when `max` is specified.
Now it always returns empty table when `pos` is larger than `subject` length. I think this is better :)

agentzh:
Merged. Thanks!

PR #82, opened 2017-01-12
Add Lua lint tool(luacheck) for testing

halfcrazy:
Luacheck is a powerful linter for Lua. 

Add it to travis-ci will avoid typo, unused variable and etc in pull requests.

Here is a list of warnings Luacheck provided.

http://luacheck.readthedocs.io/en/stable/warnings.html

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -1,0 +1,1 @@
+std = 'ngx_lua'
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,13 @@ compiler:
   - gcc
   - clang
 
+addons:
+  apt:
+    packages:
+    - cpanminus
+    - axel
+    - luarocks
+
 cache:
   directories:
   - download-cache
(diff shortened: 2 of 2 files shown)


agentzh:
@halfcrazy Seems like the Travis CI is failing on this branch. I cannot merge a PR that breaks the Travis CI, obviously.

I've looked at the luacheck warnings. Seems like some warnings indicate something worth improving or fixing while the others are false positives. Will you add fixes for the former and tune maybe the luacheck profile to mute the latter? Thanks!

agentzh:
@halfcrazy Thank you for your contribution!

agentzh:
@halfcrazy I think we should mute the warning "unused argument" in the luacheck profile. Is that configurable?

thibaultcha:
> Is that configurable?

Doable with the `--no-unused-args` flag.

Another very useful flag is `--no-redefined`. When not used, luacheck dislikes the following because the `err` variable is redefined:
```lua
local ok, err = do_thing()
if not ok then
 -- ...
end

local res, err = do_thing2()
if not res then
 -- ...
end
```

Here are all the CLI options for future reference: https://luacheck.readthedocs.io/en/stable/cli.html#command-line-options

agentzh:
@thibaultcha Thanks for the pointers.

agentzh:
It'll be ideal if luacheck can also check

1. source lines exceeding 80 columns, and
2. use of Lua global variables (like those Lua built-in functions) in Lua function bodies (which are usually hot code paths).

Hopefully we can eventually have these especially the latter too. The first one can be added manually to a script, of course. It's very simple.

agentzh:
Okay, I've fixed the issues myself and merged a modified version of this patch to master. I'm closing this.

PR #83, opened 2017-01-17
ffi_write new API: Create output.lua

hkmo:
ffi_write new API function.

Requires PR: https://github.com/openresty/lua-nginx-module/pull/960

This new FFI API function is used to print a string buffer on the client output starting from a given offset and ending at given length. This avoids the overhead of creating another string in the lua land.

### DIFF
--- a/lib/resty/core/output.lua
+++ b/lib/resty/core/output.lua
@@ -1,10 +1,11 @@
 local ffi = require "ffi"
 local base = require "resty.core.base"
 local FFI_OK = base.FFI_OK
+local FFI_ERROR = base.FFI_ERROR
 local C = ffi.C
 
 local _M = {
-    _VERSION = base.version
+    version = base.version
 }
 
 ffi.cdef[[
@@ -16,9 +17,8 @@ ffi.cdef[[
 
 function _M.ffi_write(lua_string, offset, len)
     local r = getfenv(0).__ngx_req
-
-    if not r then       
-       return false
+    if not r then
+        return false, FFI_ERROR
     end
 
     local rc = C.ngx_http_lua_ffi_write(r,
@@ -26,12 +26,12 @@ function _M.ffi_write(lua_string, offset, len)
                                         offset,
                                         len)
 
-    if rc == FFI_OK then      
-      return true
-    else      
-      return false, rc
+    if rc == FFI_OK then
+        return true
+    else
+        return false, rc
     end
- end
+end
 
 
 return _M
--- a/lib/resty/core/output.lua
+++ b/lib/resty/core/output.lua
@@ -14,7 +14,7 @@ ffi.cdef[[
                 size_t len);
 ]]
 
-function output.ffi_write(lua_string, offset, len)
+function _M.ffi_write(lua_string, offset, len)
     local r = getfenv(0).__ngx_req
 
     if not r then       
--- a/lib/resty/core/output.lua
+++ b/lib/resty/core/output.lua
@@ -18,7 +18,7 @@ function output.ffi_write(lua_string, offset, len)
     local r = getfenv(0).__ngx_req
 
     if not r then       
-       return false, rc
+       return false
     end
 
     local rc = C.ngx_http_lua_ffi_write(r,
--- a/lib/resty/core/output.lua
+++ b/lib/resty/core/output.lua
@@ -16,11 +16,9 @@ local _M = {
 
 
 ffi.cdef[[
-        int ngx_http_lua_ffi_write(ngx_http_request_t *r,
-                const char* str,
-                size_t offset,
-                size_t len,
-                char **err);
+    int ngx_http_lua_ffi_write(ngx_http_request_t *r,
+                               const char* str, size_t offset,
+                               size_t len, char **err);
 ]]
 
 
@@ -30,17 +28,14 @@ function _M.ffi_write(lua_string, offset, len)
         return error("no request found")
     end
 
-    local rc = C.ngx_http_lua_ffi_write(r,
-                                        lua_string,
-                                        offset,
-                                        len,
-                                        errmsg)
+    local rc = C.ngx_http_lua_ffi_write(r, lua_string, offset,
+                                        len, errmsg)
 
     if rc == FFI_OK then
         return true
-    else
-        return false, ffi_str(errmsg[0])
     end
+
+    return false, ffi_str(errmsg[0])
 end
 
 
--- a/lib/resty/core/output.lua
+++ b/lib/resty/core/output.lua
@@ -1,13 +1,11 @@
 local ffi = require "ffi"
 local base = require "resty.core.base"
 local FFI_OK = base.FFI_OK
-local ffi_cast = ffi.cast
-local const_char_ptr = ffi.typeof("const char *")
 local C = ffi.C
---local get_string_buf = base.get_string_buf
---local get_size_ptr = base.get_size_ptr
 
-local output = {}
+local _M = {
+    _VERSION = base.version
+}
 
 ffi.cdef[[
         int ngx_http_lua_ffi_write(ngx_http_request_t *r,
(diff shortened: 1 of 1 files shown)


lib/resty/core/output.lua
```diff
@@ -0,0 +1,50 @@
...
+       return false
+    end
+
+    local rc = C.ngx_http_lua_ffi_write(r,
+                           buff,
+                           offset,
+                           len)
+
+    if rc == FFI_OK then
+      ngx.log(ngx.INFO, "ffi_write success: code = "..rc)
```
agentzh: Better avoid this since `ngx.log()` cannot be JIT compiled IIRC.

lib/resty/core/output.lua
```diff
@@ -0,0 +1,50 @@
...
+    if not r then
+       --return false, error("no request found")
+       ngx.log(ngx.ERR, "no request found")
+       return false
+    end
+
+    local rc = C.ngx_http_lua_ffi_write(r,
+                           buff,
+                           offset,
+                           len)
```
agentzh: The indentation is bad. Please see other similar places for examples.

lib/resty/core/output.lua
```diff
@@ -0,0 +1,50 @@
...
+                const char* str,
+                size_t offset,
+                size_t len);
+]]
+
+function output.ffi_write(lua_string, offset, len)
+    --local ERR_BUF_SIZE = 128
+    --local err = get_string_buf(ERR_BUF_SIZE)
+    --local errlen = get_size_ptr()
+    local buff = ffi_cast(const_char_ptr, lua_string)
```
agentzh: I don't think we need an explicit cast here since Lua strings can already cast to `const char *` implicitly.

lib/resty/core/output.lua
```diff
@@ -0,0 +1,50 @@
...
+                           offset,
+                           len)
+
+    if rc == FFI_OK then
+      ngx.log(ngx.INFO, "ffi_write success: code = "..rc)
+      return true
+    else
+      --errlen[0] = ERR_BUF_SIZE
+      --return false, error(ffi_string(err, errlen[0]))
+      ngx.log(ngx.INFO, "ffi_write error: code = "..rc)
```
agentzh: Ditto.

lib/resty/core/output.lua
```diff
@@ -0,0 +1,50 @@
...
+      return false
+    end
+
+ end
+
```
agentzh: Style: two blank lines needed here.

lib/resty/core/output.lua
```diff
@@ -0,0 +1,50 @@
+local ffi = require "ffi"
+local base = require "resty.core.base"
+local FFI_OK = base.FFI_OK
+local ffi_cast = ffi.cast
+local const_char_ptr = ffi.typeof("const char *")
+local C = ffi.C
+--local get_string_buf = base.get_string_buf
+--local get_size_ptr = base.get_size_ptr
+
+local output = {}
```
agentzh: We usually use `_M` for the current module table. See other existing Lua modules in this library for examples.

lib/resty/core/output.lua
```diff
@@ -0,0 +1,50 @@
+local ffi = require "ffi"
+local base = require "resty.core.base"
+local FFI_OK = base.FFI_OK
+local ffi_cast = ffi.cast
+local const_char_ptr = ffi.typeof("const char *")
+local C = ffi.C
+--local get_string_buf = base.get_string_buf
+--local get_size_ptr = base.get_size_ptr
```
agentzh: Please remove these things.

lib/resty/core/output.lua
```diff
@@ -0,0 +1,50 @@
...
+
+local output = {}
+
+ffi.cdef[[
+        int ngx_http_lua_ffi_write(ngx_http_request_t *r,
```
agentzh: Where is this C function defined? Have you created a PR for the openresty/lua-nginx-module repository?

lib/resty/core/output.lua
agentzh: Also, we need a version number from `resty.core.base` for this module. See other modules for examples.

agentzh:
We also need test cases in the existing test suite for this new feature.

lib/resty/core/output.lua
hkmo: ok will remove the commented lines

lib/resty/core/output.lua
hkmo: ok will replace output by _M
and will add version number.

lib/resty/core/output.lua
hkmo: ok, will try to pass the lua string without the ffi_cast.

lib/resty/core/output.lua
hkmo: ok, will adjust the indentation.

lib/resty/core/output.lua
hkmo: not sure I understand this comment, since ngx.log is an exposed API function for any ngx lua script.

lib/resty/core/output.lua
hkmo: I created a PR in openresty/lua-nginx-module, but I beleive it was on a fork. I can move it up to master if needed.

agentzh: @hkmo You should reference the sister PR number in this PR since they are coupled.

lib/resty/core/output.lua
agentzh: @hkmo `ngx.log` should not be used on hot Lua code paths since it cannot be JIT compiled. The whole point of this function is the JIT compilability. We do not generate logs in such FFI-based Lua functions. You can see other similar modules in this library for examples. These API functions should return any errors to the caller and let the caller to handle it (like logging or aborting the request).

lib/resty/core/output.lua
agentzh: @hkmo I looked through lua-nginx-module's PR list and did not find anything. Please always reference the PR number to save the reviewers' time. Thank you!

hkmo: PR created for openresty/lua-nginx-module.

lib/resty/core/output.lua
hkmo: so, we should pass the error code up to the caller.

agentzh:
@hkmo We should return an error message instead of a low level error code. See other Lua API functions in this Lua library for examples.

agentzh:
@hkmo Also, you need to pass the Travis CI testing associated with this PR. It's currently failing (see below on the web page).

lib/resty/core/output.lua
```diff
@@ -0,0 +1,37 @@
...
+    end
+
+    local rc = C.ngx_http_lua_ffi_write(r,
+                                        lua_string,
+                                        offset,
+                                        len)
+
+    if rc == FFI_OK then
+        return true
+    else
```
agentzh: Why use `else`? The previous branch never falls through.

lib/resty/core/output.lua
```diff
@@ -0,0 +1,37 @@
...
+        return false, rc
```
agentzh: Let me repeat that we should return an error message string instead of the raw error code.

lib/resty/core/output.lua
```diff
@@ -0,0 +1,37 @@
...
+local FFI_OK = base.FFI_OK
+local FFI_ERROR = base.FFI_ERROR
+local C = ffi.C
+
+local _M = {
+    version = base.version
+}
+
+ffi.cdef[[
+        int ngx_http_lua_ffi_write(ngx_http_request_t *r,
```
agentzh: Let's reduce the indentation. How many spaces are you using here?

lib/resty/core/output.lua
```diff
@@ -0,0 +1,37 @@
...
+                const char* str,
+                size_t offset,
+                size_t len);
```
agentzh: There is no point in putting each parameter in a separate line. You are far away from 80 columns.

PR #85, opened 2017-01-31
tests: add ngx.exit() input sanitization tests

thibaultcha:
See openresty/lua-nginx-module#969

### DIFF
--- a/t/exit.t
+++ b/t/exit.t
@@ -10,7 +10,7 @@ log_level('warn');
 repeat_each(120);
 #repeat_each(2);
 
-plan tests => repeat_each() * (blocks() * 5);
+plan tests => repeat_each() * (blocks() * 5 - 2);
 
 my $pwd = cwd();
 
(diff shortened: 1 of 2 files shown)


t/exit.t
```diff
@@ -10,7 +10,7 @@ log_level('warn');
 repeat_each(120);
 #repeat_each(2);
 
-plan tests => repeat_each() * (blocks() * 5);
+plan tests => repeat_each() * (blocks() * 5) - (repeat_each() * 2);
```
agentzh: Better use the following form instead:

```
plan tests => repeat_each() * (blocks() * 5 - 2);
```

thibaultcha: Doing so would be less resistant to changes made to `repeat_each()`, and would require manual updates to the plan each time the repeat value is changed. Is that desired?

t/exit.t
```diff
@@ -84,3 +84,99 @@ GET /t
...
+GET /t
+--- response_body_like: 500 Internal Server Error
+--- error_code: 500
+--- error_log eval
+qr/\[error\] .*? bad argument: does not accept NGX_AGAIN or NGX_DONE/
+--- no_error_log eval
+qr/ -- NYI: (?!FastFunc coroutine.yield)/
+
+
+
```
thibaultcha: will remove those trailing whitelines

t/exit.t
agentzh: @thibaultcha I believe these two forms are mathematically identical and are both irrelevant to the value of `repeat_each()`. Am I overlooking anything here?

thibaultcha: With the current plan, we can change the `repeat_each` value without having to change the plan.

With the suggested change, the plan has to be updated when changing the value of `repeat_each()` indeed.

thibaultcha: IIRC (this is a bit old of a PR now), that is because TEST 6 and TEST 7 are each performing one less test than the other tests in this file. Hence `- repeat_each() * 2` does the trick.

agentzh: @thibaultcha Take a closer look at my form:

```perl
plan tests => repeat_each() * (blocks() * 5 - 2);
```

I believe you are misunderstanding it here.

agentzh: @thibaultcha I'm just applying the associative law of multiplication on your formula.

thibaultcha:
Indeed, left my laptop and thought about checking if this was what you meant, bit you've already pointed that out. Should have checked more carefully :)

thibaultcha:
Updated, sorry again for the noise!

agentzh:
@thibaultcha Will you rebase to the latest master and resolve the conflicts? Sorry for the delay on my side. Thanks!

thibaultcha:
@agentzh Do you want https://github.com/openresty/lua-nginx-module/pull/969 to be merged as well? (I thought those PRs were undesired and then forgot about them).

agentzh:
@thibaultcha I think it makes the `ngx.exit()` API safer.

thibaultcha:
Ok, glad it helps! I will proceed.

thibaultcha:
@agentzh PR rebased updated as well

zhuizhuhaomeng (approved):

zhuizhuhaomeng:
resolved conflict and merged

PR #86, opened 2017-02-02
feature: ssl: add new API function to set SSL ciphers

ghedo:
See https://github.com/openresty/lua-nginx-module/pull/961


ghedo:
For the record, tests fail because https://github.com/openresty/lua-nginx-module/pull/961 hasn't been merged yet.

PR #87, opened 2017-02-13
feature: Add FFI functions for ngx.req.get_post_args

p0pr0ck5:


t/request.t
```diff
@@ -571,3 +571,91 @@ qr/\[TRACE   \d+ content_by_lua\(nginx\.conf:\d+\):4 loop\]/
...
+        }
+    }
+--- request
+GET /t?
+--- response_body
+--- error_log eval
+qr/\[TRACE   \d+ content_by_lua\(nginx\.conf:\d+\):5 loop\]/
+--- no_error_log
+[error]
+ -- NYI:
```
thibaultcha: I don't see a test with `get_post_args(n)` and `get_post_args(0)`. I definitely think we need those :)

lib/resty/core/request.lua
```diff
@@ -132,6 +147,80 @@ function ngx.req.get_headers(max_headers, raw)
...
+
+    local n = C.ngx_http_lua_ffi_req_get_post_args_count(r, strbuf,
+        args_len, max_args)
+
+    local nargs = C.ngx_http_lua_ffi_req_get_post_args(r, strbuf,
+        kvbuf, n, args_len)
+
+    local args = new_tab(0, nargs)
+    for i = 0, nargs - 1 do
+        local arg = kvbuf[i]
```
thibaultcha: Should we go against Mike Pall's advice not to cache table lookup in such a hot code path? I am not sure, just to be clear, just raising a concern we previously talked about :)

This happens a bunch in this function, although, I do agree that it makes the code more readable :)

t/request.t
```diff
@@ -571,3 +571,91 @@ qr/\[TRACE   \d+ content_by_lua\(nginx\.conf:\d+\):4 loop\]/
...
 
+
+
+=== TEST 17: ngx.req.get_post_args
+--- http_config eval: $::HttpConfig
+--- config
+    location = /t {
+        set $foo hello;
+        content_by_lua_block {
+            local ffi = require "ffi"
```
p0pr0ck5: Yeah, looks like this is a legacy holdover from a number of other tests (I copypasta'd the test structure from the get_args() tests); i'll remove this one and push up a separate PR to get rid of this cruft in the existing tests.

t/request.t
p0pr0ck5: Quoting @agentzh from https://github.com/openresty/lua-resty-core/pull/78:

The test cases in this library for the FFI-version of ngx_lua's Lua API should focus instead of JIT-ability instead of functionality. Functionality tests are already in the ngx_lua module and ngx_lua's test suite can run with lua-resty-core with the following environment setting, for example:

`export TEST_NGINX_INIT_BY_LUA="package.path = (package.path or '') .. ';$PWD/../lua-resty-core/lib/?.lua;$PWD/../lua-resty-lrucache/lib/?.lua;' require 'resty.core'"`

lib/resty/core/request.lua
p0pr0ck5: I think style consistency is important here; this loop is a duplicate of the one for URI args, so this seems the most appropriate choice. I would be interested in looking at some jit dump output of the two approaches and seeing what the functional difference is.

t/request.t
thibaultcha: Good!

lib/resty/core/request.lua
thibaultcha: Yes, I'm really unsure too, especially since we're dealing with cdata here and not tables...

lib/resty/core/request.lua
```diff
@@ -8,6 +8,10 @@ local base = require "resty.core.base"
 local FFI_BAD_CONTEXT = base.FFI_BAD_CONTEXT
 local FFI_DECLINED = base.FFI_DECLINED
 local FFI_OK = base.FFI_OK
+local FFI_POST_ARGS_OK = 0
+local FFI_POST_ARGS_EMPTY = 1
+local FFI_POST_ARGS_UNREAD = 2
+local FFI_POST_ARGS_IN_FILE = 3
```
agentzh: No extra error code constants.

agentzh:
@p0pr0ck5 Will you please rebase to the latest master? Sorry for the delay on my side. I'm about to look at it but found it has conflicts with the latest master. Thanks!

p0pr0ck5:
@agentzh rebased and squashed.

agentzh:
@p0pr0ck5 Cool, thanks! I'll have a look.

p0pr0ck5:
@agentzh any thoughts here?

PR #91, opened 2017-03-11
chore: added 'coroutine._yield' to read-only globals

thibaultcha:
This fixes builds with Luacheck 0.19.0+


agentzh:
Merged. Thanks!

PR #92, opened 2017-03-11
feat: added ngx.math module for safe PRNG seeding

thibaultcha:
This addresses one of the biggest pitas of ngx_lua: seeding the PRNG with a good entropy seed. Seeding with a low entropy is a *user error* that can lead to **huge** issues.

Ex: producing identical UUIDs with libraries such as [lua-resty-jit-uuid](https://github.com/thibaultcha/lua-resty-jit-uuid).

There are 2 main issues with seeding the PRNG in ngx_lua:

1. We often see such seeding out there in lua-resty libraries or user code:
- `math.randomseed(os.time())` - traditional PUC-Lua seeding -> will lead to all workers having the same seed.
- `math.randomseed(ngx.time() * 1000)` -> will probably lead all workers to having the same seed.
- `math.randomseed(ngx.worker.pid())` -> safer, used by [lua-resty-jit-uuid](https://github.com/thibaultcha/lua-resty-jit-uuid), but can still lead to duplicate seeds if spawning multiple Nginx nodes via containers (where the workers pids are identical)
- Using a seed coming from OpenSSL's `RAND_bytes`, an approach we have taken with Kong.

2. Regardless of all the above methods, if the seed is later overridden by a third-party lua-resty library, all efforts are lost and the PRNG's entropy nullified is again.

To address those concerns, I propose `ngx.math`. Here is a (very) quick draft of the documentation for this module:

---

**usage**: `seed = ngx_math.randomseed(nop?)`
**contexts**: [all but `init`]

Will seed the Lua PRNG via libc's `random()`. This is safer than using a low-entropy seed such as `os.time()` or others, since each Nginx worker's PRNG is already seeded via a combination of `pid + sec + msec`.

The produced seed is returned to the caller to provide  introspection capabilities in case you suspect multiple workers to be using the same seed.

The optional `nop` argument is a boolean that will NOP the `math.randomseed()` function. Doing so will safeguard your PRNG against third-party code doing low-entropy seeding later in your application's lifetime.

You should call this function in the `init_worker_by_lua` context.

---

I believe this will raise awareness for this pita, and encourage users to be safer out there with their PRNG seeding.

### DIFF
--- a/lib/ngx/math.lua
+++ b/lib/ngx/math.lua
@@ -1,0 +1,46 @@
+local ffi = require "ffi"
+local base = require "resty.core.base"
+
+
+local C = ffi.C
+local ngx = ngx
+local error = error
+local tonumber = tonumber
+local math_randomseed = math.randomseed
+
+
+ffi.cdef [[
+    long random(void);
+]]
+
+
+local function random()
+    return tonumber(C.random())
+end
+
+
+local function fnop() end
+
+
+local _M = { version = base.version }
+
+
+function _M.randomseed(nop)
+    if ngx.get_phase() == "init" then
+        return error("API disabled in the current context")
+    end
+
+    local r = random()
+
+    math_randomseed(r)
+
+    if nop then
+        -- luacheck: globals = math.randomseed
+        math.randomseed = fnop
+    end
+
+    return r
+end
+
+
+return _M
--- a/t/math.t
+++ b/t/math.t
@@ -1,0 +1,222 @@
+# vim:set ft= ts=4 sw=4 et fdm=marker:
+use lib 'lib';
+use Test::Nginx::Socket::Lua;
+use Cwd qw(cwd);
+
+master_on();
+workers(5);
+
+repeat_each(3);
+
+plan tests => repeat_each() * (blocks() * 3);
+
+our $pwd = cwd();
+
+our $HttpConfig = <<_EOC_;
+    lua_package_path "$pwd/lib/?.lua;../lua-resty-lrucache/lib/?.lua;;";
+    init_by_lua_block {
+        local verbose = false
+        if verbose then
+            local dump = require "jit.dump"
+            dump.on("b", "$Test::Nginx::Util::ErrLogFile")
+        else
+            local v = require "jit.v"
+            v.on("$Test::Nginx::Util::ErrLogFile")
+        end
+
+        require "resty.core"
+        -- jit.off()
+    }
+_EOC_
+
+check_accum_error_log();
+run_tests();
+
+__DATA__
+
+=== TEST 1: math.random defaults to identical seeds
+--- http_config eval
+qq{
+    lua_shared_dict randoms 1m;
+
+    init_worker_by_lua_block {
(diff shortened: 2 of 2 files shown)


agentzh:
I wonder why this module should be under the `ngx.` namespace since it has no nginx specifics at all.

thibaultcha:
True, I'm not sure either, but don't know of any other convention as to where to provide methods for ngx_lua. I guess ngx_re is a similar example? ngx_math just expects libc to be already seeded by the Nginx workers, just like ngx_re expects the PCRE engine, and both simply make ngx_lua users' life easier?

lib/ngx/math.lua
```diff
@@ -0,0 +1,46 @@
...
+function _M.randomseed(nop)
+    if ngx.get_phase() == "init" then
+        return error("API disabled in the current context")
+    end
+
+    local r = random()
+
+    math_randomseed(r)
+
+    if nop then
```
thibaultcha: According to the purpose of this PR which is to make PRNG seeding safer at an easier cost for users, this NOP behavior should probably be the default, but could be prevented if the user wants to explicitly preserve `math.randomseed`.

thibaultcha:
Rebased on master for the CI builds.

Do you think this should be provided under the `resty.` namespace maybe?

agentzh:
@thibaultcha Because it does not depend on any nginx C API, I don't think it even belongs to this lua-resty-core library.

thibaultcha:
Agreed, but at the same time, I think this is a serious issue that can harm ngx_lua users when decent PRNG is required in their application. This PR's goal is to officially "acknowledge/document" this issue, and provide an easy, out-of-the-box solution.

If you are still not convinced, I will then close this and probably create a `lua-resty-math` module instead, no worries :)

thibaultcha:
Are we sure we don't want this and rather defer "safe seeding" as the user responsibility (if they are aware or this "flaw")?

agentzh:
@thibaultcha I'm afraid it is a bit too hacky to override the global `math.randomseed` API. Also, for Lua modules (pre)loaded in the `init_by_lua*` context, they could cache the original `math.randomseed` function in its own `local` variable thus may bypass the trick this PR introduces.

Nevertheless I agree we should make this issue explicit in the documentation so that the user and module writers can be aware of this risk.

agentzh:
@thibaultcha Also, as I said before, your `ngx.random` module does not use any nginx specific API, so it does not justify the `ngx.` namespace in its module name.

thibaultcha:
> Also, for Lua modules (pre)loaded in the init_by_lua* context, they could cache the original math.randomseed function in its own local variable thus may bypass the trick this PR introduces.

Yes I know, there is no way to fully prevent such a mistake, but I thought `nop`-ing the `randomseed` might cover most of the cases where libraries author out there use it wrong.

> Also, as I said before, your ngx.random module does not use any nginx specific API, so it does not justify the ngx. namespace in its module name.

Yes, this was loud and clear.

> Nevertheless I agree we should make this issue explicit in the documentation so that the user and module writers can be aware of this risk.

Alright, I'll think about it.

Thanks!

PR #95, opened 2017-04-10
Should return nil if subj is nil in split function

gnought:
If subj is nil, the split function will return nil and no error. However the nil is a string "nil", as tostring works. I think it makes sense to return nil type if subj is nil

### DIFF
--- a/lib/ngx/re.lua
+++ b/lib/ngx/re.lua
@@ -84,6 +84,9 @@ end
 
 
 function _M.split(subj, regex, opts, ctx, max, res)
+    if subj == nil then
+        return nil, nil
+    end
     -- we need to cast this to strings to avoid exceptions when they are
     -- something else.
     -- needed because of further calls to string.sub in this function.


lib/ngx/re.lua
```diff
@@ -84,6 +84,9 @@ end
 
 
 function _M.split(subj, regex, opts, ctx, max, res)
+    if subj == nil then
+        return nil, nil
```
agentzh: I don't think I like it. If you really want to detect `nil` input, you should really throw an error here instead instead of passing along `nil`.

gnought: I accept your concern. Could it be stated that behaviour in the re.md?

agentzh: @gnought Well, I prefer make it undocumented. Anything documented has to be ensured backward compatibility forever. It would be the law.

thibaultcha: > you should really throw an error here instead instead of passing along nil.

I feel like this would be a good behavior in such a case. The same way the Lua `string` API and the `ngx.re` API error out when a given subject is `nil`, we should probably also behave the same way. Shouldn't we?

PR #98, opened 2017-04-14
feature: supported intercept error logs.

membphis:
todo: add API documents.

new API:

ngx.errlog() : fetch the intercepted error logs.
ngx.filter_log() : set filter rule, only supported log level now.

this PR depends on: 

https://github.com/openresty/lua-nginx-module/pull/1037
https://github.com/openresty/openresty/pull/249

### DIFF
--- a/lib/ngx/log.lua
+++ b/lib/ngx/log.lua
@@ -1,7 +1,7 @@
 
 local ffi = require 'ffi'
-local ffi_string = ffi.string
 local base = require "resty.core.base"
+local ffi_string = ffi.string
 local get_string_buf = base.get_string_buf
 local get_size_ptr = base.get_size_ptr
 local C = ffi.C
@@ -21,7 +21,7 @@ ffi.cdef[[
     int ngx_http_lua_ffi_set_errlog_filter(int level, unsigned char *err,
         size_t *errlen);
     int ngx_http_lua_ffi_get_errlog_count(unsigned char *err, size_t *errlen);
-    int ngx_http_lua_ffi_get_errlog(ngx_http_lua_ffi_table_elt_t *out,
+    int ngx_http_lua_ffi_get_errlog_data(ngx_http_lua_ffi_table_elt_t *out,
         int max, unsigned char *err, size_t *errlen);
 ]]
 
@@ -77,7 +77,7 @@ function _M.get_errlog(max, logs)
     local raw_buf = get_string_buf(n * table_elt_size)
     local buf = ffi_cast(table_elt_type, raw_buf)
 
-    local rc = C.ngx_http_lua_ffi_get_errlog(buf, n, err, errlen)
+    local rc = C.ngx_http_lua_ffi_get_errlog_data(buf, n, err, errlen)
     if rc == FFI_ERROR then
         return nil, ffi_string(err, errlen[0])
     end
--- a/lib/ngx/log.lua
+++ b/lib/ngx/log.lua
@@ -34,6 +34,10 @@ local FFI_ERROR = base.FFI_ERROR
 
 
 function _M.set_errlog_filter(level)
+    if not level then
+        return nil, [[missing "level" argument]]
+    end
+
     local err = get_string_buf(ERR_BUF_SIZE)
     local errlen = get_size_ptr()
     errlen[0] = ERR_BUF_SIZE
--- a/lib/ngx/log.lua
+++ b/lib/ngx/log.lua
@@ -1,3 +1,4 @@
+-- Copyright (C) Yichun Zhang (agentzh)
 
 
 local ffi = require 'ffi'
@@ -56,7 +57,7 @@ function _M.get_errlog(max, logs)
     max = max or 10
 
     if not logs then
-        logs = new_tab(max, 0)
+        logs = new_tab(max * 2 + 1, 0)
     end
 
     for i = 1, max do
@@ -67,11 +68,12 @@ function _M.get_errlog(max, logs)
         end
 
         if loglen > 0 then
-            logs[i] = {loglevel[0], ffi_string(log[0], loglen)}
+            logs[i * 2 - 1] = loglevel[0]
+            logs[i * 2] = ffi_string(log[0], loglen)
         end
 
         if loglen < 0 or i == max then    -- last one
-            logs[i + 1] = nil
+            logs[i * 2 + 1] = nil
             break
         end
     end
--- a/lib/ngx/log.lua
+++ b/lib/ngx/log.lua
@@ -46,7 +46,7 @@ function _M.set_errlog_filter(level)
 end
 
 
-function _M.get_errlog(max, logs)
+function _M.get_error_logs(max, logs)
     local err = get_string_buf(ERR_BUF_SIZE)
     local errlen = get_size_ptr()
     errlen[0] = ERR_BUF_SIZE
--- a/lib/ngx/log.lua
+++ b/lib/ngx/log.lua
@@ -5,29 +5,22 @@ local ffi_string = ffi.string
 local get_string_buf = base.get_string_buf
 local get_size_ptr = base.get_size_ptr
 local C = ffi.C
-local ffi_cast = ffi.cast
 local new_tab = base.new_tab
+local ffi_new = ffi.new
+local charpp = ffi_new("char *[1]")
+local intp = ffi.new("int[1]")
 
 
 local _M = { version = base.version }
 
 
 ffi.cdef[[
-    typedef struct {
-        ngx_http_lua_ffi_str_t   key;
-        ngx_http_lua_ffi_str_t   value;
-    } ngx_http_lua_ffi_table_elt_t;
-
     int ngx_http_lua_ffi_set_errlog_filter(int level, unsigned char *err,
         size_t *errlen);
-    int ngx_http_lua_ffi_get_errlog_count(unsigned char *err, size_t *errlen);
-    int ngx_http_lua_ffi_get_errlog_data(ngx_http_lua_ffi_table_elt_t *out,
-        int max, unsigned char *err, size_t *errlen);
+    int ngx_http_lua_ffi_get_errlog_data(char **log, size_t *loglen,
+        int *loglevel, unsigned char *err, size_t *errlen);
 ]]
 
-
-local table_elt_type = ffi.typeof("ngx_http_lua_ffi_table_elt_t*")
-local table_elt_size = ffi.sizeof("ngx_http_lua_ffi_table_elt_t")
 
 local ERR_BUF_SIZE = 128
 local FFI_ERROR = base.FFI_ERROR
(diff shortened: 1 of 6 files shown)


lib/resty/core/log.lua
```diff
@@ -0,0 +1,87 @@
...
+        return error(ffi_string(err, errlen[0]))
+    end
+end
+
+
+function ngx.errlog()
+    local err = get_string_buf(ERR_BUF_SIZE)
+    local errlen = get_size_ptr()
+    errlen[0] = ERR_BUF_SIZE
+    local n = C.ngx_http_lua_ffi_errlog_count(err, errlen)
```
agentzh: I think it's better for the caller to specify how many messages at most to fetch.

lib/resty/core/log.lua
```diff
@@ -0,0 +1,87 @@
...
+
+    local raw_buf = get_string_buf(n * table_elt_size)
+    local buf = ffi_cast(table_elt_type, raw_buf)
+
+    local rc = C.ngx_http_lua_ffi_errlog(buf, err, errlen)
+    if rc == FFI_ERROR then
+        return error(ffi_string(err, errlen[0]))
+    end
+
+    local logs = new_tab(n, 0)
```
agentzh: It's better to ask the caller to supply a table so that the caller can recycle it easily. It's expensive to create a new table everytime this function gets called, especially when there is NO error logs.

lib/resty/core/log.lua
```diff
@@ -0,0 +1,87 @@
...
+
+
+function ngx.filter_log(level)
+    local err = get_string_buf(ERR_BUF_SIZE)
+    local errlen = get_size_ptr()
+    errlen[0] = ERR_BUF_SIZE
+    local rc = C.ngx_http_lua_ffi_filter_log(level, err, errlen)
+
+    if rc == FFI_ERROR then
+        return error(ffi_string(err, errlen[0]))
```
agentzh: Better return `nil` and the error string instead of raising an exception.

lib/resty/core/log.lua
```diff
@@ -0,0 +1,87 @@
+-- Copyright (C) OpenResty Inc.
```
agentzh: Please remove this since the copyright should be granted to the authors of this Lua library.

lib/resty/core/log.lua
```diff
@@ -0,0 +1,87 @@
...
+        local log = new_tab(2, 0)
+        local k = buf[i - 1].key
+        local v = buf[i - 1].value
+
+        log[1] = k.len
+        log[2] = ffi_string(v.data, v.len)
+        logs[i] = log
+    end
+
+    return logs;
```
agentzh: Better return the number of error messages instead.

lib/resty/core/log.lua
```diff
@@ -0,0 +1,87 @@
...
+
+
+function ngx.errlog()
+    local err = get_string_buf(ERR_BUF_SIZE)
+    local errlen = get_size_ptr()
+    errlen[0] = ERR_BUF_SIZE
+    local n = C.ngx_http_lua_ffi_errlog_count(err, errlen)
+
+    if n == FFI_ERROR then
+        return error(ffi_string(err, errlen[0]))
```
agentzh: Again: better return `nil` and a string describing the error instead of throwing an exception.

lib/resty/core/log.lua
membphis: if user didn't set directive `lua_intercept_error_log` or patch his nginx core, I think throw a Lua error is fine.

membphis:
@agentzh I have submitted new version right now, please take a review when you have time.

I don't fix one point: https://github.com/openresty/lua-resty-core/pull/98/files#diff-ecd4c8a2f10a7afd880e1482071845aaR57

The error here is mainly due to the lack of user configuration or Nginx core patch. I think it is appropriate to throw a Lua error here.

lib/resty/core/log.lua
```diff
@@ -0,0 +1,98 @@
+-- Copyright (C) Yichun Zhang (agentzh)
```
agentzh: We should introduce a separate `ngx.log` module. The whole point of introducing a new separate Lua module is to stop adding more stuff into the global `ngx` table. But this module still adds more things to the `ngx` global.

lib/resty/core/log.lua
```diff
@@ -0,0 +1,98 @@
...
+
+
+function ngx.errlog_filter(level)
+    local err = get_string_buf(ERR_BUF_SIZE)
+    local errlen = get_size_ptr()
+    errlen[0] = ERR_BUF_SIZE
+    local rc = C.ngx_http_lua_ffi_errlog_set_filter_level(level, err, errlen)
+
+    if rc == FFI_ERROR then
+        return error(ffi_string(err, errlen[0]))
```
agentzh: We shouldn't use `error()` for *all* errors. It's fine for you to throw an error when the user forgets to configure the directive. But it cannot be the general error handling path.

lib/resty/core/log.lua
```diff
@@ -0,0 +1,98 @@
...
+
+
+local table_elt_type = ffi.typeof("ngx_http_lua_ffi_table_elt_t*")
+local table_elt_size = ffi.sizeof("ngx_http_lua_ffi_table_elt_t")
+
+local ERR_BUF_SIZE = 128
+local FFI_ERROR = base.FFI_ERROR
+
+
+function ngx.errlog_filter(level)
```
agentzh: I think I mentioned that we should use verbs in such functions doing some actions?

lib/resty/core/log.lua
```diff
@@ -0,0 +1,98 @@
...
+    errlen[0] = ERR_BUF_SIZE
+    local rc = C.ngx_http_lua_ffi_errlog_set_filter_level(level, err, errlen)
+
+    if rc == FFI_ERROR then
+        return error(ffi_string(err, errlen[0]))
+    end
+end
+
+
+function ngx.errlog(max, logs)
```
agentzh: Ditto. Use verbs.

I have no idea if this function is for logging errors or for fetching error logs by just looking at the function name. It's so ambiguous and so confusing.

lib/resty/core/log.lua
agentzh: See the existing `ngx.semaphore` and `ngx.re` modules in this library.

membphis:
will continue on tomorrow.

membphis:
@agentzh all fixed, pls review :)

lib/ngx/log.lua
```diff
@@ -0,0 +1,83 @@
+
```
agentzh: Extra new line here?

lib/ngx/log.lua
```diff
@@ -0,0 +1,83 @@
...
+
+local _M = { version = base.version }
+
+
+ffi.cdef[[
+    int ngx_http_lua_ffi_set_errlog_filter(int level, unsigned char *err,
+        size_t *errlen);
+    int ngx_http_lua_ffi_get_errlog_data(char **log, size_t *loglen,
+        int *loglevel, unsigned char *err, size_t *errlen);
```
agentzh: Let's remove the leading indentations to save some space in this Lua string literal.

membphis: fixed https://github.com/openresty/lua-resty-core/pull/98/commits/0ac1268d854f2aa17f223149f8c551c43cf75fae :)

lib/ngx/log.lua
```diff
@@ -0,0 +1,83 @@
...
+
+    for i = 1, max do
+        local loglen = C.ngx_http_lua_ffi_get_errlog_data(log,
+                                                      loglevel, err, errlen)
+        if loglen == FFI_ERROR then
+            return nil, ffi_string(err, errlen[0])
+        end
+
+        if loglen > 0 then
+            logs[i] = {loglevel[0], ffi_string(log[0], loglen)}
```
agentzh: We shouldn't create so many little lua tables for every error log message. This is just too expensive in terms of dynamic allocations and GC overhead. How about using a flat table layout like below?

```lua
{level1, msg1, level2, msg2, ...}
```

And the user can iterate the messages like this:

```lua
for i = 1, #logs, 2 do
    local level = logs[i]
    local msg = logs[i + 1]
    ...
end
```

lib/ngx/log.md
```diff
@@ -0,0 +1,231 @@
...
+======
+
+This Lua module is currently considered experimental.
+The API is still in flux and may change in the future without notice.
+
+Synopsis
+========
+
+Intercept nginx error logs with specified log level
+-----------------------------------------
```
agentzh: Style: the dashes should at least as long as the previous line. Please fix other similar places. Thank you.

lib/ngx/log.md
```diff
@@ -0,0 +1,231 @@
...
+    [lua] content_by_lua(nginx.conf:59):5: test3
+```
+
+[Back to TOC](#table-of-contents)
+
+Methods
+=======
+
+set_errlog_filter
+---
```
agentzh: Ditto. More similar places below.

lib/ngx/log.md
```diff
@@ -0,0 +1,231 @@
...
+**syntax:** *status, err = log_module.set_errlog_filter(log_level)*
+
+**context:** *init_by_lua&#42;*
+
+Specify the filter log level, only to intercept the error log we need.
```
agentzh: "Specify" => "Specifies"

lib/ngx/log.md
```diff
@@ -0,0 +1,231 @@
...
+If we don't call this API, all of the error logs will be intercepted by default.
```
agentzh: Better add a note about debugging logs since when OpenResty or NGINX is not built with `--with-debug`, all the debug level logs are suppressed regardless.

lib/ngx/log.md
```diff
@@ -0,0 +1,231 @@
...
+
+[Back to TOC](#table-of-contents)
+
+get_errlog
+---
+**syntax:** *res, err = log_module.get_errlog(max, res?)*
+
+**context:** *init_by_lua&#42;, init_worker_by_lua&#42;, set_by_lua&#42;, rewrite_by_lua&#42;, access_by_lua&#42;, content_by_lua&#42;, header_filter_by_lua&#42;, body_filter_by_lua&#42;, log_by_lua&#42;, ngx.timer.&#42;*
+
+Return the intercepted nginx error logs if successful.
```
agentzh: "Return" => "Returns".

lib/resty/core.lua
```diff
@@ -15,7 +15,6 @@ require "resty.core.response"
 require "resty.core.time"
 require "resty.core.worker"
 
-
```
agentzh: This blank line is intentional. Do not remove it.

lib/ngx/log.lua
membphis: good idea :) submit new version later.

lib/ngx/log.md
membphis: learned :)

lib/ngx/log.md
membphis: yes, fixed https://github.com/openresty/lua-resty-core/pull/98/commits/6845ba0ab737de09150e010bbe69b3a29a1160f6

membphis:
@agentzh all fixed https://github.com/openresty/lua-resty-core/pull/98/commits/afb1cf25ffcc5a81c94cd3e7eda432f65488c093

agentzh:
@membphis Will you please rebase to the latest master? I'm seeing merge conflicts. Thanks!

membphis:
@agentzh  rebased, and I found missing PR for `openresty/openresty-devel-utils/`, submitted too.

membphis:
I rebased the related repositories too :)

t/log.t
```diff
@@ -0,0 +1,514 @@
...
+                error("FAILED " .. err)
+            end
+            ngx.say("log lines:", #res / 2)
+
+            tab_clear(res)
+            res, err = ngx_log.get_errlog(2, res)
+            if not res then
+                error("FAILED " .. err)
+            end
+            ngx.say("log lines:", #res / 2)
```
agentzh: It seems to me that *none* of your existing test cases actually checks the log data inside the table returned by the `get_errlog()` function. All the tests simply test the size of the resulting tables, which is too loose. We should actually check the *contents* inside the table, not just the error log messages in the nginx error log file, which is a very different data flow anyway.

agentzh: Also, we should use directives like `--- response_body_like` and regexes to actually verify that the timestamps and message context suffixes like `client: 127.0.0.1` and `host: "127.0.0.1"` *do* appear in the *captured* error log  messages in that Lua table. Further more, we should check if the error log levels *inside* the Lua table are expected.

lib/ngx/log.lua
```diff
@@ -0,0 +1,85 @@
...
+
+    if rc == FFI_ERROR then
+        return nil, ffi_string(err, errlen[0])
+    end
+
+    return true
+end
+
+
+function _M.get_errlog(max, logs)
```
agentzh: I wonder if we should name it `get_error_logs` instead since it fetches more than a single log message.

t/log.t
agentzh: Finally, we should test multi-line error log message bodies.

agentzh:
@membphis Oh, BTW, we should really need some tests to test if the ring buffer actually works as expected when the error log messages actually flood the capturing buffer and exceed the buffer size limit.

membphis:
@agentzh all fixed, please take a review when you have time :)

t/log.t
```diff
@@ -0,0 +1,647 @@
...
+log lines: #22
+log level:4
+log body:.*access_by_lua\(nginx.conf:\d+\):\d+: --> 99, client: 127.0.0.1,.*host: "localhost".*
+log level:5
+log body:.*access_by_lua\(nginx.conf:\d+\):\d+: --> 100, client: 127.0.0.1,.*host: "localhost".*
+log level:4
+log body:.*access_by_lua\(nginx.conf:\d+\):\d+: --> 100, client: 127.0.0.1,.*host: "localhost".*
+--- skip_nginx: 3: <1.11.2
+
+
```
agentzh: Better use reindex to re-format these new tests.

membphis: @agentzh fixed skip count and reformatted test file.

agentzh:
@membphis Merged with the following extra patch (also added a few more test cases):

https://gist.github.com/agentzh/39d14b695e6c912d5e820e5d7b3a4bac

Thanks!

PR #100, opened 2017-04-17
feature: implemented new API `ngx.process.type` and `ngx.process.privileged_agent`.

membphis:
this PR depends on: 

* git clone -b privilege_process https://github.com/membphis/openresty.git
  * nginx core patch https://github.com/openresty/openresty/pull/250

* git clone -b privilege_process https://github.com/membphis/lua-nginx-module.git 
  * export FFI function https://github.com/openresty/lua-nginx-module/pull/1044


lib/ngx/process.lua
```diff
@@ -0,0 +1,35 @@
...
+int ngx_http_lua_ffi_worker_type(void);
+]]
+
+
+function _M.type()
+    return C.ngx_http_lua_ffi_worker_type()
+end
+
+
+function _M.privilege(enable)
```
agentzh: `enable_privileged_agent()` is better.

lib/ngx/process.md
```diff
@@ -0,0 +1,195 @@
...
+    * [Chinese Mailing List](#chinese-mailing-list)
+* [Bugs and Patches](#bugs-and-patches)
+* [Author](#author)
+* [Copyright and License](#copyright-and-license)
+* [See Also](#see-also)
+
+Status
+======
+
+This Lua module is currently considered experimental.
```
agentzh: Add a note here saying "The API is still in flux and may change in the future without notice".

lib/ngx/process.lua
```diff
@@ -0,0 +1,35 @@
...
+local base = require "resty.core.base"
+local getfenv = getfenv
+
+
+local C = ffi.C
+local _M = { version = base.version }
+
+
+ffi.cdef[[
+void ngx_http_lua_ffi_worker_privileged(int enable);
```
agentzh: `ngx_http_lua_ffi_enable_privileged_agent()` is better.

lib/ngx/process.lua
```diff
@@ -0,0 +1,35 @@
...
+int ngx_http_lua_ffi_worker_type(void);
```
agentzh: `ngx_http_lua_ffi_process_type` is better.

lib/ngx/process.md
```diff
@@ -0,0 +1,195 @@
...
+Returns the current process's type, here is all of the types:
+
+```
+-- core/base.lua
+_M.FFI_PROCESS_SINGLE     = 0
+_M.FFI_PROCESS_MASTER     = 1
+_M.FFI_PROCESS_SIGNALLER  = 2
+_M.FFI_PROCESS_WORKER     = 3
+_M.FFI_PROCESS_HELPER     = 4
+_M.FFI_PROCESS_PRIVILEGE  = 5
```
agentzh: It should be `PRIVILEGED` instead of `PRIVILEGE`.

lib/ngx/process.md
```diff
@@ -0,0 +1,195 @@
...
+```
+
+[Back to TOC](#table-of-contents)
+
+Methods
+=======
+
+type
+---
+**syntax:** *type = process_module.type(n?)*
```
agentzh: What does this `n` parameter do?

lib/ngx/process.md
```diff
@@ -0,0 +1,195 @@
...
+Yichun Zhang &lt;agentzh@gmail.com&gt; (agentzh), OpenResty Inc.
+
+[Back to TOC](#table-of-contents)
+
+Copyright and License
+=====================
+
+This module is licensed under the BSD license.
+
+Copyright (C) 2015-2017, by Yichun "agentzh" Zhang, OpenResty Inc.
```
agentzh: Should be the year 2017 only.

lib/ngx/process.md
```diff
@@ -0,0 +1,195 @@
...
+
+1. creating a ticket on the [GitHub Issue Tracker](https://github.com/openresty/lua-resty-core/issues),
+1. or posting to the [OpenResty community](#community).
+
+[Back to TOC](#table-of-contents)
+
+Author
+======
+
+Yichun Zhang &lt;agentzh@gmail.com&gt; (agentzh), OpenResty Inc.
```
agentzh: I think you are the author of this particular Lua module file :)

lib/ngx/process.md
```diff
@@ -0,0 +1,195 @@
...
+
+```
+
+The example config above produces an output to `error.log` when
+server starts:
+
+```
+[lua] init_by_lua:8: process type:0             --> master process
+[lua] init_worker_by_lua:3: process type:5      --> privilege process
+[lua] init_worker_by_lua:3: process type:3      --> worker process
```
agentzh: I think it's better to return string values like "master", "worker", "privileged agent", and "helper". Constants are difficult to handle for the caller.

lib/ngx/process.lua
membphis: I finally decided to use `ngx_http_lua_ffi_privileged_agent`, I think the user can choose to enable or disable, free to switch.

lib/ngx/process.lua
membphis: `ngx.process.privileged_agent(bool)` changed to this.

lib/ngx/process.md
membphis: great, :)

lib/ngx/process.md
membphis: added :)

membphis:
@agentzh all fixed :)

lib/ngx/process.lua
agentzh: @membphis We need verbs in the method name. That's why I suggested `enable_privileged_agent()`. Otherwise it may seem like a method used to fetch info about the privileged agent process.

lib/ngx/process.lua
agentzh: @membphis I don't think there is any use in disabling this privileged agent since it is disabled by default anyway. So we can omit the parameter as well.

lib/ngx/process.lua
```diff
@@ -0,0 +1,49 @@
...
+local base = require "resty.core.base"
+local getfenv = getfenv
+
+local process_types = {
+    [base.FFI_PROCESS_SINGLE    ]  = "single process",
+    [base.FFI_PROCESS_MASTER    ]  = "master process",
+    [base.FFI_PROCESS_SIGNALLER ]  = "signaller process",
+    [base.FFI_PROCESS_WORKER    ]  = "worker process",
+    [base.FFI_PROCESS_HELPER    ]  = "helper process",
+    [base.FFI_PROCESS_PRIVILEGED]  = "privileged agent process",
```
agentzh: I think we can just omit the ` process` suffix. It's redundant info since they are all process types anyway.

lib/ngx/process.lua
```diff
@@ -0,0 +1,49 @@
...
+}
+
+
+local C = ffi.C
+local _M = { version = base.version }
+
+
+ffi.cdef[[
+void ngx_http_lua_ffi_privileged_agent(int enable);
```
agentzh: We should use `int ngx_http_lua_ffi_enable_privileged_agent(void)` here instead. It should return a status code for future extension.

lib/ngx/process.lua
```diff
@@ -0,0 +1,49 @@
...
+int ngx_http_lua_ffi_process_type(void);
```
agentzh: Again, we should use verbs in the function names. `ngx_http_lua_ffi_get_process_type` is a better name.

lib/ngx/process.lua
```diff
@@ -0,0 +1,56 @@
+-- Copyright (C) Yichun Zhang (agentzh)
+
+
+local ffi = require 'ffi'
+local base = require "resty.core.base"
+local getfenv = getfenv
+local errmsg = base.get_errmsg_ptr()
+local FFI_ERROR = base.FFI_ERROR
+local ffi_str = ffi.string
+
```
agentzh: Style: need 2 blank lines here.

lib/ngx/process.lua
```diff
@@ -0,0 +1,56 @@
...
+    local typ = C.ngx_http_lua_ffi_get_process_type()
+    if is_str_name then
+        return process_types[typ]
+    end
+    return typ
+end
+
+
+function _M.enable_privileged_agent()
+    local r = getfenv(0).__ngx_req
```
agentzh: Why do we need this?

lib/ngx/process.lua
```diff
@@ -0,0 +1,56 @@
...
+
+ffi.cdef[[
+int ngx_http_lua_ffi_enable_privileged_agent(char **err);
+int ngx_http_lua_ffi_get_process_type(void);
+]]
+
+
+function _M.type(is_str_name)
+    local typ = C.ngx_http_lua_ffi_get_process_type()
+    if is_str_name then
```
agentzh: We should always return the string names. Note that Lua's strings are always intern'd, so we can use it as "symbols", just like named constants. String comparisons in Lua is always `O(1)` operation, just like number comparisons.

lib/ngx/process.lua
membphis: `enable_privileged_agent` is only valid in `init_by_lua*` phase.

agentzh: @membphis Ah, okay. Maybe it's better to use `ngx.get_phase()` to do the test?

membphis: much better, submit

lib/ngx/process.lua
```diff
@@ -0,0 +1,53 @@
...
+
+
+function _M.type()
+    local typ = C.ngx_http_lua_ffi_get_process_type()
+    return process_type_name[typ]
+end
+
+
+function _M.enable_privileged_agent()
+    if ngx_phase() ~= "init" then
```
agentzh: It's actually not really about readable, but for portability. We'll soon add a pseudo request to the `init_by_lua*` context (like the `ngx.timer` handler context) to support yield-able Lua API functions like cosockets.

membphis: Wow, NB! @agentzh 👍

README.markdown
```diff
@@ -245,6 +247,14 @@ See the [documentation](./lib/ngx/re.md) for this Lua module for more details.
 
 [Back to TOC](#table-of-contents)
 
+## ngx.process
+
+This Lua module implements is used to manage the nginx process in Lua.
```
agentzh: This sentence has grammar issues.

"implements is used to" does not make any sense.

lib/ngx/process.md
```diff
@@ -0,0 +1,191 @@
...
+Methods
+=======
+
+type
+----
+**syntax:** *type_name = process_module.type()*
+
+**context:** *any*
+
+Returns the current process's type name, here is all of the names:
```
agentzh: ", here" => ". Here".

lib/ngx/process.md
```diff
@@ -0,0 +1,191 @@
...
+
+```lua
+-- core/base.lua
+local process_type_name = {
+    [0 ]  = "single",
+    [1 ]  = "master",
+    [2 ]  = "signaller",
+    [3 ]  = "worker",
+    [4 ]  = "helper",
+    [99]  = "privileged agent",
```
agentzh: We shouldn't expose those magic numbers to the documentation. They belong to implementation details. They are not part of the user API.

lib/ngx/process.md
```diff
@@ -0,0 +1,191 @@
...
+    local process = require "ngx.process"
+
+    -- enable privileged agent process
+    local ok, err = process.enable_privileged_agent(true)
+    if not ok then
+        ngx.log(ngx.ERR, "enable privileged agent failed error:", err)
+    end
+
+    -- output process type
+    ngx.log(ngx.INFO, "process type: ", process.type(true))
```
agentzh: What's the `true` argument for? I don't think the `process.type` function accepts any arguments.

lib/ngx/process.md
```diff
@@ -0,0 +1,191 @@
...
+
+```
+
+The example config above produces an output to `error.log` when
+server starts:
+
+```
+[lua] init_by_lua:11: process type: master process
+[lua] init_worker_by_lua:3: process type: privileged agent process
+[lua] init_worker_by_lua:3: process type: worker process
```
agentzh: I'm afraid the output is stale. We no longer have the " process" suffix in the `process.type` return values.

lib/ngx/process.lua
```diff
@@ -0,0 +1,53 @@
...
+
+local ffi = require 'ffi'
+local base = require "resty.core.base"
+local errmsg = base.get_errmsg_ptr()
+local FFI_ERROR = base.FFI_ERROR
+local ffi_str = ffi.string
+local ngx_phase = ngx.get_phase
+
+
+local process_type_name = {
```
agentzh: We should use plural form of words in container variable names. Here this definitely holds all possible names, so the identifier should be `process_type_names` instead.

lib/ngx/process.md
```diff
@@ -0,0 +1,191 @@
...
+```
+
+The example location above produces a response:
```
agentzh: "produces a response" => "produces the following response body".

lib/ngx/process.md
```diff
@@ -0,0 +1,191 @@
...
+
+[Back to TOC](#table-of-contents)
+
+enable_privileged_agent
+-----------------------
+**syntax:** *ok, err = process_module.enable_privileged_agent()*
+
+**context:** *init_by_lua&#42;*
+
+Enable the privileged agent process in Nginx.
```
agentzh: "Enable" => "Enables"

lib/ngx/process.lua
```diff
@@ -0,0 +1,53 @@
...
+
+ffi.cdef[[
+int ngx_http_lua_ffi_enable_privileged_agent(char **err);
+int ngx_http_lua_ffi_get_process_type(void);
+]]
+
+
+function _M.type()
+    local typ = C.ngx_http_lua_ffi_get_process_type()
+    return process_type_name[typ]
```
agentzh: Better use `tonumber(type)` here as the key. We don't want to use cdata as the keys anyway.

membphis: @agentzh fixed all points, and fixed some other similar points. 
b749027

t/process_type_cache.t
```diff
@@ -0,0 +1,96 @@
...
+                v = typ()
+            end
+            ngx.say("type: ", v)
+        }
+    }
+--- request
+GET /t
+--- response_body
+type: worker
+--- grep_error_log eval
```
agentzh: I'm afraid use of `grep_error_log` might be too strict. Better use `--- error_log` instead?

agentzh:
@membphis Please use dashes instead of underscores in your `.t` test file names. For example, `t/process-type-master.t` looks better than `t/process_type_master.t`.

agentzh:
@membphis Merged with the following extra patch:

```diff
diff --git a/.travis.yml b/.travis.yml
index 1f48e26..fe719ec 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,10 +53,10 @@ install:
   - if [ ! -f download-cache/openssl-$OPENSSL_VER.tar.gz ]; then wget -O download-cache/openssl-$OPENSSL_VER.tar.gz https://www.openssl.org/source/openssl-$OPENSSL_VER.tar.gz; fi
   - if [ ! -f download-cache/pcre-$PCRE_VER.tar.gz ]; then wget -P download-cache http://ftp.cs.stanford.edu/pub/exim/pcre/pcre-$PCRE_VER.tar.gz; fi
   - wget http://nginx.org/download/nginx-${NGINX_VERSION}.tar.gz
-  - git clone -b privilege_process https://github.com/membphis/openresty.git ../openresty
-  - git clone -b privilege_process https://github.com/membphis/openresty-devel-utils.git
+  - git clone https://github.com/openresty/openresty.git ../openresty
+  - git clone https://github.com/openresty/openresty-devel-utils.git
   - git clone https://github.com/simpl/ngx_devel_kit.git ../ndk-nginx-module
-  - git clone -b privilege_process https://github.com/membphis/lua-nginx-module.git ../lua-nginx-module
+  - git clone https://github.com/openresty/lua-nginx-module.git ../lua-nginx-module
   - git clone https://github.com/openresty/no-pool-nginx.git ../no-pool-nginx
   - git clone https://github.com/openresty/echo-nginx-module.git ../echo-nginx-module
   - git clone https://github.com/openresty/lua-resty-lrucache.git
diff --git a/README.markdown b/README.markdown
index 2b0b3c3..56edfea 100644
--- a/README.markdown
+++ b/README.markdown
@@ -253,6 +253,8 @@ This Lua module is used to manage the nginx process in Lua.

 See the [documentation](./lib/ngx/process.md) for this Lua module for more details.

+This module was first introduced in lua-resty-core v0.1.12.
+
 [Back to TOC](#table-of-contents)

 Caveat
diff --git a/t/process_type_cache.t b/t/process-type-cache.t
similarity index 100%
rename from t/process_type_cache.t
rename to t/process-type-cache.t
diff --git a/t/process_type_hup.t b/t/process-type-hup.t
similarity index 100%
rename from t/process_type_hup.t
rename to t/process-type-hup.t
diff --git a/t/process_type_master.t b/t/process-type-master.t
similarity index 100%
rename from t/process_type_master.t
rename to t/process-type-master.t
diff --git a/t/process_type_privileged_agent.t b/t/process-type-privileged-agent.t
similarity index 100%
rename from t/process_type_privileged_agent.t
rename to t/process-type-privileged-agent.t
```

PR #102, opened 2017-04-20
style: fixed by tools ’lua-nginx-module/util/reindex’.

membphis:


agentzh:
@membphis Merged. Thanks!

Issue #104, opened 2017-04-26
I use ngx.re.split(),but receive http 500.


wuyachao:
example:
local ngx_re = require "ngx.re"
local res, err = ngx_re.split("c:574|576|20170424155217","|")

log:
lua entry thread aborted: memory allocation error: not enough memory

use another:

```
function split(str, split_char)
    local sub_str_tab = {};
    while (true) do
        local pos = string.find(str, split_char, 1, true);
        if (not pos) then
            sub_str_tab[#sub_str_tab + 1] = str;
            break;
        end
        local sub_str = string.sub(str, 1, pos - 1);
        sub_str_tab[#sub_str_tab + 1] = sub_str;
        str = string.sub(str, pos + 1, #str);
    end

    return sub_str_tab;
end
```
local res, err = split("c:574|576|20170424155217","|")
ngx.say(res[1])

result:
c:574

I do not know if I have used it correctly.

thibaultcha:
@wuyachao Try escaping the `|`, like so: `"\\|"`. I will take a look at what went wrong exactly later today. Tanks for the report.

wuyachao:
@thibaultcha  
```
local ngx_re = require "ngx.re"
local res, err = ngx_re.split("c:574|576|20170424155217","\\|")
```
Thanks for your suggestion，i can get the result correctly.

thibaultcha:
@wuyachao Follow #106 for a discussion on the fix for this behavior.

thibaultcha:
This has been taken care of by #106, which has been merged :)

wuyachao:
@thibaultcha  thank you

PR #106, opened 2017-04-27
bugfix: ngx.re: split() might enter infinite loops

thibaultcha:
This is a proposed fix for #104. The issue encountered is similar to #79, but #79 cannot address this issue, since the introduced branch only runs when `regex == ""`.

This proposes to unify the behavior triggered by ambiguous regexes, and that behavior would be to return the subject untouched (because the regex is ambiguous). I believe this behavior might be preferable than to split all characters, but sadly, it is a breaking change, alas :'(

Thoughts on how to best handle this?

### DIFF
--- a/lib/ngx/re.lua
+++ b/lib/ngx/re.lua
@@ -77,13 +77,17 @@ local function re_split_helper(subj, compiled, compile_once, flags, ctx)
     local caps = compiled.captures
     local ncaps = compiled.ncaptures
 
-    local from = caps[0] + 1
+    local from = caps[0]
     local to = caps[1]
 
     if from < 0 or to < 0 then
         return nil, nil, nil
     end
 
+    -- convert to Lua string indexes
+
+    from = from + 1
+    to = to + 1
     ctx.pos = to + 1
 
     -- retrieve the first sub-match capture if any
@@ -134,26 +138,6 @@ function _M.split(subj, regex, opts, ctx, max, res)
         return res
     end
 
-    if regex == "" then
-        local pos = ctx.pos
-        local last = len
-        if max > 0 then
-            last = math_min(len, pos + max - 1)
-        end
-
-        local res_idx = 1
-        while pos < last do
-            res[res_idx] = sub(subj, pos, pos)
-            res_idx = res_idx + 1
-            pos = pos + 1
-        end
-
-        res[res_idx] = sub(subj, pos)
-        res[res_idx + 1] = nil
-
-        return res
-    end
-
     -- compile regex
 
     local compiled, compile_once, flags = re_match_compile(regex, opts)
@@ -164,6 +148,7 @@ function _M.split(subj, regex, opts, ctx, max, res)
 
     local sub_idx = ctx.pos
     local res_idx = 0
+    local last_empty_match
 
     -- splitting: with and without a max limiter
 
@@ -181,16 +166,30 @@ function _M.split(subj, regex, opts, ctx, max, res)
                 break
             end
 
-            count = count + 1
-            res_idx = res_idx + 1
-            res[res_idx] = sub(subj, sub_idx, from - 1)
-
-            if capture then
-                res_idx = res_idx + 1
-                res[res_idx] = capture
+            if last_empty_match then
+                sub_idx = last_empty_match
             end
 
-            sub_idx = to + 1
+            if from == to then
+                last_empty_match = from
+            end
+
+            if from > sub_idx or not last_empty_match then
+                count = count + 1
+                res_idx = res_idx + 1
+                res[res_idx] = sub(subj, sub_idx, from - 1)
+
+                if capture then
+                    res_idx = res_idx + 1
+                    res[res_idx] = capture
+                end
+
+                sub_idx = to
+
+                if sub_idx >= len then
+                    break
+                end
+            end
         end
 
         if count == max then
@@ -211,16 +210,31 @@ function _M.split(subj, regex, opts, ctx, max, res)
                 break
             end
 
-            res_idx = res_idx + 1
-            res[res_idx] = sub(subj, sub_idx, from - 1)
-
-            if capture then
-                res_idx = res_idx + 1
-                res[res_idx] = capture
+            if last_empty_match then
+                sub_idx = last_empty_match
             end
 
-            sub_idx = to + 1
+            if from == to then
+                last_empty_match = from
+            end
+
+            if from > sub_idx or not last_empty_match then
+                res_idx = res_idx + 1
+                res[res_idx] = sub(subj, sub_idx, from - 1)
+
+                if capture then
+                    res_idx = res_idx + 1
+                    res[res_idx] = capture
+                end
+
+                sub_idx = to
+
+                if sub_idx >= len then
+                    break
+                end
+            end
         end
+
     end
 
     -- trailing nil for non-cleared res tables
--- a/t/re-split.t
+++ b/t/re-split.t
@@ -766,13 +766,48 @@ GET /re
 4
 5
 len: 5
+--- error_log eval
+qr/\[TRACE   \d+/
 --- no_error_log
 [error]
-[TRACE
 
 
 
-=== TEST 23: regex is "" with pos
+=== TEST 23: regex is "" with max
+--- http_config eval: $::HttpConfig
+--- config
+    location /re {
+        content_by_lua_block {
+            local ngx_re = require "ngx.re"
+
+            local res, err = ngx_re.split("12345", "", "jo", nil, 3)
+            if err then
+                ngx.log(ngx.ERR, "failed: ", err)
+                return
+            end
+
+            for i = 1, #res do
+                ngx.say(res[i])
+            end
+
+            ngx.say("len: ", #res)
+        }
+    }
+--- request
+GET /re
+--- response_body
+1
+2
+345
+len: 3
+--- error_log eval
+qr/\[TRACE   \d+/
+--- no_error_log
+[error]
+
+
+
+=== TEST 24: regex is "" with pos
 --- http_config eval: $::HttpConfig
 --- config
     location /re {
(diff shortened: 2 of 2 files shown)


agentzh:
@thibaultcha I think we should follow perl's `split`'s behavior here:

```console
$ perl -e 'print join ",", split /|/, "abcd"'
a,b,c,d
```

agentzh:
@thibaultcha I think instead of checking if the regex is empty, we should check if the matched separator is an empty string.

thibaultcha:
@agentzh Updated the patch. We now mimic the Perl behavior with empty matches and with `/^/` as well:
```
$ perl -e 'print join ":", split //, "abcd"'
a:b:c:d

$ perl -e 'print join ":", split /|/, "abcd"'
a:b:c:d

$ perl -e 'print join ":", split /()/, "abcd"'
a::b::c::d

$ perl -e 'print join ":", split /^/, "ab\ncd"'
ab
:cd

$ perl -e 'print join ":", split /^/m, "ab\ncd"'
ab
:cd

$ perl -e 'print join ":", split / ^/, "ab\ncd"'
ab
cd

$ perl -e 'print join ":", split / ^/x, "ab\ncd"'
ab
:cd
```

What are your thoughts on respecting this behavior?

lib/ngx/re.lua
```diff
@@ -198,15 +210,25 @@ function _M.split(subj, regex, opts, ctx, max, res)
...
 
+            if from == to and not start_regex then
+                from = from + 1
+                to = from
+            end
+
             res_idx = res_idx + 1
             res[res_idx] = sub(subj, sub_idx, from - 1)
 
+
```
agentzh: Extra blank line?

lib/ngx/re.lua
```diff
@@ -89,6 +94,10 @@ function _M.split(subj, regex, opts, ctx, max, res)
     -- needed because of further calls to string.sub in this function.
     subj = tostring(subj)
 
+    if not opts then
+      opts = ""
```
agentzh: 2-space indentations?

lib/ngx/re.lua
thibaultcha: oops, oversight...

lib/ngx/re.lua
```diff
@@ -121,24 +130,21 @@ function _M.split(subj, regex, opts, ctx, max, res)
...
+    local res_idx = 0
+    local sub_idx = ctx.pos
 
-        local res_idx = 1
-        while pos < last do
-            res[res_idx] = sub(subj, pos, pos)
-            res_idx = res_idx + 1
-            pos = pos + 1
-        end
+    local start_regex = find(regex, "^", nil, true) ~= nil
```
agentzh: Hmm, this is too hacky to scan the regex literal using `string.find`.

agentzh: I think the regex engine should handle this automatically. If not, then there must be some other issues.

thibaultcha: Not satisfied with it either... Open to suggestions!

thibaultcha: > I think the regex engine should handle this automatically. If not, then there must be some other issues.

Sorry missed this. Hmm, I will investigate.

thibaultcha: On `subj = "ab\ncd"` with `/^/` If we don't add the `m` flag, we get:
```
{ "a", "b\ncd" }
```

So we do not match on newlines automatically, but only at position 0, because it is our first character, and the rest doesn't get matched. It seemed to me (after a bit of online research) that Perl's `split` was giving some special care of its own to `/^/`, but I could definitely be wrong.

agentzh: Maybe you should try the "m" regex flag in the Lua split? By default, ^ only matches the beginning of the string IIRC.

thibaultcha: > Maybe you should try the "m" regex flag in the Lua split?

You mean require users to add it themselves in the opts argument? Because Perl's `split` behavior is - apparently - to be that it adds the `m` automatically when `/^/` is given.

agentzh: @thibaultcha Well, here we should follow our own semantics instead of perl's:

```console
$ resty -e 'print((ngx.re.match("a\nb", "^b", "")) and "true" or "false")'
false

$ resty -e 'print((ngx.re.match("a\nb", "^b", "m")) and "true" or "false")'
true
```

thibaultcha: Makes our lives easier, I'm not against it! 😅 Will update.

thibaultcha:
I updated the patch to not incorporate Perl's split behavior with `/^/`. However, to preserve the behavior introduced with #79, we still check for empty regexes. I'm not sure if we should check for all empty matches regexes like `()` or `|`, which currently have a different behavior than `""`:

```lua
-- empty regex:
ngx_re.split("abcd", "") -- { "a", "b", "c", "d" }
-- others:
ngx_re.split("abcd", "|") -- { " ", "a", "b", "c", "d" }
ngx_re.split("abcd", "()") -- { " ", " ", "a", "b", "c", "d" }
```

Should we do something about it?

agentzh:
@thibaultcha I think instead of checking empty patterns, we should check empty captures instead.

thibaultcha:
I'm not sure what you mean, nor how.

agentzh:
@thibaultcha Just check if the separator pattern actually matches any non-zero length of input data (by checking the `pos` returned by PCRE). If yes, simply move the cursor to the next character of the input.

thibaultcha:
@agentzh Sorry for the delay! I pushed a new version of the patch. Moving to the next character on empty matches (what the patch was already doing, but only for empty regexes) does not seem to be enough for cases like:
```bash
$ perl -e 'print join ":", split /^/m, "ab\ncd"'
ab
:cd
```

We now mimic this behavior in `ngx_re.split()` if the user specifies the `m` flag:
```lua
ngx_re.split("ab\ncd", "^", "m")
{ "ab\n", "cd" }
```

Let me know what you think of it!

PS: I don't really like this code duplication between the max and the non-max branches. It leads to an overly complicated function at first look, and needs twice as many tests for both branches :/

thibaultcha:
@agentzh Have you had some time to review this? Considering it is a bugfix, I think it needs some attention. As pointed out in my previous comment, I think we need to look-ahead when such empty matches happen, to know how many characters must be skipped (not necessary the next character as previously mentioned, implemented). Thanks!

lib/ngx/re.lua
```diff
@@ -168,6 +152,24 @@ function _M.split(subj, regex, opts, ctx, max, res)
                 break
             end
 
+            if from == to then
+                local old_pos = ctx.pos
+
+                local from2, _, _, err2 = re_split_helper(subj, compiled,
+                                              compile_once, flags, ctx)
```
agentzh: @thibaultcha It looks strange to me to perform a second match in the same location. Why? You said it is "the only way", but I still don't understand it. Will you elaborate?

thibaultcha: Sure, this needs elaboration. So, considering the following string:
```
"abcd\nefgh"
```

And the following regex:
```
/^/m
```

We have a match at characters `0` and `5`. If the only thing we do is moving to the next character after the first match, we have something like the following result:
```lua
{ "a", "bcd", "e", "fgh" }
```

Because our function is not smart enough to know that an empty match does not necessarily mean we are splitting char-by-char. What we need is keep the first match (`0`), and find where the next match is (we know it'll be empty too), which will be `5`. We can then deduce our first sub-string is `1 to 4`:
```lua
{ "abcd" }
```

Otherwise, a simple increment would have cut `0 to 1`, and then `2 to 4`:
```lua
{ "a", "bcd" }
```

Simply incrementing the string pos by one would work for regexps like `/()/` were we know the next character is the next match, but not for subjects were we don't know where the next empty match will be at. At least, that's what **seems** to be the only way (insisting on the "seems") :)

On a side note, for such use cases as we are describing, we still perform 1 PCRE call per token in the subject + the last call which results in NOMATCH, such as:
```lua
ngx_re.split("abc\ndef", "^", "m") -- matches twice + NOMATCH
ngx_re.split("abc,def", ",") -- matches once + NOMATCH
```

Since we move `ctx.pos` after the look-ahead on empty matches. So there is no performance overhead, simply a slightly different logic.

lib/ngx/re.lua
```diff
@@ -168,6 +152,24 @@ function _M.split(subj, regex, opts, ctx, max, res)
...
+                    return nil, err2
+                end
+
+                if not from2 then
+                    break
+                end
+
+                from = from2
+                to = from2
+                ctx.pos = old_pos
```
agentzh: Maybe we should not backtrack to the old position? This way we can make this extra match not get wasted. The same applies to the `to` index returned by this extra match.

lib/ngx/re.lua
agentzh: @thibaultcha It appears clearer to me now. Now I'd just like to be this match's result not wasted.

lib/ngx/re.lua
thibaultcha: Hmmm yeah I thought this was already the case (edited my explanatory comment), but just realized it actually is backtracking to the old position. Will try to get rid of it. Would want nothing less performant than that for lua-resty-core :)

thibaultcha:
@agentzh I have updated the patch with a new, smaller implementation that I find myself quite found of :) None of the existing tests were modified, some were added to cover additional or missing ground.

thibaultcha:
Also this patch was rebased on master.

agentzh:
@thibaultcha This indeed looks much better. Thanks!

@doujiang24 @dndx Will you review this PR when you guys have a chance? Thanks!

thibaultcha:
Ping :)

agentzh:
@thibaultcha Sorry for the delay! I'm still waiting for @doujiang24's review :)

thibaultcha:
Thanks @doujiang24 !

agentzh:
@thibaultcha Merged. Thanks!

PR #111, opened 2017-05-12
doc: moved directive `ssl_session_*` to `http_config`.

membphis:


agentzh:
@membphis Merged with the following commit log message:

```
doc: fixed the code example since directives `ssl_session_*_by_lua*` are no longer allowed in `server {}`.
```

agentzh:
Thanks!

PR #114, opened 2017-05-18
bugfix: set the `nil` at current index if there was no error log.

membphis:


t/errlog.t
```diff
@@ -882,3 +882,53 @@ log body:\d{4}/\d{2}/\d{2} \d{2}:\d{2}:\d{2} \[error\] (\d+).*access_by_lua\(ngi
 new line, client: 127.0.0.1, server: localhost, request: "GET /t HTTP/1.1", host: "localhost"
 \z
 --- skip_nginx: 2: <1.11.2
+
+
+
+=== TEST 22: ctx table
```
agentzh: This test title is too short and very confusing. What exactly is this test case's intention?

t/errlog.t
```diff
@@ -882,3 +882,53 @@ log body:\d{4}/\d{2}/\d{2} \d{2}:\d{2}:\d{2} \[error\] (\d+).*access_by_lua\(ngi
...
+            ngx.log(ngx.ERR, "-->\n", "new line")
+        }
+
+        content_by_lua_block {
+            local errlog = require "ngx.errlog"
+            local t = {}
+
+            for i = 1, 2 do
+                local res = errlog.get_logs(10, t)
+                ngx.say("maybe log lines: #", #res / 2)
```
agentzh: I"ve found the "maybe" word here a bit confusing.

membphis: when we tried to get error logs，reused the table `t`. the method `errlog.get_logs` will set `nil` when there was no error log.

in this case:
first time got data, the `t` value: `t = {4, "xxxx", nil}`.
second time didn't got data, the `t` value: `t= {nil, "xxxx", nil}`, the first `nil` is hole.

if we use `#` to get the table `t` length, it maybe wrong. that's why I wrote `maybe log lines`.

membphis:
@agentzh fixed, please review.

t/errlog.t
```diff
@@ -882,3 +882,53 @@ log body:\d{4}/\d{2}/\d{2} \d{2}:\d{2}:\d{2} \[error\] (\d+).*access_by_lua\(ngi
 new line, client: 127.0.0.1, server: localhost, request: "GET /t HTTP/1.1", host: "localhost"
 \z
 --- skip_nginx: 2: <1.11.2
+
+
+
+=== TEST 22: user-supplied Lua table to hold the result(with no error log)
```
agentzh: Please always put a space before `(`.

agentzh: @membphis I'm afraid this test title does not state the test case's intention accurately. Maybe we should be more specific here?

membphis: change the test case name :

```
=== TEST 22: user-supplied Lua table to hold the result (get one log + no log)
```

and fixed some other similar points.

membphis:
@agentzh all fixed, please check.

agentzh:
@membphis Changed the commit log message to the following:

```
bugfix: ngx.errlog: get_logs(): we did not set `nil` at the current table index when exhausting the existing error log data before reaching the `max` limit.
```

agentzh:
@membphis Merged. Thanks!

Issue #116, opened 2017-05-21
[Feature Proposal] New API to get Nginx log_level


spacewander:
It would be quite useful if we could get the configured `error_log` level via Lua code.
For instance, we could do something like this:
```lua
-- config.lua
_M.log_level = ngx.config.log_level()

-- req.lua
if log_level >= ngx.WARN then
    -- Now we save an `encode` operation if we configure error_log level to `error`.
    ngx.log(ngx.WARN, cjson.encode(obj), ...)
end
```
Currently, `ngx.log` handles the lua parameters only when the given level is higher than the configured one. And maybe we don't need to care about those log levels higher than `debug` in Lua land. So what we need to do is just extracting some existent code into a new function.

Here is a question: since we could implement this feature in a separate Nginx C module, will you still accept it as a part of OpenResty's feature list?

agentzh:
@spacewander We can do this in the `ngx.errlog` module of this library. The corresponding C API could be in the `ngx_lua` core. Pull requests are welcome! Thanks!

PR #117, opened 2017-05-22
feature: implemented the ngx.errlog.get_sys_filter_level()  API function to get system default log level

spacewander:
This pr is subsequent to #116 and depended on https://github.com/openresty/lua-nginx-module/pull/1079.

I'm glad to hear any further feedback.

to the authors of this project.

### DIFF
--- a/lib/ngx/errlog.lua
+++ b/lib/ngx/errlog.lua
@@ -21,6 +21,8 @@ int ngx_http_lua_ffi_errlog_set_filter_level(int level, unsigned char *err,
     size_t *errlen);
 int ngx_http_lua_ffi_errlog_get_msg(char **log, int *loglevel,
     unsigned char *err, size_t *errlen);
+
+int ngx_http_lua_ffi_errlog_log_level(ngx_http_request_t *r);
 ]]
 
 
@@ -84,6 +86,12 @@ function _M.get_logs(max, logs)
     end
 
     return logs
+end
+
+
+function _M.log_level()
+    local r = getfenv(0).__ngx_req
+    return C.ngx_http_lua_ffi_errlog_log_level(r)
 end
 
 
--- a/lib/ngx/errlog.lua
+++ b/lib/ngx/errlog.lua
@@ -22,7 +22,7 @@ int ngx_http_lua_ffi_errlog_set_filter_level(int level, unsigned char *err,
 int ngx_http_lua_ffi_errlog_get_msg(char **log, int *loglevel,
     unsigned char *err, size_t *errlen);
 
-int ngx_http_lua_ffi_errlog_log_level(ngx_http_request_t *r);
+int ngx_http_lua_ffi_errlog_get_sys_filtering_level(ngx_http_request_t *r);
 ]]
 
 
@@ -89,9 +89,9 @@ function _M.get_logs(max, logs)
 end
 
 
-function _M.log_level()
+function _M.get_sys_filtering_level()
     local r = getfenv(0).__ngx_req
-    return C.ngx_http_lua_ffi_errlog_log_level(r)
+    return C.ngx_http_lua_ffi_errlog_get_sys_filtering_level(r)
 end
 
 
--- a/lib/ngx/errlog.lua
+++ b/lib/ngx/errlog.lua
@@ -91,7 +91,7 @@ end
 
 function _M.get_sys_filtering_level()
     local r = getfenv(0).__ngx_req
-    return C.ngx_http_lua_ffi_errlog_get_sys_filtering_level(r)
+    return tonumber(C.ngx_http_lua_ffi_errlog_get_sys_filtering_level(r))
 end
 
 
--- a/lib/ngx/errlog.lua
+++ b/lib/ngx/errlog.lua
@@ -22,7 +22,7 @@ int ngx_http_lua_ffi_errlog_set_filter_level(int level, unsigned char *err,
 int ngx_http_lua_ffi_errlog_get_msg(char **log, int *loglevel,
     unsigned char *err, size_t *errlen);
 
-int ngx_http_lua_ffi_errlog_get_sys_filtering_level(ngx_http_request_t *r);
+int ngx_http_lua_ffi_errlog_get_sys_filter_level(ngx_http_request_t *r);
 ]]
 
 
@@ -89,9 +89,9 @@ function _M.get_logs(max, logs)
 end
 
 
-function _M.get_sys_filtering_level()
+function _M.get_sys_filter_level()
     local r = getfenv(0).__ngx_req
-    return tonumber(C.ngx_http_lua_ffi_errlog_get_sys_filtering_level(r))
+    return tonumber(C.ngx_http_lua_ffi_errlog_get_sys_filter_level(r))
 end
 
 
--- a/lib/ngx/errlog.md
+++ b/lib/ngx/errlog.md
@@ -221,16 +221,14 @@ Return system default filtering level as an integer. Later the return value
 could be used as Nginx log level constant. For example:
 
 ```lua
-init_by_lua_block {
-    local errlog = require "ngx.errlog"
-    local log_level = error_log.get_sys_filtering_level()
-    -- Now the filter level is always one level lower than system default log level
-    local status, err = errlog.set_filter_level(log_level - 1)
-    if not status then
-        ngx.log(ngx.ERR, err)
-        return
-    end
-}
+local errlog = require "ngx.errlog"
+local log_level = error_log.get_sys_filtering_level()
+-- Now the filter level is always a level higher than system default log level on priority
+local status, err = errlog.set_filter_level(log_level - 1)
+if not status then
+    ngx.log(ngx.ERR, err)
+    return
+end
 ```
 
 [Back to TOC](#table-of-contents)
--- a/lib/ngx/errlog.md
+++ b/lib/ngx/errlog.md
@@ -222,7 +222,7 @@ value could be used as an Nginx log level constant. For example:
 
 ```lua
 local errlog = require "ngx.errlog"
-local log_level = error_log.get_sys_filter_level()
+local log_level = errlog.get_sys_filter_level()
 -- Now the filter level is always one level higher than system default log level on priority
 local status, err = errlog.set_filter_level(log_level - 1)
 if not status then
--- a/lib/ngx/errlog.md
+++ b/lib/ngx/errlog.md
@@ -13,7 +13,7 @@ Table of Contents
 * [Methods](#methods)
     * [set_filter_level](#set_filter_level)
     * [get_logs](#get_logs)
-    * [get_sys_filtering_level](#get_sys_filtering_level)
+    * [get_sys_filter_level](#get_sys_filter_level)
 * [Community](#community)
     * [English Mailing List](#english-mailing-list)
     * [Chinese Mailing List](#chinese-mailing-list)
@@ -211,18 +211,18 @@ clear the table yourself before feeding it into the `errlog.get_logs` function.
 
 [Back to TOC](#table-of-contents)
 
-get_sys_filtering_level
------------------------
-**syntax:** *log_level = log_module.get_sys_filtering_level()*
+get_sys_filter_level
+--------------------
+**syntax:** *log_level = log_module.get_sys_filter_level()*
 
 **context:** *any*
 
-Return the system's default filtering level as an integer. Later, the returned
+Return the system's default filter level as an integer. Later, the returned
 value could be used as an Nginx log level constant. For example:
 
 ```lua
 local errlog = require "ngx.errlog"
-local log_level = error_log.get_sys_filtering_level()
+local log_level = error_log.get_sys_filter_level()
 -- Now the filter level is always one level higher than system default log level on priority
 local status, err = errlog.set_filter_level(log_level - 1)
 if not status then
--- a/lib/ngx/errlog.md
+++ b/lib/ngx/errlog.md
@@ -221,16 +221,16 @@ Return system default filtering level as an integer. Later the return value
 could be used as Nginx log level constant. For example:
 
 ```lua
--- config.lua
-local errlog = require "ngx.errlog"
-_M.log_level = error_log.get_sys_filtering_level()
-return _M
-```
-
-```lua
--- req.lua
-local log_level = config.log_level
-errlog.set_filter_level(log_level)
+init_by_lua_block {
+    local errlog = require "ngx.errlog"
+    local log_level = error_log.get_sys_filtering_level()
+    -- Now the filter level is always one level lower than system default log level
+    local status, err = errlog.set_filter_level(log_level - 1)
+    if not status then
+        ngx.log(ngx.ERR, err)
+        return
+    end
+}
 ```
 
 [Back to TOC](#table-of-contents)
--- a/lib/ngx/errlog.md
+++ b/lib/ngx/errlog.md
@@ -217,13 +217,13 @@ get_sys_filtering_level
 
 **context:** *any*
 
-Return system default filtering level as an integer. Later the return value
-could be used as Nginx log level constant. For example:
+Return the system's default filtering level as an integer. Later, the returned
+value could be used as an Nginx log level constant. For example:
 
 ```lua
 local errlog = require "ngx.errlog"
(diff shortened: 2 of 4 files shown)


lib/ngx/errlog.lua
```diff
@@ -21,6 +21,8 @@ int ngx_http_lua_ffi_errlog_set_filter_level(int level, unsigned char *err,
     size_t *errlen);
 int ngx_http_lua_ffi_errlog_get_msg(char **log, int *loglevel,
     unsigned char *err, size_t *errlen);
+
+int ngx_http_lua_ffi_errlog_log_level(ngx_http_request_t *r);
```
agentzh: Better use verbs in the function names, like `ngx_http_lua_ffi_errlog_get_sys_filtering_level`. Also, we should emphasize that this is the system default filtering level since we already have the `set_filter_level` function to set a "soft filtering level".

lib/ngx/errlog.md
```diff
@@ -210,6 +211,33 @@ clear the table yourself before feeding it into the `errlog.get_logs` function.
 
 [Back to TOC](#table-of-contents)
 
+log_level
+--------
```
agentzh: The dashes should be at least as many as the characters in the previous line.

lib/ngx/errlog.md
```diff
@@ -210,6 +211,33 @@ clear the table yourself before feeding it into the `errlog.get_logs` function.
 
 [Back to TOC](#table-of-contents)
 
+log_level
```
agentzh: How about `get_sys_filtering_level`?

spacewander:
@agentzh 
Just updated the code. Would you have time to review it?

lib/ngx/errlog.md
```diff
@@ -210,6 +211,30 @@ clear the table yourself before feeding it into the `errlog.get_logs` function.
...
+-- config.lua
+local errlog = require "ngx.errlog"
+_M.log_level = error_log.get_sys_filtering_level()
+return _M
+```
+
+```lua
+-- req.lua
+local log_level = config.log_level
+errlog.set_filter_level(log_level)
```
agentzh: Why set the sys filter level to the soft level for our capturing facility? This is a no-op since the system filtering level is mandatory anyway.

We need a simpler and more meaningful code example here. The current code example is confusing.

spacewander:
@agentzh 
Could you review the new code example? Is it OK?

lib/ngx/errlog.md
```diff
@@ -210,6 +211,30 @@ clear the table yourself before feeding it into the `errlog.get_logs` function.
...
+-----------------------
+**syntax:** *log_level = log_module.get_sys_filtering_level()*
+
+**context:** *any*
+
+Return system default filtering level as an integer. Later the return value
+could be used as Nginx log level constant. For example:
+
+```lua
+init_by_lua_block {
```
agentzh: We can omit the directive here. Since it does not really matter which handler it is in. Besides, the code block marks  as the`lua` language instead of the `nginx` language.

lib/ngx/errlog.md
```diff
@@ -210,6 +211,30 @@ clear the table yourself before feeding it into the `errlog.get_logs` function.
...
+    local errlog = require "ngx.errlog"
+    local log_level = error_log.get_sys_filtering_level()
+    -- Now the filter level is always one level lower than system default log level
```
agentzh: This is wrong I'm afraid. The smaller the log level values, the *higher* the priority of the level. See

```
#define NGX_LOG_STDERR            0
#define NGX_LOG_EMERG             1
#define NGX_LOG_ALERT             2
#define NGX_LOG_CRIT              3
#define NGX_LOG_ERR               4
#define NGX_LOG_WARN              5
#define NGX_LOG_NOTICE            6
#define NGX_LOG_INFO              7
#define NGX_LOG_DEBUG             8
```

So when you decrement the level, you actually make it one level *higher* instead of one level lower.

spacewander: @agentzh 
Actually I want to say `one level lower` by its integer value...
Obviously my former statement is ambiguous.

spacewander:
@agentzh 
Could you spend some time reviewing the latest change? Thanks!

lib/ngx/errlog.lua
```diff
@@ -87,4 +89,10 @@ function _M.get_logs(max, logs)
 end
 
 
+function _M.get_sys_filtering_level()
+    local r = getfenv(0).__ngx_req
+    return C.ngx_http_lua_ffi_errlog_get_sys_filtering_level(r)
```
agentzh: We should not return the cdata value directly without converting it to a number via the `tonumber()` builtin.

spacewander:
@agentzh
Fixed it!

lib/ngx/errlog.md
```diff
@@ -210,6 +211,28 @@ clear the table yourself before feeding it into the `errlog.get_logs` function.
...
+
+**context:** *any*
+
+Return system default filtering level as an integer. Later the return value
+could be used as Nginx log level constant. For example:
+
+```lua
+local errlog = require "ngx.errlog"
+local log_level = error_log.get_sys_filtering_level()
+-- Now the filter level is always a level higher than system default log level on priority
```
agentzh: "a level higher" => "one level higher". The latter reads better.

agentzh:
@thibaultcha @bungle @doujiang24 @membphis What do you think of this PR please? Thanks!

t/errlog.t
```diff
@@ -932,3 +932,66 @@ maybe log lines: #1
...
+--- config
+    location /t {
+        content_by_lua_block {
+            local log_level = package.loaded.log_level
+
+            if log_level >= ngx.WARN then
+                ngx.log(ngx.WARN, "log a warning event")
+            end
+            if log_level >= ngx.NOTICE then
+                ngx.log(ngx.NOTICE, "log a notice event")
```
thibaultcha: Because this is logged with `ngx.NOTICE`, it will never get logged anyways, even if the surrounding condition fails. I think it should rather be:
```lua
if log_level >= ngx.NOTICE then
    ngx.log(ngx.WARN, "do not log another warning event")
end
```

This way, we make sure that the logging was not written due to the surrounding condition, and not due to the `ngx.log` logging level.

lib/ngx/errlog.md
```diff
@@ -210,6 +211,28 @@ clear the table yourself before feeding it into the `errlog.get_logs` function.
...
 [Back to TOC](#table-of-contents)
 
+get_sys_filtering_level
+-----------------------
+**syntax:** *log_level = log_module.get_sys_filtering_level()*
+
+**context:** *any*
+
+Return system default filtering level as an integer. Later the return value
+could be used as Nginx log level constant. For example:
```
thibaultcha: If I may, I think this would read better as well:
```
Return the system's default filtering level as an integer. Later, the returned 
value could be used as an Nginx log level constant. For example:
```

spacewander:
@agentzh @thibaultcha 
Just updated the doc and test. Would you have time to review this pr again?

lib/ngx/errlog.md
```diff
@@ -13,6 +13,7 @@ Table of Contents
 * [Methods](#methods)
     * [set_filter_level](#set_filter_level)
     * [get_logs](#get_logs)
+    * [get_sys_filtering_level](#get_sys_filtering_level)
```
agentzh: I think we should use `get_sys_filter_level` instead of `get_sys_filtering_level` to be more consistent with the existing `set_filter_level` API function.

thibaultcha: Had the same thought :+1:

spacewander:
@agentzh @thibaultcha 
Already renamed `get_sys_filtering_level` to `get_sys_filter_level`.
Have time to review the newest change?

lib/ngx/errlog.md
```diff
@@ -210,6 +211,28 @@ clear the table yourself before feeding it into the `errlog.get_logs` function.
 
 [Back to TOC](#table-of-contents)
 
+get_sys_filter_level
+-----------------------
```
thibaultcha: (there are 3 extra dashes here now)

spacewander: Good catch! Fixed in a new commit.

lib/ngx/errlog.md
```diff
@@ -210,6 +211,28 @@ clear the table yourself before feeding it into the `errlog.get_logs` function.
...
+**syntax:** *log_level = log_module.get_sys_filter_level()*
+
+**context:** *any*
+
+Return the system's default filter level as an integer. Later, the returned
+value could be used as an Nginx log level constant. For example:
+
+```lua
+local errlog = require "ngx.errlog"
+local log_level = error_log.get_sys_filter_level()
```
membphis: it should be `errlog`?

spacewander: Corrected it.

lib/ngx/errlog.md
```diff
@@ -210,6 +211,28 @@ clear the table yourself before feeding it into the `errlog.get_logs` function.
...
 [Back to TOC](#table-of-contents)
 
+get_sys_filter_level
+--------------------
+**syntax:** *log_level = log_module.get_sys_filter_level()*
+
+**context:** *any*
+
+Return the system's default filter level as an integer. Later, the returned
+value could be used as an Nginx log level constant. For example:
```
moonming: better use `NGINX` instead of `Nginx`.

spacewander: Maybe we could keep `Nginx` here, because the `set_filter_level` already used `Nginx log level constants` in its document. Just follow existing style.

agentzh: @spacewander That's fine.

t/errlog.t
```diff
@@ -932,3 +932,65 @@ maybe log lines: #1
 end
 \z
 --- skip_nginx: 2: <1.11.2
+
+
+
+=== TEST 23: the system default filter level is "debug"
```
membphis: @spacewander I think we need some test cases for different phases. 

for example `init_by_lua*`, need to cover https://github.com/openresty/lua-nginx-module/pull/1079/files#diff-7caeff0fc42c25ea398bef6b2257cfc7R419

spacewander: @membphis 
Yes, this branch is already covered with `TEST 25: get system default filter level during Nginx worker starts`.

membphis: `init_by_lua*` is special, https://github.com/openresty/lua-nginx-module/blob/master/src/ngx_http_lua_phase.c#L33

I think we'd better add a test case. @spacewander

spacewander: @membphis 
I see. Now I change `TEST 25` to test `get_sys_filter_level` under `init_by_lua*`.
Is it ok?

t/errlog.t
```diff
@@ -9,21 +9,32 @@ log_level('error');
...
     my $block = shift;
 
+    my $test_name = $block->name();
     my $http_config = $block->http_config || '';
-    $http_config .= <<_EOC_;
 
-    lua_package_path "$pwd/lib/?.lua;../lua-resty-lrucache/lib/?.lua;;";
-    init_by_lua_block {
-        require "resty.core"
+    if (index($test_name, "TEST 25: ") != -1) {
```
membphis: Personally I do not like this hard code, how about add a new separate test file to handle this? 
Just a suggestion. :)

spacewander: @membphis 
A separate test file for a single test seems wasteful...
I am not familiar with `test-nginx`. Is there any pretty solution?

membphis: I don't have better way  :(

A note: the test name is unstable, if some guy added a test case before "TEST 25", this case will not work as we expect. 

here is an example: https://github.com/openresty/lua-nginx-module/blob/master/t/134-worker-count-5.t

t/errlog.t
```diff
@@ -981,12 +973,11 @@ Is "emerg" the system default filter level? true
 
 
 === TEST 25: get system default filter level during Nginx starts
---- http_config
-    init_by_lua_block {
-        require "resty.core"
-        local errlog = require "ngx.errlog"
-        package.loaded.log_level = errlog.get_sys_filter_level()
-    }
+--- init_by_lua_block
```
membphis: @spacewander nice job 👍

lib/ngx/errlog.lua
```diff
@@ -87,4 +89,10 @@ function _M.get_logs(max, logs)
 end
 
 
+function _M.get_sys_filter_level()
+    local r = getfenv(0).__ngx_req
+    return tonumber(C.ngx_http_lua_ffi_errlog_get_sys_filter_level(r))
```
agentzh: This function may on the hot Lua code paths and we should avoid use of the Lua global variables `getfenv` and `tonumber` in this Lua function body. Better cache their values local variables in an upper-level scope.

agentzh:
@spacewander Merged with the following minor extra patch. Thanks!

```diff
diff --git a/.travis.yml b/.travis.yml
index 34a1c6b..3f5cf76 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,7 +56,7 @@ install:
   - git clone https://github.com/openresty/openresty.git ../openresty
   - git clone https://github.com/openresty/openresty-devel-utils.git
   - git clone https://github.com/simpl/ngx_devel_kit.git ../ndk-nginx-module
-  - git clone -b errlog_log_level https://github.com/spacewander/lua-nginx-module.git ../lua-nginx-module
+  - git clone https://github.com/openresty/lua-nginx-module.git ../lua-nginx-module
   - git clone https://github.com/openresty/no-pool-nginx.git ../no-pool-nginx
   - git clone https://github.com/openresty/echo-nginx-module.git ../echo-nginx-module
   - git clone https://github.com/openresty/lua-resty-lrucache.git
diff --git a/lib/ngx/errlog.lua b/lib/ngx/errlog.lua
index 12482a7..1b4c725 100644
--- a/lib/ngx/errlog.lua
+++ b/lib/ngx/errlog.lua
@@ -11,6 +11,8 @@ local new_tab = base.new_tab
 local ffi_new = ffi.new
 local charpp = ffi_new("char *[1]")
 local intp = ffi.new("int[1]")
+local getfenv = getfenv
+local tonumber = tonumber
 
 
 local _M = { version = base.version }
diff --git a/lib/ngx/errlog.md b/lib/ngx/errlog.md
index 5c59198..192de66 100644
--- a/lib/ngx/errlog.md
+++ b/lib/ngx/errlog.md
@@ -217,8 +217,13 @@ get_sys_filter_level
 
 **context:** *any*
 
-Return the system's default filter level as an integer. Later, the returned
-value could be used as an Nginx log level constant. For example:
+Return the nginx core's error log filter level (defined via the [error_log](http://nginx.org/r/error_log)
+configuration directive in `nginx.conf`) as an integer value matching the nginx error log level
+constants documented below:
+
+https://github.com/openresty/lua-nginx-module/#nginx-log-level-constants
+
+For example:
 
 ```lua
 local errlog = require "ngx.errlog"
diff --git a/t/errlog.t b/t/errlog.t
index e961e85..04b197e 100644
--- a/t/errlog.t
+++ b/t/errlog.t
@@ -9,7 +9,7 @@ log_level('error');
 
 repeat_each(2);
 
-plan tests => repeat_each() * (blocks() * 2 + 12);
+plan tests => repeat_each() * (blocks() * 2 + 13);
 
 my $pwd = cwd();
 
@@ -972,7 +972,7 @@ Is "emerg" the system default filter level? true
 
 
 
-=== TEST 25: get system default filter level during Nginx starts
+=== TEST 25: get system default filter level during Nginx starts (init)
 --- init_by_lua_block
     require "resty.core"
     local errlog = require "ngx.errlog"
@@ -997,3 +997,31 @@ GET /t
 log a warning event
 --- no_error_log
 do not log another warning event
+
+
+
+=== TEST 26: get system default filter level during Nginx worker starts (init worker)
+--- http_config
+    init_worker_by_lua_block {
+        local errlog = require "ngx.errlog"
+        package.loaded.log_level = errlog.get_sys_filter_level()
+    }
+--- config
+    location /t {
+            local log_level = package.loaded.log_level
+
+            if log_level >= ngx.WARN then
+                ngx.log(ngx.WARN, "log a warning event")
+            else
+                ngx.log(ngx.WARN, "do not log another warning event")
+            end
+        }
+    }
+--- log_level: warn
+--- request
+GET /t
+--- error_log
+log a warning event
+--- no_error_log
+do not log another warning event
```

PR #122, opened 2017-06-22
bugfix: convert regex to string before get its length

spacewander:
It seems to be a miss of a previous bugfix.

When we execute `ngx.re.match("1234", 123, "jo")`:
With lua-resty-core, there would be a `attempt to get length of local 'regex' (a number value)` error.
Without lua-resty-core, there would be `123` and everything is ok.

The same with `ngx.re.sub("1234", 123, 456, "jo")`.

This pr eliminates the difference and makes the behavior consistent.

to the authors of this project.

### DIFF
--- a/lib/resty/core/regex.lua
+++ b/lib/resty/core/regex.lua
@@ -342,6 +342,7 @@ local function re_match_compile(regex, opts)
         -- print("compiled regex not found, compiling regex...")
         local errbuf = get_string_buf(MAX_ERR_MSG_LEN)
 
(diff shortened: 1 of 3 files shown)


agentzh:
I don't think it's a good idea to to the automatic stringification for the regex argument. This may make misuses of the API (i.e., feeding bad-typed arguments) very hard to detect and/or debug.

agentzh:
@spacewander We'll abandon the CFunction-based Lua API in `ngx_lua` pretty soon anyway.

PR #124, opened 2017-06-27
feature: supported to fetch the log time.

membphis:
TODO: docs

### DIFF
--- a/lib/ngx/errlog.lua
+++ b/lib/ngx/errlog.lua
@@ -11,6 +11,7 @@ local new_tab = base.new_tab
 local ffi_new = ffi.new
 local charpp = ffi_new("char *[1]")
 local intp = ffi.new("int[1]")
+local num_value = ffi_new("double[1]")
 local getfenv = getfenv
 local tonumber = tonumber
 
@@ -22,7 +23,7 @@ ffi.cdef[[
 int ngx_http_lua_ffi_errlog_set_filter_level(int level, unsigned char *err,
     size_t *errlen);
 int ngx_http_lua_ffi_errlog_get_msg(char **log, int *loglevel,
-    unsigned char *err, size_t *errlen);
+    unsigned char *err, size_t *errlen, double *log_time);
 
 int ngx_http_lua_ffi_errlog_get_sys_filter_level(ngx_http_request_t *r);
 ]]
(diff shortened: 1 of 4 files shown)


lib/ngx/errlog.lua
```diff
@@ -50,39 +51,51 @@ function _M.set_filter_level(level)
 end
 
 
-function _M.get_logs(max, logs)
+function _M.get_logs(max, logs, opts)
+    local is_fetch_time = opts and opts.fetch_time
```
agentzh: I think we can simply return the UNIX time unconditionally. This Lua API has never included in any releases, it's fine for us to drop backward compatibility here. (Well, even if we release this code, we can still change the API without notice since this API is marked experimental in the docs).

membphis: got it, fix it later

membphis:
@agentzh please take a look when you have time :)

t/errlog.t
```diff
@@ -1025,3 +1023,91 @@ GET /t
...
+            local res, err = errlog.get_logs(nil, nil, {fetch_time = true})
+            if not res then
+                error("FAILED " .. err)
+            end
+            ngx.say("log lines: ", #res / 3)
+
+            for i = 1, #res, 3 do
+                ngx.say("log level: ", res[i])
+                ngx.say("log body: ",  res[i + 1])
+                ngx.say("log time: ", res[i + 2])
```
agentzh: I think the time field should be before the body field?

membphis:
@agentzh all fixed, please take look :)

agentzh:
@membphis Merged with the following extra patch:

```diff
diff --git a/.travis.yml b/.travis.yml
index ed0fe7d..3f5cf76 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,7 +56,7 @@ install:
   - git clone https://github.com/openresty/openresty.git ../openresty
   - git clone https://github.com/openresty/openresty-devel-utils.git
   - git clone https://github.com/simpl/ngx_devel_kit.git ../ndk-nginx-module
-  - git clone -b log-time https://github.com/membphis/lua-nginx-module.git ../lua-nginx-module
+  - git clone https://github.com/openresty/lua-nginx-module.git ../lua-nginx-module
   - git clone https://github.com/openresty/no-pool-nginx.git ../no-pool-nginx
   - git clone https://github.com/openresty/echo-nginx-module.git ../echo-nginx-module
   - git clone https://github.com/openresty/lua-resty-lrucache.git
diff --git a/lib/ngx/errlog.md b/lib/ngx/errlog.md
index 7513131..452ed1c 100644
--- a/lib/ngx/errlog.md
+++ b/lib/ngx/errlog.md
@@ -161,6 +161,16 @@ The resulting table has the following structure:
 { level1, time1, msg1, level2, time2, msg2, ... }


+The `levelX` values are constants defined below:
+
+https://github.com/openresty/lua-nginx-module/#nginx-log-level-constants
+
+The `timeX` values are UNIX timestamps in seconds with millisecond precision. The sub-second part is presented as the decimal part.
+The time format is exactly the same as the value returned by [ngx.now](https://github.com/openresty/lua-nginx-module/#ngxnow). It is
+also subject to NGINX core's time caching.
+
+The `msgX` values are the error log message texts.
+
 So to traverse this array, the user can use a loop like this:

```

PR #125, opened 2017-06-27
tests: added test cases for error log ringbuf bugs

pracj3am:
There are several bugs in `ngx_http_lua_log_ringbuf.c` addressed by these test cases. Test case 31 demonstrates a buffer over-read.

### DIFF
--- a/t/errlog.t
+++ b/t/errlog.t
@@ -1111,3 +1111,123 @@ enter 11
 "
 ]
 --- skip_nginx: 3: <1.11.2
+
+
+
+
+=== TEST 29: ringbuf overflow bug
+--- http_config
+    lua_capture_error_log 4k;
+--- config
+    location /t {
+        access_by_lua_block {
+            local errlog = require "ngx.errlog"
+            local msg = string.rep("*", 10)
+
+            for i = 1, 10 do
+                ngx.log(ngx.ERR, msg)
+            end
+        }
+
+        content_by_lua_block {
+            local errlog = require "ngx.errlog"
+            local msg = string.rep("*", 10)
+
+            for i = 1, 40 do
+                local res = errlog.get_logs(1)
+                if res and #res then
+                    ngx.log(ngx.ERR, msg)
+                end
+            end
+
+            local res = errlog.get_logs(10)
+            ngx.say("log lines: #", #res / 3)
+        }
+    }
+--- log_level: info
+--- request
+GET /t
+--- response_body
+log lines: #10
+--- skip_nginx: 2: <1.11.2
+
+
+
+=== TEST 30: ringbuf sentinel bug1
+--- http_config
+    lua_capture_error_log 4k;
+--- config
+    location /t {
+        access_by_lua_block {
+            local errlog = require "ngx.errlog"
+            local msg = string.rep("a", 20)
+            local bigmsg = string.rep("A", 3000)
+
+            for i = 1, 10 do
+                ngx.log(ngx.ERR, msg)
+            end
+            ngx.log(ngx.ERR, bigmsg)
+            ngx.log(ngx.ERR, msg)
+        }
+
+        content_by_lua_block {
+            local errlog = require "ngx.errlog"
+
+            local res = errlog.get_logs(2)
+            ngx.say("log lines: #", #res / 3)
+
+            for i = 1, #res, 3 do
+                ngx.say(string.gsub(res[i + 2], "^.*([Aa][Aa][Aa]).*$", "%1"), "")
+            end
+        }
+    }
+--- log_level: info
+--- request
+GET /t
+--- response_body
+log lines: #2
+AAA
+aaa
+--- skip_nginx: 2: <1.11.2
+
+
+
+=== TEST 31: ringbuf sentinel bug2
+--- http_config
+    lua_capture_error_log 4k;
+--- config
+    location /t {
+        access_by_lua_block {
+            local errlog = require "ngx.errlog"
+            local msg = string.rep("a", 20)
+
+            for i = 1, 20 do
+                ngx.log(ngx.ERR, msg)
+            end
+        }
+
+        content_by_lua_block {
+            local errlog = require "ngx.errlog"
+            local msg = string.rep("a", 20)
+
+            local res = errlog.get_logs(18)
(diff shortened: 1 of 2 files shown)


agentzh:
@pracj3am Seems like travis ci testing is broken on your branch?

https://travis-ci.org/openresty/lua-resty-core/builds/247623544?utm_source=github_status&utm_medium=notification

The test suite is broken for the nginx 1.11.2 core.

pracj3am:
That's right. I added tests to show bugs in lua nginx module.

agentzh:
@pracj3am Will you use your own branch of lua-nginx-module in `.travis.yml` to ensure your new fixes indeed pass all the tests?

pracj3am:
Good point, I will.

pracj3am:
And it's done :)

agentzh:
@pracj3am Thanks!

@membphis Will you please review this PR? Thanks!

t/errlog.t
```diff
@@ -1025,3 +1025,122 @@ GET /t
...
+
+            for i = 1, 40 do
+                local res = errlog.get_logs(1)
+                if res and #res then
+                    ngx.log(ngx.ERR, msg)
+                end
+            end
+
+            local res = errlog.get_logs(10)
+            ngx.say("log lines: #", #res / 2)
```
membphis: The resulting table has the following structure: 
`{ level1, time1, msg1, level2, time2, msg2, ... }`

https://github.com/openresty/lua-resty-core/blob/master/lib/ngx/errlog.md#get_logs

we should use `#res / 3` now.

pracj3am:
It had not yesterday :)

membphis:
@pracj3am please rebase your PR

pracj3am:
It is rebased.

t/errlog.t
```diff
@@ -1111,3 +1111,123 @@ enter 11
...
+            ngx.say("log lines: #", #res / 3)
+
+            for i = 1, 18 do
+                ngx.log(ngx.ERR, msg)
+            end
+
+            local bigmsg = string.rep("A", 2000)
+            ngx.log(ngx.ERR, bigmsg)
+
+            local res = errlog.get_logs(2)
```
membphis: @pracj3am we can get all of the logs in here, more meaningful

`local res = errlog.get_logs()`

t/errlog.t
```diff
@@ -1111,3 +1111,123 @@ enter 11
...
+            local msg = string.rep("*", 10)
+
+            for i = 1, 40 do
+                local res = errlog.get_logs(1)
+                if res and #res then
+                    ngx.log(ngx.ERR, msg)
+                end
+            end
+
+            local res = errlog.get_logs(10)
```
membphis: we can get all of the logs in here, more meaningful too

pracj3am: Okay

t/errlog.t
```diff
@@ -1111,3 +1111,123 @@ enter 11
...
+=== TEST 29: ringbuf overflow bug
+--- http_config
+    lua_capture_error_log 4k;
+--- config
+    location /t {
+        access_by_lua_block {
+            local errlog = require "ngx.errlog"
+            local msg = string.rep("*", 10)
+
+            for i = 1, 10 do
```
membphis: How about we only add a small amount of error log here? we will get all the error logs finally(small amount). 

This is more convenient to check the contents of the log more stringent, like this: https://github.com/openresty/lua-resty-core/blob/master/t/errlog.t#L615

membphis: It is not enough to check only the number of error log rows.

t/errlog.t
pracj3am: In this case I don't want all the log, because I don't know how many was threw away (it depends on how many characters are added by nginx), or do I? Goal of this test is to show, that reading the second log causes buffer over-read.

t/errlog.t
pracj3am: Goal of this test was to show, that messages disappear even if there is enough space for them, that's why I check just number. To show it there should be enough messages to fill the buffer (so it could be smaller amount of bigger messages). Then read a single message and then write a single message.

t/errlog.t
membphis: the size of `lua_capture_error_log ` is `4k`, and we only collected `error` level log in this case.

I think we can get a fixed number of logs.

We can remove the code, I think this test case can also be passed. But that's not what we expected.
```
local bigmsg = string.rep("A", 2000)
ngx.log(ngx.ERR, bigmsg)
```

pracj3am: I don't understand what do you mean by the last paragraph.

t/errlog.t
membphis: We need to confirm that the contents of the log are correct :)

pracj3am: Okay

t/errlog.t
membphis: you can make a try with this code in this test case, I think this test case can also be passed. 

```nginx
        content_by_lua_block {
            local errlog = require "ngx.errlog"
            local msg = string.rep("a", 20)

            local res = errlog.get_logs(18)
            ngx.say("log lines: #", #res / 3)

            for i = 1, 18 do
                ngx.log(ngx.ERR, msg)
            end

            -- local bigmsg = string.rep("A", 2000)
            -- ngx.log(ngx.ERR, bigmsg)

            local res = errlog.get_logs(2)
            ngx.say("log lines: #", #res / 3)
        }
```

if we fetch all error logs at the end of scope, it's more strict.

```nginx
        content_by_lua_block {
            local errlog = require "ngx.errlog"
            local msg = string.rep("a", 20)

            local res = errlog.get_logs(18)
            ngx.say("log lines: #", #res / 3)

            for i = 1, 18 do
                ngx.log(ngx.ERR, msg)
            end

            -- local bigmsg = string.rep("A", 2000)
            -- ngx.log(ngx.ERR, bigmsg)

            -- get all error logs
            local res = errlog.get_logs()
            ngx.say("log lines: #", #res / 3)
        }
```

agentzh:
@pracj3am This PR is superseded by #127, which has already been merged. Thanks!

pracj3am:
Okay, thanks!

PR #126, opened 2017-07-03
change: ngx.errlog: a mistake in readme

yanxurui:
it should be 30 instead of 20


agentzh:
@yanxurui Merged. Thanks!

PR #129, opened 2017-07-09
test: mark fractional part as optional to avoid unexpected failure

spacewander:
If the fractional part of a number is zero only, ngx.say will eliminate it.
For example, the output of ngx.say(1499575361.0) is 1499575361.

### DIFF
--- a/t/errlog.t
+++ b/t/errlog.t
@@ -1092,11 +1092,11 @@ GET /t
 --- response_body_like chomp
 \Alog lines: 2
 log level: 3
-log time: \d+\.\d+
+log time: \d+(?:\.\d+)?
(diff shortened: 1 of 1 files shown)


t/errlog.t
```diff
@@ -1092,11 +1092,11 @@ GET /t
 --- response_body_like chomp
 \Alog lines: 2
 log level: 3
-log time: \d+\.\d+
+log time: \d+(\.\d+)?
```
agentzh: I think we should use `(?:)` instead of `()` to avoid useless submatch capturing?

spacewander:
@agentzh 
OK, I have changed it.

agentzh:
@spacewander Merged. Thanks!

PR #131, opened 2017-07-10
test: added test cases, cover the semaphore was used in `ngx.timer`.

membphis:
this PR depends on: 

https://github.com/openresty/lua-nginx-module/pull/1109
https://github.com/openresty/test-nginx/pull/58


agentzh:
@membphis I think you can just add these tests to lua-nginx-module? This is because lua-nginx-module's test suite already assumes the presence of lua-resty-core.

membphis:
will move the test case to `lua-nginx-module`.

PR #134, opened 2017-08-01
doc: added a performance note for set_cert & set_priv_key 

filipslavik:
Hi,

I added a simple note to the `set_cert` and `set_priv_key` functions regarding the performance benefit they provide if compared to the  `set_der_cert` and `set_der_priv_key` variants.

(more info [here](https://groups.google.com/forum/#!topic/openresty-en/VEXG8w0sbp0))

Thanks!

### DIFF
--- a/lib/ngx/ssl.md
+++ b/lib/ngx/ssl.md
@@ -369,7 +369,7 @@ Sets the SSL certificate chain opaque pointer returned by the
 
 Returns `true` on success, or a `nil` value and a string describing the error otherwise.
 
-Note that the `set_cert` function will perform slightly better, in terms of CPU cycles wasted, than then
+Note that the `set_cert` function will perform slightly better, in terms of CPU cycles wasted, than the
 [set_der_cert](#set_der_cert) variant, since the first function uses opaque cdata pointers
 which do not require any additional conversion needed to be performed by the NGINX core during the SSL session
 negotiation.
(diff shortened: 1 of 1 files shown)


lib/ngx/ssl.md
```diff
@@ -369,6 +369,11 @@ Sets the SSL certificate chain opaque pointer returned by the
 
 Returns `true` on success, or a `nil` value and a string describing the error otherwise.
 
+Note that the `set_cert` function will perform slightly better, in terms of CPU cycles wasted, than then
```
agentzh: Typo: "than then" => "than the".

lib/ngx/ssl.md
```diff
@@ -369,6 +369,11 @@ Sets the SSL certificate chain opaque pointer returned by the
...
+[set_der_cert](#set_der_cert) variant, since the first function uses opaque cdata pointers
+which do not require any additional conversion needed to be performed by the NGINX core during the SSL session
```
agentzh: Actually the additional conversion is done by OpenSSL instead of the NGINX core. Better be more accurate here.

lib/ngx/ssl.md
filipslavik: Fixed, thank you!

lib/ngx/ssl.md
filipslavik: Yes, you are absolutely right. Actually, my first commit at bb1302ebcf3bc6e1543eb587c0ac3b819a392d83 
mentioned that the conversion is being made in OpenSSL, but I thought it was too detailed. 

Anyway, I suggest the following change :

` .... since the first function uses opaque cdata pointers which do not require any 
additional conversion needed to be performed by the NGINX core during the SSL 
session negotiation.`

will be replaced with :

` .... since the first function uses opaque cdata pointers which do not require any 
additional conversion to be performed by OpenSSL, while the NGINX core tries to
complete the session negotiation.`

What do you think ?

agentzh:
@filipslavik Merged with minor tweaks. Thanks!

PR #139, opened 2017-08-23
Ngx ssl add client addr function

smartwjw:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -79,6 +79,7 @@ script:
   - make -j$JOBS > build.log 2>&1 || (cat build.log && exit 1)
   - sudo make PATH=$PATH install_sw > build.log 2>&1 || (cat build.log && exit 1)
   - cd ../mockeagain/ && make CC=$CC -j$JOBS && cd ..
+  - cd ../lua-nginx-module && curl https://patch-diff.githubusercontent.com/raw/openresty/lua-nginx-module/pull/1145.patch | git apply && cd ..
   - tar zxf download-cache/pcre-$PCRE_VER.tar.gz
   - export PATH=$PWD/work/nginx/sbin:$PWD/openresty-devel-utils:$PATH
   - export LD_PRELOAD=$PWD/mockeagain/mockeagain.so
--- a/.travis.yml
+++ b/.travis.yml
@@ -79,7 +79,7 @@ script:
   - make -j$JOBS > build.log 2>&1 || (cat build.log && exit 1)
   - sudo make PATH=$PATH install_sw > build.log 2>&1 || (cat build.log && exit 1)
   - cd ../mockeagain/ && make CC=$CC -j$JOBS && cd ..
-  - cd ../lua-nginx-module && curl https://patch-diff.githubusercontent.com/raw/openresty/lua-nginx-module/pull/1145.patch | git apply && cd ..
+  - pushd ../lua-nginx-module && curl https://patch-diff.githubusercontent.com/raw/openresty/lua-nginx-module/pull/1145.patch | git apply && popd
   - tar zxf download-cache/pcre-$PCRE_VER.tar.gz
   - export PATH=$PWD/work/nginx/sbin:$PWD/openresty-devel-utils:$PATH
   - export LD_PRELOAD=$PWD/mockeagain/mockeagain.so
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,7 +56,7 @@ install:
   - git clone https://github.com/openresty/openresty.git ../openresty
   - git clone https://github.com/openresty/openresty-devel-utils.git
   - git clone https://github.com/simpl/ngx_devel_kit.git ../ndk-nginx-module
-  - git clone https://github.com/openresty/lua-nginx-module.git ../lua-nginx-module
+  - git clone -b ngx-ssl-add-client-addr-function https://github.com/smartwjw/lua-nginx-module.git ../lua-nginx-module
   - git clone https://github.com/openresty/no-pool-nginx.git ../no-pool-nginx
   - git clone https://github.com/openresty/echo-nginx-module.git ../echo-nginx-module
   - git clone https://github.com/openresty/lua-resty-lrucache.git
@@ -79,7 +79,6 @@ script:
   - make -j$JOBS > build.log 2>&1 || (cat build.log && exit 1)
   - sudo make PATH=$PATH install_sw > build.log 2>&1 || (cat build.log && exit 1)
   - cd ../mockeagain/ && make CC=$CC -j$JOBS && cd ..
-  - pushd ../lua-nginx-module && curl https://patch-diff.githubusercontent.com/raw/openresty/lua-nginx-module/pull/1145.patch | git apply && popd
   - tar zxf download-cache/pcre-$PCRE_VER.tar.gz
   - export PATH=$PWD/work/nginx/sbin:$PWD/openresty-devel-utils:$PATH
   - export LD_PRELOAD=$PWD/mockeagain/mockeagain.so
--- a/lib/ngx/ssl.lua
+++ b/lib/ngx/ssl.lua
@@ -37,6 +37,9 @@ int ngx_http_lua_ffi_ssl_raw_server_addr(ngx_http_request_t *r, char **addr,
 int ngx_http_lua_ffi_ssl_server_name(ngx_http_request_t *r, char **name,
     size_t *namelen, char **err);
 
+int ngx_http_lua_ffi_ssl_client_addr(ngx_http_request_t *r, char **addr,
+    size_t *addrlen, char **err);
+
 int ngx_http_lua_ffi_cert_pem_to_der(const unsigned char *pem, size_t pem_len,
     unsigned char *der, char **err);
 
@@ -164,6 +167,25 @@ function _M.server_name()
     return nil, ffi_str(errmsg[0])
 end
 
+function _M.client_addr()
+    local r = getfenv(0).__ngx_req
+    if not r then
+        return error("no request found")
+    end
+
+    local sizep = get_size_ptr()
+
+    local rc = C.ngx_http_lua_ffi_ssl_client_addr(r, charpp, sizep, errmsg)
+    if rc == FFI_OK then
+        return ffi_str(charpp[0], sizep[0])
+    end
+
+    if rc == FFI_DECLINED then
+        return nil
+    end
+
+    return nil, ffi_str(errmsg[0])
+end
 
 function _M.cert_pem_to_der(pem)
     local outbuf = get_string_buf(#pem)
--- a/lib/ngx/ssl.lua
+++ b/lib/ngx/ssl.lua
@@ -37,8 +37,8 @@ int ngx_http_lua_ffi_ssl_raw_server_addr(ngx_http_request_t *r, char **addr,
 int ngx_http_lua_ffi_ssl_server_name(ngx_http_request_t *r, char **name,
     size_t *namelen, char **err);
 
-int ngx_http_lua_ffi_ssl_client_addr(ngx_http_request_t *r, char **addr,
-    size_t *addrlen, char **err);
+int ngx_http_lua_ffi_ssl_raw_client_addr(ngx_http_request_t *r, char **addr,
+    size_t *addrlen, int *addrtype, char **err);
 
 int ngx_http_lua_ffi_cert_pem_to_der(const unsigned char *pem, size_t pem_len,
     unsigned char *der, char **err);
@@ -167,7 +167,7 @@ function _M.server_name()
     return nil, ffi_str(errmsg[0])
 end
 
-function _M.client_addr()
+function _M.raw_client_addr()
     local r = getfenv(0).__ngx_req
     if not r then
         return error("no request found")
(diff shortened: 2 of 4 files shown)


smartwjw:
It seems that ci fails at patching openssl.
```
The command "if [ ! -f openssl-1.0.2h-sess_set_get_cb_yield.patch ]; then wget -O openssl-1.0.2h-sess_set_get_cb_yield.patch https://raw.githubusercontent.com/openresty/openresty/master/patches/openssl-1.0.2h-sess_set_get_cb_yield.patch; fi" exited with 8.
0.01s$ patch -p1 < openssl-1.0.2h-sess_set_get_cb_yield.patch
```
I'm a C and Lua newbie, please correct me if I were wrong.

lib/ngx/ssl.lua
```diff
@@ -37,6 +37,9 @@ int ngx_http_lua_ffi_ssl_raw_server_addr(ngx_http_request_t *r, char **addr,
 int ngx_http_lua_ffi_ssl_server_name(ngx_http_request_t *r, char **name,
     size_t *namelen, char **err);
 
+int ngx_http_lua_ffi_ssl_client_addr(ngx_http_request_t *r, char **addr,
```
agentzh: Where is this C function defined? Am I missing something here?

lib/ngx/ssl.md
```diff
@@ -299,6 +300,43 @@ This function can be called in whatever contexts where downstream https is used.
 
 [Back to TOC](#table-of-contents)
 
+client_addr
+---------------
+**syntax:** *addr, err = ssl.client_addr()*
+
+**context:** *any*
+
+Returns the client ip address in the current SSL connection.
```
agentzh: What is the actual representation of the IP address? Is it in raw binary form or in textual representation? We should be explicit here. Also, the client address may not be an IP address. It could be a unix domain socket client which does not use the IP protocol at all.

lib/ngx/ssl.lua
```diff
@@ -164,6 +167,25 @@ function _M.server_name()
     return nil, ffi_str(errmsg[0])
 end
 
+function _M.client_addr()
```
agentzh: Need some test cases in the existing test suite to cover this new Lua API function.

lib/ngx/ssl.lua
agentzh: Never mind, I'm seeing your openresty/lua-nginx-module#1145 PR now. I think you should temporarily patch `.travis.yml` file in your branch to use your lua-nginx-module branch for your openresty/lua-nginx-module#1145 PR in this PR. Otherwise it wouldn't be compatible.

lib/ngx/ssl.lua
```diff
@@ -164,6 +167,25 @@ function _M.server_name()
...
+    end
+
+    local sizep = get_size_ptr()
+
+    local rc = C.ngx_http_lua_ffi_ssl_client_addr(r, charpp, sizep, errmsg)
+    if rc == FFI_OK then
+        return ffi_str(charpp[0], sizep[0])
+    end
+
+    if rc == FFI_DECLINED then
```
spacewander: The called C function doesn't return NGX_DECLINED, maybe we don't need to handle it here.

lib/ngx/ssl.md
```diff
@@ -299,6 +300,43 @@ This function can be called in whatever contexts where downstream https is used.
...
+
+The following code snippet shows how to delaying TLS handshake to mitigate a kind of TSL handshake attack(eg: close connection immediately after handshake complete):
```
spacewander: I am not a native English speaker, so please correct me if I am wrong. I think it should be `how to delay` instead of `how to delaying`? And `after handshake complete` should be `after handshake completed`?

Moreover, it should be `TLS` instead of `TSL`.

lib/ngx/ssl.md
```diff
@@ -299,6 +300,43 @@ This function can be called in whatever contexts where downstream https is used.
...
+```lua
+local ssl = require "ngx.ssl"
+local blacklist = ngx.shared.ip_blacklist
+
+local addr, err = ssl.client_addr()
+
+-- in blacklist , depay for 30s
+local value, flags = blacklist:get(addr)
+if value then
+    ngx.log(ngx.WARN, "delaying handshake for ip: ", addr .. " ")
```
spacewander: Is the trailing space necessary in the code snippet?

lib/ngx/ssl.md
```diff
@@ -299,6 +300,43 @@ This function can be called in whatever contexts where downstream https is used.
...
+
+Returns the client ip address in the current SSL connection.
+
+The following code snippet shows how to delaying TLS handshake to mitigate a kind of TSL handshake attack(eg: close connection immediately after handshake complete):
+
+```lua
+local ssl = require "ngx.ssl"
+local blacklist = ngx.shared.ip_blacklist
+
+local addr, err = ssl.client_addr()
```
spacewander: It is better to handle all declared variables (especially for err) in your code. Your example will be copied by many people who want to use this method. Therefore, the code style matters.

lib/ngx/ssl.md
```diff
@@ -299,6 +300,43 @@ This function can be called in whatever contexts where downstream https is used.
...
+
+-- in blacklist , depay for 30s
```
spacewander: `depay` => `delay`.

lib/ngx/ssl.md
```diff
@@ -299,6 +300,43 @@ This function can be called in whatever contexts where downstream https is used.
...
+    ngx.log(ngx.WARN, "delaying handshake for ip: ", addr .. " ")
+    ngx.sleep(30)
+    return ngx.exit(ngx.ERROR)
+end
+
+-- add to blacklist at first time, remove it at `access` phase
+ngx.log(ngx.INFO, "adding to backlist, ip: ", addr .. "")
+local succ, err, forcible = blacklist:set(addr, '1')
+if not succ then
+    ngx.log(ngx.ERR, "failed to add backlist, ip: ", addr, err)
```
spacewander: Maybe a space is need between `addr` and `err`.

lib/ngx/ssl.lua
smartwjw: alreay fixed.

lib/ngx/ssl.lua
smartwjw: alreay fixed.

lib/ngx/ssl.lua
smartwjw: Already fixed.

lib/ngx/ssl.md
smartwjw: Already fixed.

lib/ngx/ssl.md
```diff
@@ -299,6 +300,50 @@ This function can be called in whatever contexts where downstream https is used.
...
+-- in blacklist , delay for 30s
+local value, flags = blacklist:get(addr)
+if value then
+    ngx.log(ngx.WARN, "delaying handshake for ip: ", addr)
+    ngx.sleep(30)
+    return ngx.exit(ngx.ERROR)
+end
+
+-- add to blacklist at first time, remove it at `access` phase
+ngx.log(ngx.INFO, "adding to backlist, ip: ", addr .. "")
```
spacewander: Why should we concatenate `addr` with an empty string?

lib/ngx/ssl.md
```diff
@@ -299,6 +300,50 @@ This function can be called in whatever contexts where downstream https is used.
...
+The following code snippet shows how to delay TLS handshake to mitigate a kind of TLS
+handshake attack(eg: close connection immediately after handshake completed):
+
+```lua
+local ssl = require "ngx.ssl"
+local blacklist = ngx.shared.ip_blacklist
+
+local addr, err = ssl.client_addr()
+
+if err then
```
spacewander: Prefer `if not addr then`, which is more common.

lib/ngx/ssl.lua
```diff
@@ -157,14 +160,29 @@ function _M.server_name()
...
+    end
+
+    local sizep = get_size_ptr()
+
+    local rc = C.ngx_http_lua_ffi_ssl_client_addr(r, charpp, sizep, errmsg)
+    if rc == FFI_OK then
+        return ffi_str(charpp[0], sizep[0])
+    end
+
     if rc == FFI_DECLINED then
```
spacewander: You should remove the check in `client_addr` instead of the one in `server_name`.

smartwjw:
@spacewander thanks for your opinions, all problems are fixed.

lib/ngx/ssl.md
```diff
@@ -299,6 +300,50 @@ This function can be called in whatever contexts where downstream https is used.
 
 [Back to TOC](#table-of-contents)
 
+client_addr
+---------------
+**syntax:** *addr, err = ssl.client_addr()*
+
+**context:** *any*
+
+Returns the client ip address in textual representation of the current SSL connection,
```
agentzh: The API name suggests that it is not specific to the IP protocol. We should support other kinds of clients like unix domain socket clients.

smartwjw:
Is there any way I can run test locally, so I can push the code after all tests pass instead of pushing the code and wait for  travis-ci build  result.

agentzh:
@smartwjw Just check out the commands for setting the test env is setup in `.travis.yml`. Any Linux distributions should work just fine though `.travis.yml` is for Ubuntu trusty, in particular. For example, I run the tests in Fedora locally.

agentzh:
@smartwjw I think raw client address is much more useful. Let's just use raw server address approach here? It would be much more difficult to parse the textual representation than formatting a binary form to a textual form.

smartwjw:
@agentzh yes, raw client address is more useful, I will change it soon.

.travis.yml
```diff
@@ -79,6 +79,7 @@ script:
   - make -j$JOBS > build.log 2>&1 || (cat build.log && exit 1)
   - sudo make PATH=$PATH install_sw > build.log 2>&1 || (cat build.log && exit 1)
   - cd ../mockeagain/ && make CC=$CC -j$JOBS && cd ..
+  - pushd ../lua-nginx-module && curl https://patch-diff.githubusercontent.com/raw/openresty/lua-nginx-module/pull/1145.patch | git apply && popd
```
agentzh: You can just change the `git clone` line to clone your branch instead. It's easier.

lib/ngx/ssl.lua
```diff
@@ -164,6 +167,26 @@ function _M.server_name()
     return nil, ffi_str(errmsg[0])
 end
```
agentzh: Style: we use 2 blank lines to separate Lua function definitions consistently in this Lua library.

lib/ngx/ssl.lua
```diff
@@ -164,6 +167,26 @@ function _M.server_name()
...
+    if rc == FFI_OK then
+        local typ = addr_types[intp[0]]
+        if not typ then
+            return nil, nil, "unknown address type: " .. intp[0]
+        end
+        return ffi_str(charpp[0], sizep[0]), typ
+    end
+
+    return nil, nil, ffi_str(errmsg[0])
+end
```
agentzh: Ditto.

.travis.yml
smartwjw: thx.

agentzh:
@smartwjw Seems like the travis ci testing is still failing horribly? ;)

smartwjw:
@agentzh I setup a local develop environment today, it indeed find some problems, but after I fix the problems, the test still fails, and I don't how to deal with it, please help.

image

agentzh:
@smartwjw That's just because you use the `--- ONLY` section and it's expected that the test plan declared at the beginning of the `.t` file would not match the actual number of subtests run.

lib/ngx/ssl.md
```diff
@@ -299,6 +300,56 @@ This function can be called in whatever contexts where downstream https is used.
...
+local byte = string.byte
+
+local addr, addrtyp, err = ssl.raw_client_addr()
+if not addr then
+    ngx.log(ngx.ERR, "failed to fetch raw client addr: ", err)
+    return
+end
+
+if addrtyp == "inet" then  -- IPv4
+    ip = string.format("%d.%d.%d.%d", byte(addr, 1), byte(addr, 2),
```
spacewander: @smartwjw 
Better to localize `string.format` just like the `string.byte`?

t/ssl.t
```diff
@@ -2065,3 +2065,106 @@ qr/\[error\] .*? failed to parse pem key: PEM_read_bio_PrivateKey\(\) failed/
...
+    lua_package_path "$TEST_NGINX_LUA_PACKAGE_PATH/?.lua;;";
+
+    server {
+        listen 127.0.0.1:12345 ssl;
+        server_name   test.com;
+        ssl_certificate_by_lua_block {
+            local ssl = require "ngx.ssl"
+            local byte = string.byte
+            local addr, addrtype, err = ssl.raw_client_addr()
+            local ip = string.format("%d.%d.%d.%d", byte(addr, 1), byte(addr, 2),
```
spacewander: @smartwjw 
Better to localize `string.format` just like the `string.byte`?

agentzh: For test cases, I think it's fine, but for documentation, it's worth it since the user usually just does copy&paste :)

lib/ngx/ssl.md
agentzh: Agreed here.

agentzh:
@smartwjw Merged with minor fixes. Thanks!

PR #140, opened 2017-08-31
feature: implemented shdict:ttl() and shdict:expire() APIs

thibaultcha:

Hi there,

This is a stab at:
- openresty/lua-nginx-module#552
- openresty/lua-nginx-module#1101

No documentation for now, just a proposal. I thought about including this API in a `ngx.shared` module the user would have to import explicitly (since they don't have Lua_CFunction counterparts), but went the easy way for now.

Probably needs a bit more polishing, open to feedback here!

I would have opened a PR for `shdict:get_info()` as well but apparently have been beaten to it :)

### DIFF
--- a/lib/resty/core/shdict.lua
+++ b/lib/resty/core/shdict.lua
@@ -17,6 +17,8 @@ local type = type
 local error = error
 local ngx_shared = ngx.shared
 local getmetatable = getmetatable
+local FFI_ERROR = base.FFI_ERROR
+local FFI_DECLINED = base.FFI_DECLINED
 
 
 ffi.cdef[[
@@ -36,6 +38,12 @@ ffi.cdef[[
         int *forcible);
 
     int ngx_http_lua_ffi_shdict_flush_all(void *zone);
+
+    int ngx_http_lua_ffi_shdict_get_ttl(void *zone,
+        const unsigned char *key, size_t key_len);
+
+    int ngx_http_lua_ffi_shdict_set_expire(void *zone,
+        const unsigned char *key, size_t key_len, int exptime);
 ]]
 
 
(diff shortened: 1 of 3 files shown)


thibaultcha:
As soon as I hear this API is a go, I'll provide documentation for it!

agentzh:
@thibaultcha Sorry for the delay on my side. I'll have a look later today. Thanks!

agentzh:
@thibaultcha What is the sister PR in lua-nginx-module?

agentzh:
@thibaultcha Never mind, just found the sister ngx_lua PR myself: https://github.com/openresty/lua-nginx-module/pull/1150

thibaultcha:
Ok :) (It's in the list of referenced PRs in the GitHub UI, but since I referenced this PR in several other places, it became harder to distinguish. Sorry!)

agentzh:
@thibaultcha No worries :)

lib/resty/core/shdict.lua
```diff
@@ -387,6 +395,79 @@ local function shdict_flush_all(zone)
...
+
+    if type(key) ~= "string" then
+        key = tostring(key)
+    end
+
+    local key_len = #key
+    if key_len == 0 then
+        return nil, "empty key"
+    end
+    if key_len > 65535 then
```
agentzh: Better add a blank line between successive if statement blocks.

lib/resty/core/shdict.lua
```diff
@@ -387,6 +395,79 @@ local function shdict_flush_all(zone)
...
```

lib/resty/core/shdict.lua
thibaultcha: I certainly concur. It was only like so to be consistent with the above `incr()` method. Fixed.

thibaultcha:
It seems like a recent release of Luacheck might have (once again) broken its default settings and now reports errors for this codebase?

agentzh:
@thibaultcha Should we make the luacheck warnings non-fatal in travis testing?

thibaultcha:
@agentzh I am not sure if you are asking for my opinion, but if so, then I am a big fan of enforcing the linter check in CI. However, I only dedicate a single Travis job to linting, like here: https://travis-ci.org/thibaultcha/lua-resty-mlcache

This way, the CI test suites of a project is not impacted by potential linter failures (tests can still be successful even if the linter is complaining about something). It is a better separation of concerns in the CI imho.
Plus, if we want to, we can even declare the linter job to be an `allowed_failure` in Travis CI, so that the CI test suites can still be passing, but if one of us opens the Travis build, we can spot the linter job failed. I have that setup for [another project](https://github.com/thibaultcha/lua-argon2/blob/master/.travis.yml#L27-L28). 

Happy to add those CI changes in #143 if you think they'd be valuable for lua-resty-core.

agentzh:
@thibaultcha I'm not sure if that would be a good use of your valuable time :) Maybe we can just stick with a particular version of luacheck to avoid such regressions?

thibaultcha:
> Maybe we can just stick with a particular version of luacheck to avoid such regressions

Definitely a solution I have used in the past as well 👍 (I can try to keep an eye for Luacheck releases to avoid lagging too far behind)

agentzh:
@thibaultcha Cool, thanks!

thibaultcha:
@agentzh I just added a commit to #143 which locks the Luacheck version to the latest available!

thibaultcha:
@dndx Thank you for your review!

agentzh:
@thibaultcha I'd propose the following patch to make the test better:

```
diff --git a/t/shared.t b/t/shared.t
index a4041b5..41d3bf8 100644
--- a/t/shared.t
+++ b/t/shared.t
@@ -129,10 +129,11 @@ failed to get ttl: not found
     }
 --- request
 GET /t
---- response_body_like
-0.2
+--- response_body_like chomp
+\A0.2
 sleep for 0.1s...
-0.0\d*
+-?0.0\d*
+\z
 --- no_error_log
 [error]
 [alert]
```

What do you think?

agentzh:
Merged. Thanks!

Will you create another PR to lua-nginx-module to document these new API functions? Thanks!

thibaultcha:
Great! Good idea for the test. Yes, I'll propose those docs as soon as I can. Do you have a prefered timeline? Some deadline? ASAP maybe?

agentzh:
@thibaultcha This week maybe? Next week is fine too :) Not that urgent. Thanks!

thibaultcha:
Sounds good! I'll try to allocate some time for it.

agentzh:
@thibaultcha Great! Thanks!

PR #141, opened 2017-08-31
Add get_stats method to ngx.shared.DICT

hnakamur:
Signed-off-by: Hiroaki Nakamura <hnakamur@gmail.com>

See https://github.com/openresty/lua-nginx-module/pull/1149 for details.


lib/resty/core/shdict.lua
```diff
@@ -387,6 +401,25 @@ local function shdict_flush_all(zone)
 end
 
 
+local function shdict_get_used_bytes(zone)
+    zone = check_zone(zone)
+
+    local slot_count_buf = ffi_new("size_t[1]")
+    C.ngx_http_lua_ffi_shdict_get_stats_slot_count(zone, slot_count_buf)
+    local stats = ffi_new("ngx_slab_stat_t[?]", slot_count_buf[0])
```
agentzh: Similar to new Lua table creation, we should also avoid creating new cdata objects on hot code paths.

lib/resty/core/shdict.lua
```diff
@@ -387,6 +401,25 @@ local function shdict_flush_all(zone)
 end
 
 
+local function shdict_get_used_bytes(zone)
+    zone = check_zone(zone)
+
+    local slot_count_buf = ffi_new("size_t[1]")
```
agentzh: Like below.

lib/resty/core/shdict.lua
```diff
@@ -387,6 +401,25 @@ local function shdict_flush_all(zone)
...
+    C.ngx_http_lua_ffi_shdict_get_stats_slot_count(zone, slot_count_buf)
+    local stats = ffi_new("ngx_slab_stat_t[?]", slot_count_buf[0])
+    local min_size = ffi_new("size_t[1]")
```

lib/resty/core/shdict.lua
```diff
@@ -387,6 +401,25 @@ local function shdict_flush_all(zone)
...
+    C.ngx_http_lua_ffi_shdict_get_stats_slots_and_min_size(zone, slot_count_buf[0], stats, min_size)
+    local slot_count = tonumber(slot_count_buf[0])
+    local slot_size = tonumber(min_size[0])
+    local total_used = 0
+    for i = 1, slot_count do
+        total_used = total_used + slot_size * tonumber(stats[i-1].used)
+        slot_size = slot_size * 2
+    end
+    return total_used
```
agentzh: It might be useful to have the more detailed stats in each slot. Not just the totals.

hnakamur:
Thanks for your comments.
I've updated the pull request.

hnakamur:
I added the second argument `detail`.
If you pass `true` then the return value of `get_stats` has detailed stats in `slots` key as well as `total_used_bytes`.
If you pass `false` then the return value of `get_stats` has only `total_used_bytes` key.
Could you take another look?
Thanks!

lib/resty/core/shdict.lua
```diff
@@ -387,6 +403,39 @@ local function shdict_flush_all(zone)
...
+local function shdict_get_stats(zone, detail)
+    zone = check_zone(zone)
+
+    C.ngx_http_lua_ffi_shdict_get_stats_slot_count(zone, slot_count_buf)
+    local stats = ffi_new("ngx_slab_stat_t[?]", slot_count_buf[0])
+    C.ngx_http_lua_ffi_shdict_get_stats_slots_and_min_size(zone, slot_count_buf[0], stats, slot_min_size_buf)
+    local slot_count = tonumber(slot_count_buf[0])
+    local slot_size = tonumber(slot_min_size_buf[0])
+    local total_used = 0
+    local slots = {}
```
agentzh: Again, can we allow the caller to recycle the Lua table here? This way we can avoid allocating a new Lua table in every invocation of this Lua function.

lib/resty/core/shdict.lua
```diff
@@ -387,6 +403,39 @@ local function shdict_flush_all(zone)
...
+    for i = 1, slot_count do
+        local stat = stats[i - 1]
+        if detail then
+            slots[i] = {
```

lib/resty/core/shdict.lua
```diff
@@ -387,6 +403,39 @@ local function shdict_flush_all(zone)
...
+                used = tonumber(stat.used),
+                reqs = tonumber(stat.reqs),
+                fails = tonumber(stat.fails),
+            }
+        end
+        total_used = total_used + slot_size * tonumber(stat.used)
+        slot_size = slot_size * 2
+    end
+
+    local stats = {total_used_bytes = total_used}
```

hnakamur:
I've changed the APIs for caller to recyle Lua tables.
Could you take another look?
Thanks!

hnakamur:
I created another version of API at https://github.com/openresty/lua-resty-core/compare/master...hnakamur:shdict_get_stats
Which are better?

lib/resty/core/shdict.lua
```diff
@@ -387,6 +403,40 @@ local function shdict_flush_all(zone)
...
+
+    C.ngx_http_lua_ffi_shdict_get_stats_slot_count(zone, slot_count_buf)
+    return tonumber(slot_count_buf[0])
+end
+
+
+local function shdict_get_stats(zone, slots)
+    zone = check_zone(zone)
+
+    C.ngx_http_lua_ffi_shdict_get_stats_slot_count(zone, slot_count_buf)
```
agentzh: We discard the return value here? Better do error handling properly?

t/shdict.t
```diff
@@ -999,3 +999,70 @@ number expected, got string
...
+    location = /t {
+        content_by_lua_block {
+            local ffi = require "ffi"
+            local dogs = ngx.shared.dogs
+            local slot_count = dogs:get_pool_slot_count()
+            local slots = {}
+            for i = 1, slot_count do
+                slots[i] = {}
+            end
+            local total_used_bytes = dogs:get_stats(slots)
```
agentzh: Sorry, I changed my mind again. Seems like it's not worth the complexity and overhead of the API to get all the details in each slot. Indeed, 99% of the times, the user is only interested in the total stats. Sorry again for the inconvenience!

lib/resty/core/shdict.lua
```diff
@@ -404,6 +454,8 @@ if ngx_shared then
                 mt.replace = shdict_replace
                 mt.delete = shdict_delete
                 mt.flush_all = shdict_flush_all
+                mt.get_pool_slot_count = shdict_get_pool_slot_count
```
agentzh: Let's remove this API function.

lib/resty/core/shdict.lua
```diff
@@ -404,6 +454,8 @@ if ngx_shared then
...
+                mt.get_stats = shdict_get_stats
```
agentzh: Let's rename this method to just `stats()`?

lib/resty/core/shdict.lua
hnakamur: Currently `C.ngx_http_lua_ffi_shdict_get_stats_slot_count` and `C.ngx_http_lua_ffi_shdict_get_stats_slots_and_min_size` always return `NGX_OK`, so I omit the error handling?
https://github.com/openresty/lua-nginx-module/pull/1149/files

Should I do more defensive check in above two C APIs and do error handling in Lua side?

lib/resty/core/shdict.lua
```diff
@@ -387,6 +403,40 @@ local function shdict_flush_all(zone)
...
+    C.ngx_http_lua_ffi_shdict_get_stats_slot_count(zone, slot_count_buf)
+    return tonumber(slot_count_buf[0])
+end
+
+
+local function shdict_get_stats(zone, slots)
+    zone = check_zone(zone)
+
+    C.ngx_http_lua_ffi_shdict_get_stats_slot_count(zone, slot_count_buf)
+    local stats = ffi_new("ngx_slab_stat_t[?]", slot_count_buf[0])
```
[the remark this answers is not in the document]
hnakamur: Thanks, I didn't notice the string buffer provided by `resty.core.base`.
I've updated this pull request to use it.

lib/resty/core/shdict.lua
hnakamur: Done.

lib/resty/core/shdict.lua
hnakamur: OK, done.

t/shdict.t
hnakamur: Don't worry! No need for apology here. What is important here is providing a API which is both easy to use and performant. I'd like to choose the best balance too.
I agree the user is only interested in the total memory used bytes 99% of times.

Oh, by the way, did you verify the calculation of  the total memory used?
Please make sure the calculation is correct.

I've updated the other patch at https://github.com/openresty/lua-resty-core/compare/master...hnakamur:shdict_get_stats.
Now the difference is this pull request needs dynamic record table allocations for each element of the array table, and the other patch needs dynamic allocations for keys like `i .. '.size'`.
Which do you prefer?

hnakamur:
I added my comments to all your comments at https://github.com/openresty/lua-resty-core/pull/141#pullrequestreview-59941652
Could you `Show outdated` links and see my comments?

t/shdict.t
hnakamur: By the way, this review is a great learning experience for me to implement an openresty API in the FFI way.
Thanks for your time!

lib/resty/core/shdict.lua
agentzh: @hnakamur It's better for the Lua wrapper not to assume the internal workings of the C API. If the return value is never meaningful we can simply make the C API void typed.

lib/resty/core/shdict.lua
```diff
@@ -387,6 +406,34 @@ local function shdict_flush_all(zone)
...
+                total = tonumber(stat.total),
+                used = tonumber(stat.used),
+                reqs = tonumber(stat.reqs),
+                fails = tonumber(stat.fails),
+            }
+        end
+        total_used = total_used + slot_size * tonumber(stat.used)
+        slot_size = slot_size * 2
+    end
+    return total_used
```
agentzh: Let's just return the total used and total size values directly? This way we do not need the dynamically-sized cdata or Lua tables.

hnakamur:
I've updated both this pull request and [C API pull request](https://github.com/openresty/lua-nginx-module/pull/1149/files).
Now the Lua api returns only the total used and the total size.

I'm going to add document now.

hnakamur:
I looked at https://github.com/openresty/lua-resty-core#restycoreshdict and noticed links like ngx.shared.DICT.get points to the corresponding API document of lua-nginx-module.

However `ngx.shared.DICT.stats` exists only in lua-resty-core and does not exist in lua-nginx-module.
How should we do here?

agentzh:
@hnakamur We can just add the documentation in lua-resty-core's README.

hnakamur:
Thanks, I've added document for ngx.shared.DICT.stats in lua-resty-core's README.
Could you review it?
Thanks!

hnakamur:
@agentzh ping. Could you take a look at document I added? Thanks!

thibaultcha:
What can we do to help here? This is a very interesting API for shdicts. It seems like this PR still needs some polishing, but before diving into a review on my end, I'd like to know: is this the function signature and the approach (roughly, of course) you are envisioning @agentzh?

README.markdown
```diff
@@ -150,6 +150,32 @@ API Implemented
...
+
+or use Lua's syntactic sugar for method calls:
+
+```lua
+
+ local cats = ngx.shared.cats
+ local total_used, total_size = cats:stats()
+```
+
+These two forms are fundamentally equivalent.
```
agentzh: It's better not elucidate the Lua method call syntax here. Otherwise we should add it to add to *all* the shdict methods' documentation, which is definitely not desired :)

t/shdict.t
```diff
@@ -999,3 +999,26 @@ number expected, got string
...
+
+=== TEST 31: get stats
+--- http_config eval: $::HttpConfig
+--- config
+    location = /t {
+        content_by_lua_block {
+            local dogs = ngx.shared.dogs
+            local total_used, total_size = dogs:stats()
+            ngx.say("total_used type: ", type(total_used))
+            ngx.say("total_size type: ", type(total_size))
```
agentzh: Can we do more strict checks here? For one, the numbers returned should all be non-negative integer values. Also, the first value must be smaller than the second. Finally, the total size should be predetermined and is a constant (though may differ on 32-bit and 64-bit systems, need to be prepared for that).

Also I think the first value should be a constant for this particular test case and it only differs on 32-bit and 64-bit systems and we just accept those 2 different constants in the test.

README.markdown
```diff
@@ -150,6 +150,32 @@ API Implemented
...
+ local cats = ngx.shared.cats
+ local total_used, total_size = cats.stats(cats)
+```
+
+or use Lua's syntactic sugar for method calls:
+
+```lua
+
+ local cats = ngx.shared.cats
+ local total_used, total_size = cats:stats()
```
agentzh: `total_size` is confusing when we also have `total_used` since the latter is also a "size" per se. Better introduce clearer and shorter names like `used` and `total`?

agentzh:
Sorry for the delay on my side. Will you please rebase to the latest master? There seems to be merge conflicts in this branch now.

agentzh:
BTW, please temporarily use your own branch of ngx_lua in the `.travis.yml` file so that we can know the Travis CI testing is passing.

t/shdict.t
```diff
@@ -999,3 +999,26 @@ number expected, got string
...
+    }
+--- request
+GET /t
+--- response_body
+total_used type: number
+total_size type: number
+--- no_error_log
+[error]
+[alert]
+[crit]
```
agentzh: I think we need to add more test cases for the following scenarios:

1. the shdict is empty,
2. the shdict is full,
3. the shdict is half-full,
4. the shdict has tiny key-value pairs (just a few bytes).
5. the shidct has big key-value pairs (larger than a page size).

Of course, we all need to check the actual values in all these cases rather than simply check the values are of the number type. Otherwise it makes little sense :)

README.markdown
hnakamur: done

README.markdown
hnakamur: I copied it from [ngx.shared.DICT.set](https://github.com/openresty/lua-nginx-module#ngxshareddictset), but I noticed the other methods don't have this explanation.
So I deleted it all together.

hnakamur:
https://github.com/openresty/lua-resty-core/pull/141#issuecomment-331656680
https://github.com/openresty/lua-resty-core/pull/141#issuecomment-331656764
Never mind. I've updated the pull request for these two comments.

Now I'll think about adding test cases.

lib/resty/core/shdict.lua
```diff
@@ -469,6 +477,22 @@ local function shdict_expire(zone, key, exptime)
...
+    shdict_stats = function(zone)
+        zone = check_zone(zone)
+
+        C.ngx_http_lua_ffi_shdict_get_stats(zone, stats_used_buf,
+            stats_total_buf)
+        return tonumber(stats_used_buf[0]), tonumber(stats_total_buf[0])
+    end
+else
+    shdict_stats = function(zone)
+        -- dummy implementation
```
agentzh: I think we should throw out an exception here instead of giving the wrong values.

t/shdict.t
```diff
@@ -999,3 +999,25 @@ number expected, got string
 --- no_error_log
 [alert]
 [crit]
+
+
```
agentzh: Please use the `reindex` script to format your tests. We need 3 blank lines here, for example.

README.markdown
```diff
@@ -152,6 +152,14 @@ API Implemented
...
 * [ngx.shared.DICT.expire](https://github.com/openresty/lua-nginx-module#ngxshareddictexpire)
 * [ngx.shared.DICT.flush_all](https://github.com/openresty/lua-nginx-module#ngxshareddictflush_all)
+* [ngx.shared.DICT.stats](#ngxshareddictstats)
+
+### ngx.shared.DICT.stats
+**syntax:** *used, total = ngx.shared.DICT:stats()
+
+**context:** *set_by_lua&#42;, rewrite_by_lua&#42;, access_by_lua&#42;, content_by_lua&#42;, header_filter_by_lua&#42;, body_filter_by_lua&#42;, log_by_lua&#42;, ngx.timer.&#42;, balancer_by_lua&#42;, ssl_certificate_by_lua&#42;, ssl_session_fetch_by_lua&#42;, ssl_session_store_by_lua&#42;*
+
+Retrieving the total used counts and total size in bytes for the dictionary [ngx.shared.DICT](#ngxshareddict).
```
agentzh: This should go to the ngx_lua module's documentation instead, like @thibaultcha's new `expire()` and `ttl()` methods' docs. We'll drop support for the CFunction based Lua API in the ngx_lua core anyway and lua-resty-core will soon become a hard prerequisite for ngx_lua (probably in the next OpenResty release).

hnakamur:
Is there a way to skip tests for ngx.shared.DICT.stats when the stats method is not defined?

t/shdict.t
hnakamur: done

README.markdown
hnakamur: done

agentzh:
@hnakamur There is a `--- skip_nginx` section for that purpose. See

https://metacpan.org/pod/Test::Nginx::Socket#skip_nginx

hnakamur:
Currently, I got unexpected test results at https://travis-ci.org/openresty/lua-resty-core/jobs/279525893

I'd like to check debug log output from
https://github.com/nginx/nginx/blob/release-1.13.5/src/core/ngx_slab.c#L220-L221

How to enable the debug log for NGX_LOG_DEBUG_ALLOC and keep the debug log for `prove -Itest-nginx/lib -j$JOBS -r t` ?

agentzh:
@hnakamur You should have debugging logs by default. Check the log filtering level at the beginning of the `.t` file though.

hnakamur:
Thanks for you comment.
However still I'm not successful to get debug log.

As a workaround, I add a lot of `ngx_error_log` to nginx and lua-nginx-module in my local environment which I set up with the same command in .travis.yml.

I confirmed that `ngx_http_lua_shdict_init_zone` calls `ngx_slab_alloc` 2 times, 1 time each for below lines.

* https://github.com/openresty/lua-nginx-module/blob/bf14723e4e7749c989134c029742185db1c78255/src/ngx_http_lua_shdict.c#L108
    * `sizeof(ngx_http_lua_shdict_shctx_t)` = 80 and the slot of size 128 is used.
* https://github.com/openresty/lua-nginx-module/blob/bf14723e4e7749c989134c029742185db1c78255/src/ngx_http_lua_shdict.c#L120
   * If the shared dict name is `empty_cats`, `sizeof(" in lua_shared_dict zone \"\"") + shm_zone->shm.name.len` = 38  and the slot of size 64 is used.

So, for an empty shared dict, used = 2, total = 128 + 64 = 192.
As a result, the actual result for an empty shared dict in https://travis-ci.org/openresty/lua-resty-core/jobs/279525893 is correct.

```
#   Failed test 'TEST 31: stats, empty - response_body - response is expected (repeated req 1, req 0)'
#   at /home/travis/build/openresty/lua-resty-core/test-nginx/lib/Test/Nginx/Socket.pm line 1505.
#          got: ..."ber\x{0a}used: 2\x{0a}total: 192\x{0a}"...
#       length: 56
#     expected: ..."ber\x{0a}used: 0\x{0a}total: 0\x{0a}"...
#       length: 54
#     strings begin to differ at char 44 (line 3 column 7)
```

I'll modify the expected result in my test case.

I'll investigate in two other test cases, one for a half full shared dict, the other for a full shared dict.

agentzh:
@hnakamur Great! Tests definitely help us better understand the results and behavior ;)

hnakamur:
I noticed that `forcible` returned by `ngx.shared.DICT.set` is true for adding 32 pairs of 8-byte-long key and 8-byte-long value at the test case 10 in `t/shdict_stats.t`.
https://github.com/hnakamur/lua-resty-core/blob/85897f6445754fc81b96ceb5a3fab2530fd0d70d/t/shdict_stats.t#L327-L364

The total size of the dict was 4160 bytes and the declared size of the dict is 12 kbytes.
https://github.com/hnakamur/lua-resty-core/blob/85897f6445754fc81b96ceb5a3fab2530fd0d70d/t/shdict_stats.t#L18

The actual total size 4160 bytes is a way smaller than the declared size 12 kbytes.
Is this an expected behavior?

agentzh:
@hnakamur Maybe this is due to memory fragmentation or shm zones' own memory overhead. You need to dig deeper to be sure though :)

hnakamur:
I digged deeper to read lua-nginx-module and nginx slab allocator code.
I wrote a blog in Japanese at https://hnakamur.github.io/blog/2017/09/27/code-reading-lua-nginx-shared-dict/

I learned that the nginx slab allocator prepares slots for 8, 16, 32, ..., 2048 bytes and 4KB pages at the initialization and use 4KB pages for each slot on demand.

So even if there is a room in pages already allocated for some slots, if there is no room in pages for other slots and no more free page left, ngx.shared.DICT.set may returns forcible=true or non nil err.

The main purpose I'd like to add `stats` method is to check that there is enough room to set a new key without `set` method returning forcible=true.

And I'd like a pessimistic estimation here. Maybe you can set a new key for some slots, but if you cannot set a new key for other slots, I'd like to get an alert notification.

So what I really needed to know is if there is a free page left or not.
How about renaming `stats` method to `free_page_sizes` or something and changing the signature like:

```
sizes_in_bytes = ngx.shared.DICT.free_page_sizes()
```

@agentzh @thibaultcha 
Could you tell me what do you think?
Also if you come up with a better name, could you tell me?

agentzh:
@hnakamur I think `stats` is fine and the free page size is a good enough first order approximation.

hnakamur:
@agentzh Thanks for your comment. I've updated the `stats` method to return free page size.
Could you review it?
Thanks!

.travis.yml
```diff
@@ -57,7 +57,8 @@ install:
   - git clone https://github.com/openresty/openresty.git ../openresty
   - git clone https://github.com/openresty/openresty-devel-utils.git
   - git clone https://github.com/simpl/ngx_devel_kit.git ../ndk-nginx-module
-  - git clone https://github.com/openresty/lua-nginx-module.git ../lua-nginx-module
+  - git clone https://github.com/hnakamur/lua-nginx-module.git ../lua-nginx-module
```
agentzh: You can use the `-b BRANCH` option o fthe `git clone` command here BTW.

hnakamur: Thanks for the tip! I didn't know that.
But it's temporary until the https://github.com/openresty/lua-resty-core/pull/141 is merged, so I'll leave it as is.

README.markdown
```diff
@@ -152,8 +152,7 @@ API Implemented
 * [ngx.shared.DICT.ttl](https://github.com/openresty/lua-nginx-module#ngxshareddictttl)
 * [ngx.shared.DICT.expire](https://github.com/openresty/lua-nginx-module#ngxshareddictexpire)
 * [ngx.shared.DICT.flush_all](https://github.com/openresty/lua-nginx-module#ngxshareddictflush_all)
-
-[Back to TOC](#table-of-contents)
+* [ngx.shared.DICT.stats](https://github.com/openresty/lua-nginx-module#ngxshareddictstats)
```
agentzh: We should remove this now?

lib/resty/core/shdict.lua
```diff
@@ -469,6 +476,18 @@ local function shdict_expire(zone, key, exptime)
     return true
 end
```
agentzh: Style: we use 2 blank lines to separate Lua function definitions consistently in this library. Please fix other similar places as well.

t/shdict.t
```diff
@@ -999,3 +1000,213 @@ number expected, got string
...
+            end
+            local free_page_bytes = cats:free_space()
+            ngx.say("free_page_bytes type: ", type(free_page_bytes))
+            ngx.say("free_page_bytes: ", free_page_bytes)
+        }
+    }
+--- request
+GET /t
+--- response_body
+got forcible, i=64
```
agentzh: We may have to relax this response body test for 32-bit systems.

t/shdict.t
```diff
@@ -15,6 +15,7 @@ my $pwd = cwd();
 
 our $HttpConfig = <<_EOC_;
     lua_shared_dict dogs 1m;
+    lua_shared_dict cats 16k;
```
agentzh: A 16k zone might be a bit too small for our testing purposes. I'd add some tests for 100k at least ;)

agentzh:
@hnakamur Merged with the following extra patch:

```diff
diff --git a/.travis.yml b/.travis.yml
index 39a80c4..1277256 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,8 +60,7 @@ install:
   - git clone https://github.com/openresty/openresty.git ../openresty
   - git clone https://github.com/openresty/openresty-devel-utils.git
   - git clone https://github.com/simpl/ngx_devel_kit.git ../ndk-nginx-module
-  - git clone https://github.com/hnakamur/lua-nginx-module.git ../lua-nginx-module
-  - (cd ../lua-nginx-module && git checkout shared_dict_stats)
+  - git clone https://github.com/openresty/lua-nginx-module.git ../lua-nginx-module
   - git clone https://github.com/openresty/no-pool-nginx.git ../no-pool-nginx
   - git clone https://github.com/openresty/echo-nginx-module.git ../echo-nginx-module
   - git clone https://github.com/openresty/lua-resty-lrucache.git
@@ -98,6 +97,4 @@ script:
   - ngx-build $NGINX_VERSION --with-ipv6 --with-http_realip_module --with-http_ssl_module --with-pcre-jit --with-cc-opt="-I$OPENSSL_INC -I$PCRE_INC" --with-ld-opt="-L$OPENSSL_LIB -Wl,-rpath,$OPENSSL_LIB -L$PCRE_LIB -Wl,-rpath,$PCRE_LIB" --add-module=../ndk-nginx-module --add-module=../echo-nginx-module --add-module=../headers-more-nginx-module --add-module=../lua-nginx-module --with-debug > build.log 2>&1 || (cat build.log && exit 1)
   - nginx -V
   - ldd `which nginx`|grep -E 'luajit|ssl|pcre'
-  - git branch
-  - git log -1
   - prove -Itest-nginx/lib -j$JOBS -r t
diff --git a/README.markdown b/README.markdown
index 5af9c4c..0f220e5 100644
--- a/README.markdown
+++ b/README.markdown
@@ -153,7 +153,8 @@ API Implemented
 * [ngx.shared.DICT.ttl](https://github.com/openresty/lua-nginx-module#ngxshareddictttl)
 * [ngx.shared.DICT.expire](https://github.com/openresty/lua-nginx-module#ngxshareddictexpire)
 * [ngx.shared.DICT.flush_all](https://github.com/openresty/lua-nginx-module#ngxshareddictflush_all)
-* [ngx.shared.DICT.stats](https://github.com/openresty/lua-nginx-module#ngxshareddictstats)
+* [ngx.shared.DICT.free_space](https://github.com/openresty/lua-nginx-module#ngxshareddictfree_space)
+* [ngx.shared.DICT.capacity](https://github.com/openresty/lua-nginx-module#ngxshareddictcapacity)
 
 ## resty.core.var
 
diff --git a/lib/resty/core/shdict.lua b/lib/resty/core/shdict.lua
index 70d645a..b9fc044 100644
--- a/lib/resty/core/shdict.lua
+++ b/lib/resty/core/shdict.lua
@@ -476,12 +476,14 @@ local function shdict_expire(zone, key, exptime)
     return true
 end
 
+
 local function shdict_capacity(zone)
     zone = check_zone(zone)
 
     return tonumber(C.ngx_http_lua_ffi_shdict_capacity(zone))
 end
 
+
 local function shdict_free_space(zone)
     zone = check_zone(zone)
 
diff --git a/t/shdict.t b/t/shdict.t
index 3d3fe9d..e860208 100644
--- a/t/shdict.t
+++ b/t/shdict.t
@@ -16,6 +16,7 @@ my $pwd = cwd();
 our $HttpConfig = <<_EOC_;
     lua_shared_dict dogs 1m;
     lua_shared_dict cats 16k;
+    lua_shared_dict birds 100k;
     lua_package_path "$pwd/lib/?.lua;../lua-resty-lrucache/lib/?.lua;;";
     init_by_lua_block {
         local verbose = false
@@ -33,7 +34,7 @@ our $HttpConfig = <<_EOC_;
 _EOC_
 
 #no_diff();
-#no_long_string();
+no_long_string();
 check_accum_error_log();
 run_tests();
 
@@ -1026,7 +1027,7 @@ capacity: 16384
 
 
 
-=== TEST 32: free_space, empty
+=== TEST 32: free_space, empty (16k zone)
 --- skip_nginx: 5: < 1.11.7
 --- http_config eval: $::HttpConfig
 --- config
@@ -1052,7 +1053,33 @@ free_page_bytes: 4096
 
 
 
-=== TEST 33: free_space, about half full, one page left
+=== TEST 33: free_space, empty (100k zone)
+--- skip_nginx: 5: < 1.11.7
+--- http_config eval: $::HttpConfig
+--- config
+    location = /t {
+        content_by_lua_block {
+            local birds = ngx.shared.birds
+            birds:flush_all()
+            birds:flush_expired()
+            local free_page_bytes = birds:free_space()
+            ngx.say("free_page_bytes type: ", type(free_page_bytes))
+            ngx.say("free_page_bytes: ", free_page_bytes)
+        }
+    }
+--- request
+GET /t
+--- response_body
+free_page_bytes type: number
+free_page_bytes: 90112
+--- no_error_log
+[error]
+[alert]
+[crit]
+
+
+
+=== TEST 34: free_space, about half full, one page left
 --- skip_nginx: 5: < 1.11.7
 --- http_config eval: $::HttpConfig
 --- config
@@ -1092,7 +1119,7 @@ free_page_bytes: 4096
 
 
 
-=== TEST 34: free_space, about half full, no page left
+=== TEST 35: free_space, about half full, no page left
 --- skip_nginx: 5: < 1.11.7
 --- http_config eval: $::HttpConfig
 --- config
@@ -1132,7 +1159,7 @@ free_page_bytes: 0
 
 
 
-=== TEST 35: free_space, full
+=== TEST 36: free_space, full
 --- skip_nginx: 5: < 1.11.7
 --- http_config eval: $::HttpConfig
 --- config
@@ -1172,7 +1199,7 @@ free_page_bytes: 0
 
 
 
-=== TEST 36: free_space, got forcible
+=== TEST 37: free_space, got forcible
 --- skip_nginx: 5: < 1.11.7
 --- http_config eval: $::HttpConfig
 --- config
@@ -1210,3 +1237,46 @@ free_page_bytes: 0
 [error]
 [alert]
 [crit]
+
+
+
+=== TEST 38: free_space, full (100k)
+--- skip_nginx: 5: < 1.11.7
+--- http_config eval: $::HttpConfig
+--- config
+    location = /t {
+        content_by_lua_block {
+            local birds = ngx.shared.birds
+            birds:flush_all()
+            birds:flush_expired()
+            for i = 1, 1000 do
+                local key = string.format("key%05d", i)
+                local val = string.format("val%05d", i)
+                local ok, err, forcible = birds:set(key, val)
+                if err ~= nil then
+                    ngx.say(string.format("got error, i=%d, err=%s", i, err))
+                end
+                if forcible then
+                    ngx.say(string.format("got forcible, i=%d", i))
+                    break
+                end
+                if not ok then
+                    ngx.say(string.format("got not ok, i=%d", i))
+                    break
+                end
+            end
+            local free_page_bytes = birds:free_space()
+            ngx.say("free_page_bytes type: ", type(free_page_bytes))
+            ngx.say("free_page_bytes: ", free_page_bytes)
+        }
+    }
+--- request
+GET /t
+--- response_body
+got forcible, i=736
+free_page_bytes type: number
+free_page_bytes: 0
+--- no_error_log
+[error]
+[alert]
+[crit]
```

hnakamur:
Thanks for your review and merging!

thibaultcha:
Very nice! Good job @hnakamur and thanks @agentzh

hnakamur:
@thibaultcha Thanks!

PR #145, opened 2017-09-22
refactor: make lua-resty-core support stream balancer and disable all

dndx:
other FFI APIs for now.


lib/resty/core/base.lua
```diff
@@ -5,6 +5,7 @@ local ffi = require 'ffi'
 local ffi_new = ffi.new
 local error = error
 local ceil = math.ceil
+local subsystem = ngx.config.subsystem
```
thibaultcha: Merely a suggestion, hoping it is not intrusive: since this is being called before running the version ngx_lua check (for obvious and valid reasons), the error will be somewhat hard to decipher for users if this module is being used with an older ngx_lua module. Should we fool-proof it (like we fool-proof the ngx_lua version below) with something like:

```lua
local subsystem
do
  local ok = pcall(function() subsystem = ngx.config.subsystem end)
  if not ok then
    -- 0.10.10 was the first version to support ngx.config.subsystem, but this module now needs 0.10.11 anyways
    error("ngx_lua 0.10.11+ required")
  end
end
```

I haven't tested this.

thibaultcha: Actually, this might be enough, `config` will not throw an error on an `nil` key iirc:
```lua
if not subsystem then
  error("ngx_lua 0.10.11+ required")
end
```

thibaultcha: (The below test is even more conservative in that it is even testing for the existence of `ngx.config`)

lib/resty/core/base.lua
```diff
@@ -13,10 +14,22 @@ local size_ptr
...
+    end
+
+elseif subsystem == 'stream' then
+    if not ngx.config
+       or not ngx.config.ngx_lua_version
+       or ngx.config.ngx_lua_version < 10009
+    then
+        error("stream_ngx_lua 0.10.09+ required")
+    end
+else
```
agentzh: Style: need a blank line before this line.

lib/resty/core/base.lua
```diff
@@ -72,28 +85,50 @@ if not pcall(ffi.typeof, "ngx_str_t") then
...
-    ]]
+elseif subsystem == 'stream' then
+    if not pcall(ffi.typeof, "ngx_stream_lua_request_t") then
+        ffi.cdef[[
+            struct ngx_stream_lua_request_s;
+            typedef struct ngx_stream_lua_request_s  ngx_stream_lua_request_t;
+        ]]
+    end
+
+
```
agentzh: Style: extra blank line. One is enough.

lib/ngx/balancer.lua
```diff
@@ -15,22 +16,62 @@ local getfenv = getfenv
...
+
+    int ngx_stream_lua_ffi_balancer_get_last_failure(
+        ngx_stream_lua_request_t *r, int *status, char **err);
+
+    int ngx_stream_lua_ffi_balancer_set_timeouts(ngx_stream_lua_request_t *r,
+        long connect_timeout, long timeout, char **err);
+    ]]
+
+    ngx_subsystem_lua_ffi_balancer_set_current_peer =
+    C.ngx_stream_lua_ffi_balancer_set_current_peer
```
agentzh: Need more indentations on the continued line. Please fix other similar places as well.

lib/resty/core/base.lua
```diff
@@ -174,6 +209,20 @@ function _M.ref_in_table(tb, key)
 end
 
 
+function _M.check_subsystem(...)
+    local subsystems = {...}
```
agentzh: Can we get rid of this Lua table? We can just use `select(i, ...)` to index individual arguments.

lib/ngx/balancer.lua
```diff
@@ -15,22 +16,62 @@ local getfenv = getfenv
...
+
+    int ngx_stream_lua_ffi_balancer_set_timeouts(ngx_stream_lua_request_t *r,
+        long connect_timeout, long timeout, char **err);
+    ]]
+
+    ngx_subsystem_lua_ffi_balancer_set_current_peer =
+        C.ngx_stream_lua_ffi_balancer_set_current_peer
+    ngx_subsystem_lua_ffi_balancer_set_more_tries =
+        C.ngx_stream_lua_ffi_balancer_set_more_tries
+    ngx_subsystem_lua_ffi_balancer_get_last_failure =
```
agentzh: Can we add a blank line between these assignments? It should look better.

t.stream/balancer.t
```diff
@@ -0,0 +1,220 @@
+# vim:set ft= ts=4 sw=4 et fdm=marker:
```
agentzh: I don't like the `t.stream` directory. Can we use `t/stream/` instead?

lib/resty/core/base.lua
```diff
@@ -174,6 +209,20 @@ function _M.ref_in_table(tb, key)
...
+    local total = select("#", ...)
+
+    for i = 1, total do
+        if subsystems[i] == subsystem then
+            return
+        end
+    end
+
+    error("unsupported subsystem: " .. subsystem)
```
agentzh: Seems like there is no test cases triggering this error?

t.stream/balancer.t
dndx: We can certainly do that. Only concern is that since the `stream-lua-nginx-module` containing the balancer implementation has not been released yet, putting it under `t/` will cause Travis tests to turn red. I think we can change this later but merge first?

lib/resty/core/base.lua
dndx: Added for both `stream` and `http` subsystem.

t.stream/balancer.t
agentzh: @dndx I'll take care of the merging thing in stream-lua-nginx-module.

.gitignore
```diff
@@ -3,8 +3,10 @@
 *~
 go
 t/servroot/
+t.stream/servroot/
```
agentzh: The C API for balancers is already merged into ngx_stream_lua_module's master. Let's change this.

lib/ngx/balancer.lua
```diff
@@ -1,10 +1,11 @@
 -- Copyright (C) Yichun Zhang (agentzh)
 
 
-local ffi = require "ffi"
 local base = require "resty.core.base"
+base.check_subsystem('http', 'stream')
```
agentzh: The `check_subsystem` is a little bit confusing. Maybe `allows_subsystem()` is a better name? Better be more explicit about our intention (like allows or not allows).

lib/ngx/balancer.lua
```diff
@@ -15,22 +16,66 @@ local getfenv = getfenv
...
+
+    int ngx_http_lua_ffi_balancer_get_last_failure(ngx_http_request_t *r,
+        int *status, char **err);
+
+    int ngx_http_lua_ffi_balancer_set_timeouts(ngx_http_request_t *r,
+        long connect_timeout, long send_timeout,
+        long read_timeout, char **err);
+    ]]
+
+    ngx_subsystem_lua_ffi_balancer_set_current_peer =
```
agentzh: Shall we use the prefix `ngx_lua_ffi_` instead of `ngx_subsystem_lua_ffi_`? The current prefix is just too long :)

Please change the other similar places.

lib/ngx/balancer.lua
```diff
@@ -135,10 +180,23 @@ function _M.set_timeouts(connect_timeout, send_timeout, read_timeout)
...
+
+    elseif subsystem == 'stream' then
+        local timeout = max(send_timeout, read_timeout)
+
+        rc = C.ngx_stream_lua_ffi_balancer_set_timeouts(r, connect_timeout,
+                                                        timeout, errmsg)
+
+    else
+        error("unknown subsystem: " .. subsystem)
+    end
```
agentzh: I don't like these branches in such a hot code path. Can we add an `ngx_lua_ffi_balancer_set_timeouts` function which is in the same spirit as the other dynamically dispatched helpers?

lib/ngx/errlog.lua
```diff
@@ -1,8 +1,10 @@
 -- Copyright (C) Yichun Zhang (agentzh)
```
agentzh: Style: need 2 blank lines here. Please fix other similar places.

lib/resty/core/base.lua
```diff
@@ -174,6 +210,19 @@ function _M.ref_in_table(tb, key)
 end
 
 
+function _M.check_subsystem(...)
```
agentzh: I think `allows_subsystem` is a better name here.

lib/ngx/errlog.lua
dndx: Fixed.

lib/ngx/balancer.lua
dndx: Problem is that those two functions have different signature since the stream version does not have as much timeout values as the http version.

dndx:
Closed in 1d975d521e7f85c8d800d806acd019ecad7b5c33.

PR #147, opened 2017-09-23
feature: add ssl.set_http_version API

vislee:

Note:  the unit test relying on the [test-nginx/pull/68](https://github.com/openresty/test-nginx/pull/68)

### DIFF
--- a/t/http2.t
+++ b/t/http2.t
@@ -1,0 +1,80 @@
+# vim:set ft= ts=4 sw=4 et fdm=marker:
+
+use Test::Nginx::Socket::Lua;
+use Cwd qw(cwd);
+
+log_level('debug');
+# server_port('443');
+server_name('test.com');
+
+repeat_each(2);
+
+plan tests => repeat_each() * blocks();
+
+our $CWD = cwd();
+
+#no_diff();
+#no_long_string();
+no_shuffle();
+
+$ENV{TEST_NGINX_USE_HTTP2} = 1;
+$ENV{TEST_NGINX_LUA_PACKAGE_PATH} = "$::CWD/lib/?.lua;;";
+
+run_tests();
+
+
+__DATA__
+
+
+=== TEST 1: test set http2
+--- http_config
+    lua_package_path "$TEST_NGINX_LUA_PACKAGE_PATH/?.lua;;";
+
+
+--- config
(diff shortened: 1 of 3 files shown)


t/http2.t
```diff
@@ -0,0 +1,80 @@
...
+    lua_package_path "$TEST_NGINX_LUA_PACKAGE_PATH/?.lua;;";
+
+
+--- config
+    ssl_certificate      ../../cert/test.crt;
+    ssl_certificate_key  ../../cert/test.key;
+
+    ssl_certificate_by_lua_block {
+        local ssl = require "ngx.ssl";
+        ssl.set_http_version(2);
```
agentzh: I think it's better to use the `ssl.use_http2(true)` and `ssl.use_http2(false)` here instead. We cannot really set other HTTP version numbers like 0.9, 1.0, and 1.1 here anyway.

t/http2.t
```diff
@@ -0,0 +1,80 @@
...
+
+
+=== TEST 1: test set http2
+--- http_config
+    lua_package_path "$TEST_NGINX_LUA_PACKAGE_PATH/?.lua;;";
+
+
+--- config
+    ssl_certificate      ../../cert/test.crt;
+    ssl_certificate_key  ../../cert/test.key;
```
agentzh: Better let the test scaffold automatically generate these under the hood. It's too much a burden for us test writers to write these in every HTTP/2 over TLS test cases ;)

agentzh:
@vislee I wonder if the protocol selection result in the TLS handshake can be remembered by subsequent new connections doing TLS session resumption (session ID or session ticket). I doubt it though. Have you tested the TLS session resumption case?

lziest:
I feel we need to really test session resumption, because in case of session resumption, the certificate callback will be skipped entirely. So I am a little skeptical about this PR right now.

agentzh:
@lziest Agreed. We need test cases for different kinds of session resumptions, SSL session IDs and TLS session tickets.

vislee:
@agentzh  @lziest  Thanks for review and comments. I will try to modify it. @agentzh  you can close this PR. Sorry!

agentzh:
@vislee You can close it yourself, BTW :) Closing this.

PR #148, opened 2017-10-03
consume errlen so we don't use garbage pass end of string

kahing:
counterpart of https://github.com/openresty/lua-nginx-module/pull/1165

### DIFF
--- a/lib/resty/core/var.lua
+++ b/lib/resty/core/var.lua
@@ -25,7 +25,7 @@ ffi.cdef[[
     int ngx_http_lua_ffi_var_set(ngx_http_request_t *r,
         const unsigned char *name_data, size_t name_len,
         unsigned char *lowcase_buf, const unsigned char *value,
-        size_t value_len, unsigned char *errbuf, size_t errlen);
+        size_t value_len, unsigned char *errbuf, size_t *errlen);
 ]]
 
 
@@ -84,8 +84,9 @@ local function var_set(self, name, value)
     end
     local name_len = #name
 
-    local errlen = 256
-    local lowcase_buf = get_string_buf(name_len + errlen)
+    local errlen = get_size_ptr()
+    errlen[0] = 256
+    local lowcase_buf = get_string_buf(name_len + errlen[0])
 
     local value_len
     if value == nil then
@@ -108,7 +109,7 @@ local function var_set(self, name, value)
     end
 
     if rc == -1 then  -- NGX_ERROR
-        return error(ffi_str(errbuf, errlen))
+        return error(ffi_str(errbuf, errlen[0]))
     end
 end
 


lib/resty/core/var.lua
```diff
@@ -108,7 +109,7 @@ local function var_set(self, name, value)
     end
 
     if rc == -1 then  -- NGX_ERROR
-        return error(ffi_str(errbuf, errlen))
+       return error(ffi_str(errbuf, errlen[0]))
```
agentzh: 3 space indentation?

agentzh:
@kahing I wonder if it's possible to add a test case to the existing test suite to actually cover this fix?

lib/resty/core/var.lua
kahing: fixed

kahing:
I considered that but I am not familiar with the test suite enough. This problem in theory only exhibits itself in the error log.

agentzh:
@kahing The test cases are declarative and the test scaffold is explained in detail in this tutorial:

And the test scaffold can surely check error logs, among many other things. Will you give it a shot?

kahing:
honestly not sure if I have time for that, this internally is a very low priority fix.

lib/resty/core/var.lua
```diff
@@ -108,7 +109,7 @@ local function var_set(self, name, value)
     end
 
     if rc == -1 then  -- NGX_ERROR
-        return error(ffi_str(errbuf, errlen))
+        return error(ffi_str(errbuf, errlen[0]))
     end
 end
```
agentzh: This fix should be covered by a test case like below:

```
diff --git a/t/var.t b/t/var.t
index bdb0e59..7fc1880 100644
--- a/t/var.t
+++ b/t/var.t
@@ -215,3 +215,28 @@ qr/\[TRACE   \d+ content_by_lua\(nginx\.conf:\d+\):3 loop\]/
 --- no_error_log
 [error]
  -- NYI:
+
+
+
+=== TEST 8: error buffer overread
+--- config
+    location = /test {
+            local ok1, err1 = pcall(function () ngx.var.foo = 32; end)
+            local ok2, err2 = pcall(function () ngx.var.server_port = 32; end)
+            assert(not ok1)
+            ngx.say(err1)
+            assert(not ok2)
+            ngx.say(err2)
+        }
+    }
+--- request
+GET /test
+--- response_body
+content_by_lua(nginx.conf:62):2: variable "foo" not found for writing; maybe it is a built-in variable that is not changeable or you forgot to use "set $foo '';" in the config file to define it first
+content_by_lua(nginx.conf:62):3: variable "server_port" not changeable
+--- no_error_log
+[error]
+[alert]
+ -- NYI:
```

agentzh:
@kahing Committed with commit c0dcdc657e4 with various fixes suggested above. Thanks for your contribution.

kahing:
thanks for pushing this over the finish line!

PR #149, opened 2017-10-04
bugfix: ngx.re: fix an edge-case where split() might not destroy  compiled regexes

thibaultcha:

I might be wrong, but it seems to me like we are missing some calls to `destroy_compiled_regex()` when we get the `"capture size too small"` error and when we exit naturally from the "non-max" loop (the second one).

This unifies the compiled regex destruction like so:
- Any error encountered while the regex executes makes `re_split_helper()` destroy the regex before returning `nil, err`
- If the loops terminate normally, `split()` will destroy the compiled regex before returning `res`

---

(The error tail call is a minor change I noticed while doing this change, and iirc, we try to avoid this pattern now - the combination of a tail call and a level argument of `2` would be problematic anyways)

### DIFF
--- a/lib/ngx/re.lua
+++ b/lib/ngx/re.lua
@@ -129,7 +129,7 @@ function _M.split(subj, regex, opts, ctx, max, res)
         res = new_tab(narr, 0)
 
     elseif type(res) ~= "table" then
-        return error("res is not a table", 2)
+        error("res is not a table", 2)
     end
 
     local len = #subj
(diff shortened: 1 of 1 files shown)


lib/ngx/re.lua
```diff
@@ -53,9 +53,6 @@ local function re_split_helper(subj, compiled, compile_once, flags, ctx)
     end
 
     if rc == PCRE_ERROR_NOMATCH then
-        if not compile_once then
-            destroy_compiled_regex(compiled)
-        end
```
agentzh: Why removing this?

agentzh: Never mind, it's the helper function.

agentzh:
Merged. Thanks!

PR #153, opened 2017-11-17
bugfix: library loaded by package.path should be installed under share/lua and LUA_VERSION is required.

spacewander:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,9 @@
 OPENRESTY_PREFIX=/usr/local/openresty
 
-#LUA_VERSION := 5.1
+LUA_VERSION := 5.1
 PREFIX ?=          /usr/local
 LUA_INCLUDE_DIR ?= $(PREFIX)/include
-LUA_LIB_DIR ?=     $(PREFIX)/lib/lua/$(LUA_VERSION)
+LUA_LIB_DIR ?=     $(PREFIX)/share/lua/$(LUA_VERSION)
 INSTALL ?= install
 
 .PHONY: all test install


Makefile
```diff
@@ -1,9 +1,9 @@
 OPENRESTY_PREFIX=/usr/local/openresty
 
-#LUA_VERSION := 5.1
+LUA_VERSION := 5.1
 PREFIX ?=          /usr/local
 LUA_INCLUDE_DIR ?= $(PREFIX)/include
-LUA_LIB_DIR ?=     $(PREFIX)/lib/lua/$(LUA_VERSION)
+LUA_LIB_DIR ?=     $(PREFIX)/share/lua/$(LUA_VERSION)
```
agentzh: No, no, better not change the default values of Makefile since these Makefile variables can always be overridden on the command line anyway, like this:

```bash
make LUA_LIB_DIR=/usr/local/share/lua/5.1
```

spacewander: @agentzh 
The default value is incorrect. If we run `make install` directly, the lua library will be installed to `/usr/local/lib/lua/`, which is out of `package.path`.

agentzh: @spacewander I think this is operating system specific and better not touch it. This Makefile is never meant to be used by the end user anyway. It's mainly for the OpenResty package building toolchain.

agentzh: @spacewander For local development, just set the `lua_package_path` to point to your `lua-resty-core` local checkout directory. Do not install it.

spacewander: @agentzh 
It is not operating system specific.
The default `package.path` doesn't include `/usr/local/lib/lua`: https://github.com/openresty/luajit2/blob/master/src/luaconf.h#L61
So the default Makefile variable doesn't work with the default `package.path` value in Lua.

PR #155, opened 2017-11-27
feature : add a shortcut to ngx.re.split and ngx.re.opt

chronolaw:
ngx.re.split is a very nice method, but before we use it we must write 'require', it feels a little noisy.

So I put some code in 'resty.core.lua', add 'split' and 'opt' to the table 'ngx.re'. Then we can use split method just like its brothers 'match' or 'gmatch'.

For example:
ngx.re.split(...)   -- do not need any require

Do you think it is useful or helpful?

### DIFF
--- a/lib/resty/core.lua
+++ b/lib/resty/core.lua
@@ -17,6 +17,10 @@ if subsystem == 'http' then
     require "resty.core.response"
     require "resty.core.time"
     require "resty.core.worker"
+
+    local ngx_re = require "ngx.re"
+    ngx.re.split = ngx_re.split
+    ngx.re.opt = ngx_re.opt
 end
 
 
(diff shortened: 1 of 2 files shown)


lib/resty/core.lua
```diff
@@ -17,6 +17,10 @@ if subsystem == 'http' then
     require "resty.core.response"
     require "resty.core.time"
     require "resty.core.worker"
+
+    local ngx_re = require "ngx.re"
+    ngx.re.split = ngx_re.split
+    ngx.re.opt = ngx_re.opt
```
agentzh: Actually we want to retire the `ngx.re.xxx` way at some point in the future. This is because for applications that do not need `ngx.re`, there is currently no way to avoid its GC overhead and memory overhead.

The future is to modularize the Lua API instead of growing the flat API namespace and tables forever.

PR #156, opened 2017-11-27
Update balancer.t

gnought:
This PR could be applied after the lua-resty-core supports nginx v1.13.7. The patch https://github.com/nginx/nginx/commit/f1be23bc8aed6c6e6bbba1320d85a8943550b253 fixes the proxy returns 503/504 rather than 502 if proxy_next_upstream includes http_503/http_504.

### DIFF
--- a/t/balancer.t
+++ b/t/balancer.t
@@ -332,7 +332,7 @@ last peer failure: failed 500
 --- response_body_like: 503 Service Temporarily Unavailable
 --- error_code: 503
 --- grep_error_log eval: qr{last peer failure: \S+ \S+}
---- grep_error_log_out
+--- grep_error_log_out eval
 qr{last peer failure: nil nil
 last peer failure: failed 50[23]
 last peer failure: failed 50[23]}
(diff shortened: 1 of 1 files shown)


t/balancer.t
```diff
@@ -334,8 +334,8 @@ last peer failure: failed 500
 --- grep_error_log eval: qr{last peer failure: \S+ \S+}
 --- grep_error_log_out
 last peer failure: nil nil
-last peer failure: failed 502
-last peer failure: failed 502
+last peer failure: failed 503
+last peer failure: failed 503
```
agentzh: Maybe we should make this an regex so that it also passes with older versions of the nginx core?

gnought: Sure, I made a change.

t/balancer.t
```diff
@@ -332,10 +332,10 @@ last peer failure: failed 500
 --- response_body_like: 503 Service Temporarily Unavailable
 --- error_code: 503
 --- grep_error_log eval: qr{last peer failure: \S+ \S+}
---- grep_error_log_out
-last peer failure: nil nil
-last peer failure: failed 502
-last peer failure: failed 502
+--- grep_error_log_out eval
+qr{last peer failure: nil nil
```
agentzh: We should use `\A` and `\z` to anchor the match otherwise any unexpected leading or trailing garbage cannot be detected.

agentzh:
@gnought Merged with minor fixes. Thanks!

PR #157, opened 2017-12-19
feature: added urlsafe base64 encoding/decoding functions.

dndx:


agentzh:
@dndx Travis is failing due to missing dependencies, it seems.

lib/ngx/encoding.lua
```diff
@@ -0,0 +1,76 @@
...
+ngx_int_t ngx_decode_base64url(ngx_str_t *dst, ngx_str_t *src);
+]]
+
+
+local get_string_buf = base.get_string_buf
+
+
+local dst = ffi.new("ngx_str_t[1]")
+local src = ffi.new("ngx_str_t[1]")
+
```
agentzh: Needs 2 blank lines here.

lib/ngx/encoding.lua
```diff
@@ -0,0 +1,76 @@
...
+void ngx_encode_base64url(ngx_str_t *dst, ngx_str_t *src);
+ngx_int_t ngx_decode_base64url(ngx_str_t *dst, ngx_str_t *src);
+]]
+
+
+local get_string_buf = base.get_string_buf
+
+
+local dst = ffi.new("ngx_str_t[1]")
+local src = ffi.new("ngx_str_t[1]")
```
agentzh: Better use longer names for such top-level Lua variables.

lib/ngx/encoding.lua
```diff
@@ -0,0 +1,76 @@
...
+end
+
+
+local function base64_decoded_length(len)
+    return ((len + 3) / 4) * 3
+end
+
+
+function _M.encode_base64url(s)
+    local len = #s
```
agentzh: It's safer to do `tostring(s)` before trying to use it as a string.

lib/ngx/encoding.lua
```diff
@@ -0,0 +1,76 @@
...
+    dst[0].len = trans_len
+
+    C.ngx_encode_base64url(dst, src)
+
+    return ffi_str(dst[0].data, dst[0].len)
+end
+
+
+function _M.decode_base64url(s)
+    local len = #s
```
agentzh: Ditto.

lib/ngx/encoding.lua
```diff
@@ -0,0 +1,76 @@
...
+    return ((len + 3) / 4) * 3
+end
+
+
+function _M.encode_base64url(s)
+    local len = #s
+    local trans_len = base64_encoded_length(len)
+
+    src[0].data = s
+    src[0].len = len
```
agentzh: Better cache the value `src[0]` in a local variable since it is referenced for more than one time.

lib/ngx/encoding.lua
```diff
@@ -0,0 +1,76 @@
...
+
+    dst[0].data = get_string_buf(trans_len)
+    dst[0].len = trans_len
```
agentzh: Ditto.

t/encode-base64.t
```diff
@@ -225,3 +225,38 @@ qr/\[TRACE   \d+ content_by_lua\(nginx\.conf:\d+\):3 loop\]/
...
+
+=== TEST 9: encode_base64url
+--- http_config eval: $::HttpConfig
+--- config
+    location = /t {
+        content_by_lua_block {
+            local enc = require("ngx.encoding")
+
+            -- RFC 4648 test vectors
+            ngx.say("encode_base64url(\"\") = \"", enc.encode_base64url(""), "\"")
```
agentzh: Better use `"a\0b"` as a test input string?

lib/ngx/encoding.lua
dndx: Fixed.

t/encode-base64.t
dndx: Added.

t/decode-base64.t
```diff
@@ -9,7 +9,7 @@ use Cwd qw(cwd);
 
 repeat_each(2);
 
-plan tests => repeat_each() * (blocks() * 5);
+plan tests => repeat_each() * ((blocks() * 5) - 2);
```
agentzh: Seems like we can save the parens around the `*` expression since it's obvious that `*` takes precedence over `-`.

lib/ngx/encoding.lua
```diff
@@ -0,0 +1,92 @@
+-- Copyright (C) by Yichun Zhang (agentzh)
```
agentzh: I don't like the `ngx.encoding` module name. Can we make it `ngx.base64`?

Also, we need some docs for this module :)

dndx: Changed. But I was thinking about `ngx.encoding` so that we may be able to add more variants of encodings in the future without introducing a new module.

agentzh: @dndx I prefer smaller Lua modules since it can save memory footprint and GC overhead by skip loading stuff the user does not need.

lib/ngx/base64.lua
```diff
@@ -0,0 +1,92 @@
+-- Copyright (C) by Yichun Zhang (agentzh)
+-- Copyright (C) by OpenResty Inc.
+
+-- I hereby assign copyright in this code to the lua-resty-core project,
```
agentzh: We do not need to include this statement.

agentzh:
@dndx We need some docs for this `ngx.base64` module in the file `lib/ngx/base64.md` (better link this markdown file to the README file as well).

dndx:
@agentzh ping

agentzh:
@dndx Merged. Thanks!

PR #89, opened 2017-02-21
feature: ssl.create_ctx and tcp:setsslctx

detailyang:
This is a sister PR at [lua-nginx-module/pull/997](https://github.com/openresty/lua-nginx-module/pull/997).

I refer to [Node.js JS Design](https://nodejs.org/api/tls.html#tls_tls_createsecurecontext_options) as a guide. Now the API is the fowllong:

```lua
local ssl = require "ngx.ssl"
local ctx, err = ssl.create_ctx{
    method = "SSLv23_method", #default SSLv23_method
    cert = "pem-data", #parsed by ssl.parse_pem_cert
    priv_key = "pem-data", #parsed by ssl.parse_pem_priv_key
}

if ctx ~= nil then
    return error("create ssl ctx error" .. err)
end

-- we can cache the ctx object in a cache like lua-resty-lrucache

local sock = ngx.socket.tcp()
local ok, err = sock:setsslctx(ctx) #the setsslctx is a Lua function atop FFI whcich been implement on lua-resty-core
sock:sslhandshake(...)
```

Now we proivde 3 arguments as the follwoing to feed `SSL_CTX` object.

* method 
Openssl protocols are listed as [OPENSSL PROTOCOL METHODS](https://www.openssl.org/docs/man1.0.2/ssl/ssl.html#DEALING-WITH-PROTOCOL-METHODS)
* cert
Optional certificate in PEM format which can be parsed by `ssl.parse_pem_cert`
* priv_key
Optional private keys in PEM format which can be parsed by `ssl.parse_pem_priv_key`

It 's convenient that if we want to expose more arguments to user in future if we want.

Now it is not ready to provide the README document. when we agree on the API Design, I will provide.

@agentzh @thibaultcha @doujiang24 

Can have a look at this PR ? Many thanks :D

PR #158, opened 2017-12-28
ngx.process.master_pid

chronolaw:
Here is the new API `ngx.process.master_pid`, which needs ffi C function `ngx_http_lua_ffi_master_pid` in ngx_lua.

I also write a test case in` t/process-master-pid.t`, But I am not familar with Perl and have not found the detail API documents of test::nginx, so I just copied `worker.t` then modified on it.

That test case may has error, please help me to correct it. Thanks very much.

### DIFF
--- a/t/process-master-pid.t
+++ b/t/process-master-pid.t
@@ -68,7 +68,7 @@ GET /t
 ^true
 \d+$
 --- error_log eval
-qr/\[TRACE   \d+ content_by_lua\(nginx\.conf:\d+\):4 loop\]/
+qr/\[TRACE   \d+ content_by_lua\(nginx\.conf:\d+\):\d loop\]/
 --- no_error_log
 [error]
  -- NYI:
--- a/t/process-master-pid.t
+++ b/t/process-master-pid.t
@@ -50,12 +50,22 @@ __DATA__
             for i = 1, 400 do
                 v = pid()
             end
+
+            local f = io.open(ngx.config.prefix().."/logs/nginx.pid", "r")
+            if not f then
+                ngx.say(false)
+            else
+                local str = f:read("*l")
+                ngx.say(v == tonumber(str or "0"))
+                f:close()
+            end
             ngx.say(v)
         }
     }
 --- request
 GET /t
 --- response_body_like chop
+^true
 \d+$
 --- error_log eval
 qr/\[TRACE   \d+ content_by_lua\(nginx\.conf:\d+\):4 loop\]/
--- a/t/process-master-pid.t
+++ b/t/process-master-pid.t
@@ -1,0 +1,68 @@
+# vim:set ft= ts=4 sw=4 et fdm=marker:
+use lib 'lib';
+use Test::Nginx::Socket::Lua;
+use Cwd qw(cwd);
+
+#worker_connections(1014);
+#master_process_enabled(1);
+#log_level('warn');
+
+repeat_each(2);
+
+plan tests => repeat_each() * (blocks() * 6);
+
+my $pwd = cwd();
+
+our $HttpConfig = <<_EOC_;
+    lua_shared_dict dogs 1m;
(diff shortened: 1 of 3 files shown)


membphis:
you need to change the travis config, use your own branch to test.

https://github.com/openresty/lua-resty-core/blob/master/.travis.yml#L62

t/process-master-pid.t
```diff
@@ -0,0 +1,68 @@
...
+            for i = 1, 400 do
+                v = pid()
+            end
+            ngx.say(v)
+        }
+    }
+--- request
+GET /t
+--- response_body_like chop
+\d+$
```
agentzh: This test is too loose. We could do better here by comparing the number returned by `master_pid()` against the real master pid saved in the nginx pid file.

t/process-master-pid.t
```diff
@@ -0,0 +1,68 @@
...
+
+=== TEST 1: ngx.process.master_pid
+--- http_config eval: $::HttpConfig
+--- config
+    location = /t {
+        content_by_lua_block {
+            local process = require "ngx.process"
+
+            local v
+            local pid = process.master_pid
```
agentzh: This is definitely wrong. You missed a pair of parentheses after `process.master_pid`. Have you actually run this test yourself?

This is another reason why you should make your PR pass the travis ci checks as @membphis suggested ;)

chronolaw: Sorry, I'm not familar with travis ci, I will try to correct it ,thanks.

chronolaw: And here is not an error ,because `pid` will be called in then next line ：
`local v = pid()`

chronolaw:
I have run test in my own nginx 1.13.8+ngx_lua，and I will try to config travis ci after soon.

chronolaw:
I have tried travis-ci build, but I got an error.

Then reason is that the `process.master_pid` needs nginx 1.13.8, but now OpenResty build will fail on it(no proper patch for it).

And in nginx 1.13.6, this API will return nil for no `ngx_parent` in nginx core.

Would you give me some advise to write the test case?

chronolaw:
I created a new git branch to fit nginx 1.13.6, now `process.master_pid` passed ,but got another error
```
#   Failed test 'TEST 1: ngx.process.master_pid - pattern "(?^:\[TRACE   \d+ content_by_lua\(nginx\.conf:\d+\):4 loop\])" should match a line in error.log (req 0)'
#   at /home/travis/build/chronolaw/lua-resty-core/test-nginx/lib/Test/Nginx/Socket.pm line 1206.
#   Failed test 'TEST 1: ngx.process.master_pid - pattern "(?^:\[TRACE   \d+ content_by_lua\(nginx\.conf:\d+\):4 loop\])" should match a line in error.log (req 1)'
```
So could you tell my how to write the right `error_log eval` in xxx.t?

chronolaw:
After some modify, my own test has passed, please see:https://travis-ci.org/chronolaw/lua-resty-core/jobs/324018825

Changes:
use `chronolaw/lua-nginx-module.git` and branch `test_ffi_master_pid`. In this branch code is 
```
#if (nginx_version >= 1013008)
    return (int) ngx_parent;
#else
    return (int) ((ngx_process == NGX_PROCESS_SINGLE) ? ngx_pid : getppid());
    //return -1;
#endif

```

agentzh:
@chronolaw You can skip the tests for nginx cores older than 1.13.8 with the `--- skip_nginx` section. You can see such examples in this library's existing test suite.

agentzh:
@chronolaw And you should point to your own fork of lua-nginx-module in `.travis.yml` in your branch, see others' PRs for such examples, like this:

https://github.com/openresty/lua-resty-core/pull/89

agentzh:
Merged by fixing the aforementioned issues myself. Thanks!

PR #159, opened 2017-12-28
bugfix: ngx.re: fixed a split() edge-case when using control characters in the regex.

thibaultcha:

Proposed fix to openresty/lua-nginx-module#1217.

### DIFF
--- a/lib/ngx/re.lua
+++ b/lib/ngx/re.lua
@@ -50,7 +50,7 @@ local _M = { version = base.version }
 local function re_split_helper(subj, compiled, compile_once, flags, ctx)
     local rc
     do
-        local pos = math_max(ctx.pos - 1, 0)
+        local pos = math_max(ctx.pos, 0)
 
         rc = C.ngx_http_lua_ffi_exec_regex(compiled, flags, subj, #subj, pos)
     end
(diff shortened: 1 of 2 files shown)


lib/ngx/re.lua
```diff
@@ -153,6 +160,9 @@ function _M.split(subj, regex, opts, ctx, max, res)
     local res_idx = 0
     local last_empty_match
 
+    -- update to split_helper indexes
+    ctx.pos = ctx.pos - 1
```
agentzh: Should we use `sub_idx` here on the RHS to avoid a duplicate `ctx.pos` table lookup?

lib/ngx/re.lua
```diff
@@ -87,11 +87,18 @@ local function re_split_helper(subj, compiled, compile_once, flags, ctx)
         return nil, nil, nil
     end
 
+    if from == to then
+      -- empty match, skip to next char
+      ctx.pos = to + 1
+
+    else
+      ctx.pos = to
```
agentzh: Please use 4-space indentations consistently.

thibaultcha: Sorry, pushed this fix from an usual setup while on holidays, will update.

lib/ngx/re.lua
thibaultcha: Fine with me, updated.

thibaultcha:
@agentzh Ping for 1.13.8.1, I don't think we should leave bugfixes behind :/ I'm hoping this API will finally stabilize now...

agentzh:
@thibaultcha Merged. Thanks!

PR #161, opened 2018-01-04
docs : sample code for get_master_pid

chronolaw:


lib/ngx/process.md
```diff
@@ -32,7 +32,7 @@ The API is still in flux and may change in the future without notice.
 Synopsis
 ========
 
-Enables privileged agent process and get process type
+Enables privileged agent process, get process type and get master process pid
 -----------------------------------------------------
```
agentzh: Hmm, this section name is too long. Maybe we should make it normal text instead of a title.

lib/ngx/process.md
```diff
@@ -80,6 +81,7 @@ The example location above produces the following response body:
 
 ```
 process type: worker
+master process pid: xxx
```
agentzh: It might be better to use an arbitrary number here instead of `xxx`?

agentzh:
@chronolaw Merged with some edits. Thanks!

PR #164, opened 2018-01-08
feature: reimplement the remain time relative API with FFI

spacewander:

### DIFF
--- a/lib/resty/core/time.lua
+++ b/lib/resty/core/time.lua
@@ -13,7 +13,6 @@ local ffi_str = ffi.string
 local get_size_ptr = base.get_size_ptr
 local get_string_buf = base.get_string_buf
 local ngx = ngx
-local subsystem = ngx.config.subsystem
 local FFI_ERROR = ngx.ERROR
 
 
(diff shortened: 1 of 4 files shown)


lib/resty/core/time.lua
```diff
@@ -26,6 +41,81 @@ function ngx.time()
...
 
+function ngx.update_time()
+    C.ngx_http_lua_ffi_update_time()
+end
+
+
+function ngx.today()
+    -- the format of today is 2010-11-19
+    local today_buf_size = 10
+    local buf = get_string_buf(today_buf_size)
```
membphis: if the buffer size is less than `4096`, this buffer can be placed to top level.

https://github.com/spacewander/lua-resty-core/blob/9e4147e8a311a781a478e57fda0c547e9fe5bb82/lib/resty/core/base.lua#L182

spacewander: @membphis 
I am afraid this change will make code hard to comprehend. All those top-level buffer actually point to one buffer, which is counter-intuitive. And function call is cheap.

agentzh: @spacewander We should reuse the `resty.core.base` buffer instead of creating an extra cdata buffer. The former is readily usable at this point anyway.

agentzh:
@spacewander Merged with fixes in your commit log message (there are several English grammar mistakes ;))

lib/resty/core/time.lua
membphis: @agentzh I means we can cache the `resty.core.base` buffer at top-level, then we can use it directly to avoid an extra function call.

```lua
local buf = base.get_string_buf(4096)

-- some code

function ngx.today()
    -- the format of today is 2010-11-19
    local today_buf_size = 10
    C.ngx_http_lua_ffi_today(buf)
    return ffi_str(buf, today_buf_size)
end

```

agentzh: @membphis Your way can be more expensive since the size of the string buffer of `resty.core.base` is tunable, which means it is not necessarily 4096. So you may still allocate a separate memory buffer than necessary when the actual buffer size is smaller than 4096 while the `today()` function only needs 10 bytes, for example.

I think this is a case of over-optimization. The LuaJIT compiler is usually smart enough to inline the Lua function call and optimize the linear code paths reducing branch mispredictions.

PR #165, opened 2018-01-08
feature: shdict:incr(): added the "exptime" argument to set the expiry of values when they are first created via the "init" argument.

thibaultcha:

Implement the `exptime` argument in `shdict:incr()` as discussed in openresty/openresty#328. Contrary to openresty/lua-resty-core#10, this is only implemented in the FFI counterpart of the shdict API.

If deemed necessary (e.g., for backwards compatibility reasons), I have a commit with support for this argument in the CFunction as well. As per recent guidance, this should apparently not be necessary.

We only set/support the `exptime` argument when the `init` argument is given/takes effect, as we do not wish to tamper with the expiry time of values created by other actors than `incr()`.

Sister PR at openresty/lua-nginx-module#1226

### DIFF
--- a/lib/resty/core/shdict.lua
+++ b/lib/resty/core/shdict.lua
@@ -29,7 +29,7 @@ ffi.cdef[[
 
     int ngx_http_lua_ffi_shdict_incr(void *zone, const unsigned char *key,
         size_t key_len, double *value, char **err, int has_init, double init,
-        int *forcible);
+        long init_ttl, int *forcible);
 
     int ngx_http_lua_ffi_shdict_store(void *zone, int op,
         const unsigned char *key, size_t key_len, int value_type,
@@ -337,7 +337,7 @@ local function shdict_get_stale(zone, key)
 end
 
 
-local function shdict_incr(zone, key, value, init)
+local function shdict_incr(zone, key, value, init, init_ttl)
     zone = check_zone(zone)
 
     if key == nil then
@@ -361,8 +361,6 @@ local function shdict_incr(zone, key, value, init)
     end
     num_value[0] = value
 
-    local has_init
-
     if init then
         local typ = type(init)
         if typ ~= "number" then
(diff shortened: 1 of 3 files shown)


lib/resty/core/shdict.lua
```diff
@@ -380,9 +380,31 @@ local function shdict_incr(zone, key, value, init)
...
+                error("bad exptime arg: number expected, got " .. typ, 2)
+            end
+        end
+
+        if exptime < 0 then
+            error('bad "exptime" argument', 2)
+        end
+
+        if has_init == 0 then
+            error('must provide "init" when providing "exptime"', 2)
```
thibaultcha: This is a new case of an shdict API both creating values and updating already existing ones, and now having to deal with conditionally setting their exptime as well. 
Unlike `set()` and `expire()` which unconditionally replace the exptime, I made `incr()` only replace it when the value did not exist. If you believe this does not make sense, let me know.

lib/resty/core/shdict.lua
```diff
@@ -29,7 +29,7 @@ ffi.cdef[[
 
     int ngx_http_lua_ffi_shdict_incr(void *zone, const unsigned char *key,
         size_t key_len, double *value, char **err, int has_init, double init,
-        int *forcible);
+        int exptime, int *forcible);
```
agentzh: `exptime` is confusing. We should call it `init_ttl`?

lib/resty/core/shdict.lua
```diff
@@ -337,7 +337,7 @@ local function shdict_get_stale(zone, key)
 end
 
 
-local function shdict_incr(zone, key, value, init)
+local function shdict_incr(zone, key, value, init, exptime)
```

lib/resty/core/shdict.lua
```diff
@@ -380,9 +380,31 @@ local function shdict_incr(zone, key, value, init)
...
+            if not exptime then
+                error("bad exptime arg: number expected, got " .. typ, 2)
+            end
+        end
+
+        if exptime < 0 then
+            error('bad "exptime" argument', 2)
+        end
+
+        if has_init == 0 then
```
agentzh: We should use a boolean typed value here instead of using an integer? It's Lua anyway. Seems like an old issue.

thibaultcha: Do you want me to update this logic in the above handling of the `init` argument in the same commit to use a boolean instead?

agentzh: @thibaultcha Yes, please.

thibaultcha: @agentzh Updated, got rid of the `has_init` variable altogether.

lib/resty/core/shdict.lua
```diff
@@ -372,22 +370,39 @@ local function shdict_incr(zone, key, value, init)
...
 
     else
-        has_init = 0
-        init = 0
+        init_ttl = 0
     end
 
     local rc = C.ngx_http_lua_ffi_shdict_incr(zone, key, key_len, num_value,
-                                              errmsg, has_init, init,
+                                              errmsg, init and 1 or 0,
```
agentzh: OK, I now see why it used `has_init` as an integer in the first place...

thibaultcha: Yes... However, this patch still looks like an improvement to me though.

agentzh: @thibaultcha I didn't disagree :)

thibaultcha: 😅

lib/resty/core/shdict.lua
```diff
@@ -372,22 +370,39 @@ local function shdict_incr(zone, key, value, init)
...
+            end
+        end
+
+        if init_ttl < 0 then
+            error('bad "init_ttl" argument', 2)
+        end
+
+        if not init then
+            error('must provide "init" when providing "init_ttl"', 2)
+        end
```
agentzh: I'm not really happy with such complex argument value check. It looks branchy and expensive. Will you simplify it in a future PR? Thanks!

agentzh:
Merged with comments :) Thanks!

PR #166, opened 2018-01-09
docs: add context for ngx.re API

chronolaw:
Does the document of ngx.re miss the context field?


agentzh:
@chronolaw Where is no `context` field present, then it means *any* :)

chronolaw:
But in `base64.md` and `process.md` there are `context`, maybe it will let the reader more clear.

chronolaw:
Another issue:  now new ffi-based ngx.re.* can work well in `init_by_lua*`, Shall we add notice in `ngx_lua`'s document?

PR #171, opened 2018-01-11
tests: corrected the handshake flag position.

spacewander:
We should set the handshake done flag when the handshake is totally
finished, not just before the first handshake.
The incorrect handshake flag caused some random test failures seen in
travis build and developers' machine.


agentzh:
@spacewander Merged. Thanks!

PR #172, opened 2018-01-25
bugfix: ngx.re: fixed a split() edge-case when remaining characters is matched

xiaocang:
when `re_split_helper` traverse the subject string, the remaining characters are push to array `res`, forgot to check the remaining characters whether matched by regex.

reproduce
---

``` lua
local res, err = ngx_re.split("a,b,cd,,,", ",")

ngx.say(#res)  
-- output is 5
-- it shoule be 6 here

ngx.say(table.concat(res, "|")) 
-- outout is a|b|cd||, 
-- it should be a|b|cd|||
```

### DIFF
--- a/lib/ngx/re.lua
+++ b/lib/ngx/re.lua
@@ -251,20 +251,18 @@ function _M.split(subj, regex, opts, ctx, max, res)
     -- trailing nil for non-cleared res tables
 
     -- delete empty trailing ones (without max)
-    local remaining_str = sub(subj, sub_idx)
-
-    if max <= 0 and remaining_str == "" then
+    if max <= 0 and sub_idx > len then
         for ety_idx = res_idx, 1, -1 do
-            if res[ety_idx] == "" then
-                res[ety_idx] = nil
-            else
+            if res[ety_idx] ~= "" then
                 res_idx = ety_idx
                 break
             end
+
+            res[ety_idx] = nil
         end
     else
         res_idx = res_idx + 1
-        res[res_idx] = remaining_str
+        res[res_idx] = sub(subj, sub_idx)
     end
 
     res[res_idx + 1] = nil
--- a/lib/ngx/re.lua
+++ b/lib/ngx/re.lua
@@ -201,8 +201,6 @@ function _M.split(subj, regex, opts, ctx, max, res)
 
                 if sub_idx > len then
                     break
-                elseif sub_idx == len and last_empty_match then
-                    break
                 end
             end
         end
@@ -240,8 +238,6 @@ function _M.split(subj, regex, opts, ctx, max, res)
 
                 if sub_idx > len then
                     break
-                elseif sub_idx == len and last_empty_match then
-                    break
                 end
             end
         end
@@ -263,14 +259,15 @@ function _M.split(subj, regex, opts, ctx, max, res)
                 res[ety_idx] = nil
             else
                 res_idx = ety_idx
-                res[res_idx + 1] = nil
                 break
             end
         end
     else
-        res[res_idx + 1] = remaining_str
-        res[res_idx + 2] = nil
+        res_idx = res_idx + 1
+        res[res_idx] = remaining_str
     end
+
+    res[res_idx + 1] = nil
 
     return res
 end
--- a/lib/ngx/re.lua
+++ b/lib/ngx/re.lua
@@ -199,7 +199,9 @@ function _M.split(subj, regex, opts, ctx, max, res)
 
                 sub_idx = to
 
-                if sub_idx >= len then
+                if sub_idx > len then
+                    break
+                elseif sub_idx == len and last_empty_match then
                     break
                 end
             end
@@ -236,7 +238,9 @@ function _M.split(subj, regex, opts, ctx, max, res)
 
                 sub_idx = to
 
-                if sub_idx >= len then
+                if sub_idx > len then
+                    break
+                elseif sub_idx == len and last_empty_match then
                     break
                 end
             end
--- a/lib/ngx/re.lua
+++ b/lib/ngx/re.lua
@@ -258,18 +258,15 @@ function _M.split(subj, regex, opts, ctx, max, res)
     local remaining_str = sub(subj, sub_idx)
 
     if max <= 0 and remaining_str == "" then
-        local ety_idx = res_idx
-
         for ety_idx = res_idx, 1, -1 do
             if res[ety_idx] == "" then
                 res[ety_idx] = nil
             else
+                res_idx = ety_idx
+                res[res_idx + 1] = nil
                 break
             end
         end
-
-        res_idx = ety_idx
-        res[res_idx + 1] = nil
     else
         res[res_idx + 1] = remaining_str
         res[res_idx + 2] = nil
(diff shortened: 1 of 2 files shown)


t/re-split.t
```diff
@@ -1245,3 +1245,35 @@ GET /re
 qr/\[TRACE   \d+/
 --- no_error_log
 [error]
+
+
+
+=== TEST 38: remaining characters are matched by regex
```
thibaultcha: It'd be more maintainable if we split the 2 test cases (with and without max) in two separate tests - one could fail without the other one failing, this helps narrowing down the issue for a maintainer tackling it.

lib/ngx/re.lua
```diff
@@ -199,7 +199,9 @@ function _M.split(subj, regex, opts, ctx, max, res)
 
                 sub_idx = to
 
-                if sub_idx >= len then
+                if sub_idx > len then
+                    break
+                elseif sub_idx == len and last_empty_match then
```
thibaultcha: I tested this locally and this patch seems unnecessary,  this code path (with max) seems unaffected by this behavior, and succeeds without this patch.

t/re-split.t
```diff
@@ -1245,3 +1245,35 @@ GET /re
...
+                end
+
+                ngx.say(#res, " ", table.concat(res, "|"))
+            end
+        }
+    }
+--- request
+GET /re
+--- response_body
+6 a|b|cd|||
```
thibaultcha: This isn't perl's `split()` behavior:
```
$ perl -e 'print join "|", split /,/, "a,b,cd,,,", 0'
a|b|cd
```

lib/ngx/re.lua
xiaocang: test case like re-split.t TEST 39:

``` lua
local res, err = ngx_re.split("a,b,cd,,,", ",", 6)

ngx.say(#res)  
-- output is 5
-- it shoule be 6 here

ngx.say(table.concat(res, "|")) 
-- outout is a|b|cd||, 
-- it should be a|b|cd|||
```

perl's behavior
``` bash
perl -le 'print join "|", split ",", "a,b,cd,,,", 6'
a|b|cd|||
```

xiaocang:
I notice that in perl split:

> By default, empty leading fields are preserved, and empty trailing ones are deleted. (If all fields are empty, they are considered to be trailing.)

So I delete the empty trailling ones when max is not specified, when max is specified, I preseved the empty trailling ones.

---

**perl's behavior**

``` perl
#!/usr/bin/env perl
use strict;
use warnings;

my $str = "a,b,cd,,,";

# TEST 38: remaining characters are matched by regex (without max)
# without max
print(join("|", split(",", $str)), "\n");

# TEST 39: remaining characters are matched by regex (with max)
# with max
for my $ix (1..7) {
    print(join("|", split(",", $str, $ix)), "\n");
}
```

lib/ngx/re.lua
thibaultcha: Apologies, I had the following patch when testing this locally and failed to spot the difference with diff here:
```diff
diff --git a/lib/ngx/re.lua b/lib/ngx/re.lua
index 06b5e9b..6ccc958 100644
--- a/lib/ngx/re.lua
+++ b/lib/ngx/re.lua
@@ -199,11 +199,17 @@ function _M.split(subj, regex, opts, ctx, max, res)

                 sub_idx = to

-                if sub_idx >= len then
+                if sub_idx > len then
                     break
                 end
```

This patch suffices to make TEST 38 and 39 pass :)

xiaocang: :)

lib/ngx/re.lua
```diff
@@ -250,8 +254,23 @@ function _M.split(subj, regex, opts, ctx, max, res)
...
-    res[res_idx + 2] = nil
+    -- delete empty trailing ones (without max)
+    local remaining_str = sub(subj, sub_idx)
+
+    if max <= 0 and remaining_str == "" then
+        for ety_idx = res_idx, 1, -1 do
+            if res[ety_idx] == "" then
+                res[ety_idx] = nil
+            else
+                res_idx = ety_idx
```
thibaultcha: There is no need for this assignment here, we could use `[ety_idx + 1] = nil` directly on the below line. Or, we could to `res_idx = ety_idx`, and group the trailing nil assignment below this `if/else` construct and grouping it for both branches, right before the `return`.

lib/ngx/re.lua
```diff
@@ -236,7 +238,9 @@ function _M.split(subj, regex, opts, ctx, max, res)
 
                 sub_idx = to
 
-                if sub_idx >= len then
+                if sub_idx > len then
+                    break
+                elseif sub_idx == len and last_empty_match then
```
thibaultcha: With the latest patch and the addition of the trailing removing loop, those branches should not be necessary anymore. Applying this patch to your PR runs successfully as well:
```diff
diff --git a/lib/ngx/re.lua b/lib/ngx/re.lua
index 06b5e9b..6e0602f 100644
--- a/lib/ngx/re.lua
+++ b/lib/ngx/re.lua
@@ -201,8 +201,6 @@ function _M.split(subj, regex, opts, ctx, max, res)

                 if sub_idx > len then
                     break
-                elseif sub_idx == len and last_empty_match then
-                    break
                 end
             end
         end
@@ -240,8 +238,6 @@ function _M.split(subj, regex, opts, ctx, max, res)

                 if sub_idx > len then
                     break
-                elseif sub_idx == len and last_empty_match then
-                    break
                 end
             end
         end
```

xiaocang:
Thx, removed useless `if` control branch, and move `res[res_idx + 1] = nil` to the place as your suggested.

thibaultcha:
@agentzh Ping here as well - in my opinion this PR deserves attention as well (a few comments I left there are rendered obsolete since @xiaocang reworked the patch)

agentzh:
@spacewander Will you help review this PR please? Many thanks!

spacewander:
@xiaocang 
Maybe the trailing nil handling could be simpler? I apply this patch and all tests still pass:
```diff
diff --git a/lib/ngx/re.lua b/lib/ngx/re.lua
index e1e2f95..c011ed3 100644
--- a/lib/ngx/re.lua
+++ b/lib/ngx/re.lua
@@ -251,20 +251,18 @@ function _M.split(subj, regex, opts, ctx, max, res)
     -- trailing nil for non-cleared res tables
 
     -- delete empty trailing ones (without max)
-    local remaining_str = sub(subj, sub_idx)
-
-    if max <= 0 and remaining_str == "" then
+    if max <= 0 and sub_idx > len then
         for ety_idx = res_idx, 1, -1 do
-            if res[ety_idx] == "" then
-                res[ety_idx] = nil
-            else
+            if res[ety_idx] ~= "" then
                 res_idx = ety_idx
                 break
             end
+
+            res[ety_idx] = nil
         end
     else
         res_idx = res_idx + 1
-        res[res_idx] = remaining_str
+        res[res_idx] = sub(subj, sub_idx)
     end
 
     res[res_idx + 1] = nil
```

It save a `string.sub` operation.
Please point it out if I have made any mistake.

xiaocang:
I applied the patch and found it's a good optimize 👍

spacewander:
@agentzh 
I think this pr could be merged now.

agentzh:
Merged. Thanks!

Issue #108, opened 2017-05-05
Replace `return error()` with `error()` to avoid tailcalls


agentzh:
Currently this library uses the `return error()` form. This triggers tail call optimization in the Lua VM and makes the resulting backtrace not so useful.

membphis:
can we close this issue now?

thibaultcha:
Considering this resolved indeed! Thanks @spacewander and @membphis :)

PR #173, opened 2018-02-04
improvement: replaced `return error()` with `error()`

spacewander:
For #108 

### DIFF
--- a/lib/resty/core/base64.lua
+++ b/lib/resty/core/base64.lua
@@ -47,7 +47,8 @@ ngx.encode_base64 = function (s, no_padding)
 
     if no_padding then
         if no_padding ~= true then
-            error("boolean argument only")
+            local typ = type(no_padding)
+            error("bad no_padding: boolean expected, got " .. typ, 2)
         end
 
         no_padding_bool = true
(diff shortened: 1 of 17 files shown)


lib/resty/core/base64.lua
```diff
@@ -47,7 +47,7 @@ ngx.encode_base64 = function (s, no_padding)
 
     if no_padding then
         if no_padding ~= true then
-            return error("boolean argument only")
+            error("boolean argument only")
```
thibaultcha: We will need to make use of the second argument to `error` here (and in many other places where it makes sense) so the error is raised at the correct stack level. The previous approach with `return` would tail call and incidentally report the correct stack trace previously.

Here (and in other places where the error needs to appear as originating from the user's code stack level):
```
error("boolean argument only", 2)
```

spacewander:
@thibaultcha 
Fixed in the new commit.
I also change the error message from `"boolean argument only"` to `"bad no_padding: boolean expected, got " .. typ`. The latter one may be more meaningful in the context, since there are two arguments in `encode_base64`.

agentzh:
@spacewander Merged. Thanks!

PR #175, opened 2018-02-05
remove my_require

16m:
Because the function `require()` is not yet implemented by LuaJit. So, I guess the implemented function `my_require()` aims to avoid loading the package that has already been loaded, right?

However, `require()` function itself will check whether the package is already loaded. So is there any need to implement `my_require()` function? If yes, could you explain the reason beneath the surface.

Thank you!

### DIFF
--- a/lib/resty/core/base.lua
+++ b/lib/resty/core/base.lua
@@ -60,24 +60,6 @@ if not ok then
 end
 
 
--- XXX for now LuaJIT 2.1 cannot compile require()
--- so we make the fast code path Lua only in our own
--- wrapper so that most of the require() calls in hot
--- Lua code paths can be JIT compiled.
-do
-    local orig_require = require
-    local pkg_loaded = package.loaded
-    local function my_require(name)
-        local mod = pkg_loaded[name]
-        if mod then
-            return mod
-        end
-        return orig_require(name)
-    end
-    getfenv(0).require = my_require
-end
-
-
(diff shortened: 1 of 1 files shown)


agentzh:
@16m No, it is important. Seems like you didn't even read the comment of the code you are trying to remove:

```
--- XXX for now LuaJIT 2.1 cannot compile require()
 --- so we make the fast code path Lua only in our own
 --- wrapper so that most of the require() calls in hot
 --- Lua code paths can be JIT compiled.
```

It already explains the purpose well.

To be more elaborate, LuaJIT's own require() function is implemented as a CFunction, which always aborts the JIT compilation even when the module in question is already loaded and cached in `package.loaded`. We shouldn't remove our wrapper which can be fully JIT compiled for such fast code paths.

16m:
@agentzh After your explanation, I think the wrapper is designed to avoid **_calling_** the raw `require()` function.
The aim of lua wrapper is that the JIT compilation will abort when "calling" the raw `require()` function, while the lua wrapper will check the `package.loaded` table first, which can be JIT compiled, to avoid loading the cached module.
And more, if I load a not-yet-cached module by the wrapper, the JIT compilation will abort? Is that my understanding right? Thank you!

agentzh:
@16m Yes, that's correct. Hitting the real `require()` should be rare due to the `package.loaded` caching.

PR #170, opened 2018-01-11
optimize: shdict: switched exptime argument type to 'long' to avoid potential overflows.

thibaultcha:
> I hereby granted the copyright of the changes in this pull request
to the authors of this lua-resty-core project.

Sister PR of https://github.com/openresty/lua-nginx-module/pull/1229

PR #178, opened 2018-03-16
change param expire to int64

gcliupeng:
when use the new shdict api in resty/core，the expire may not process correctly，it use type int to story expire but not uint64 as before。when the expire is large，for example 10000000，it goes wrong。I also change the function in the c file


agentzh:
@gcliupeng Travis CI is failing. Better point lua-nginx-module to your own branch for now? Thanks!

agentzh:
@gcliupeng Also, better add a test case to cover this fix? Thanks!

thibaultcha:
Hum, this is already addressed by this PR opened since January (which has a test case and a sister PR on lua-nginx-module as well) https://github.com/openresty/lua-resty-core/pull/170

If we want to switch `long` to `uint64` I could update it there.

thibaultcha:
Addressed by #170 which has been merged for some time already. Thanks.

PR #180, opened 2018-03-26
feature: implemented ngx.resp.add_header.

spacewander:
The `ngx.resp.add_header` works like the `add_header` Nginx directive.
Unlike the `ngx.header.HEADER=`, this method appends new header to the old one
instead of overriding it.
Unlike the `add_header` directive, this method overrides the builtin header
instead of appending it. IMHO, it is reasonable since appending to the
builtin header like `Content-Length` is confusing.

TODO: add documentation.

This pr depends on https://github.com/openresty/meta-lua-nginx-module/pull/32.


lib/resty/core/response.lua
```diff
@@ -52,8 +52,13 @@ local function set_resp_header(tb, key, value)
 
     local rc
     if value == nil then
-        rc = C.ngx_http_lua_ffi_set_resp_header(r, key, #key, true, nil, 0,
-                                                nil, 0, errmsg)
+        if no_override then
+            error("invalid header value", 3)
+        else
```
doujiang24: we can save the `else` here.

agentzh: Agreed.

spacewander:
@agentzh 
Updated.

t/response.t
```diff
@@ -9,7 +9,7 @@ use Cwd qw(cwd);
```
agentzh: We can just use shorter test file names like `resp.t` here? ;)

lib/resty/core/response.lua
```diff
@@ -160,6 +165,11 @@ do
 end
 
 
+function ngx.resp.add_header(key, value)
```
agentzh: Let's introduce an `ngx.resp` Lua module which can be loaded on user's demand.

lib/resty/core/response.lua
```diff
@@ -31,7 +31,7 @@ ffi.cdef[[
     int ngx_http_lua_ffi_set_resp_header(ngx_http_request_t *r,
         const char *key_data, size_t key_len, int is_nil,
         const char *sval, size_t sval_len, ngx_http_lua_ffi_str_t *mvals,
-        size_t mvals_len, char **errmsg);
+        size_t mvals_len, unsigned override, char **errmsg);
```
agentzh: Please use `int` instead of `unsigned` for the `override` parameter.

agentzh:
Also, will you please add some docs for this new API function to the `lib/ngx/resp.md` file and add a link to README? Thanks!

spacewander:
@agentzh 
Have been moved to `lib/ngx/resp`.

PR #183, opened 2018-04-10
feature: added the sig_restart API to the ngx.process module.

thibaultcha:

This is one of 4 PRs to properly handle `EINTR` errors in resty-cli (and/or OpenResty if desired by the user).

- https://github.com/openresty/luajit2/pull/18 Revert the "retry on EINTR" patch
- https://github.com/openresty/lua-nginx-module/pull/1296 Implement an FFI API to enable `SA_RESTART` on signal handlers
- (this PR) Implement the Lua `sig_restart()` API
- https://github.com/openresty/resty-cli/pull/41 Enable the `SA_RESTART` flag on interrupting signals

If all PRs are merged:
- we get rid of the "retry on EINTR" LuaJIT patch which was refused upstream
- resty-cli users can safely use APIs like `io.popen()` https://github.com/openresty/resty-cli/issues/35
- OpenResty users can toggle this behaviour on/off as desired (off by default)

### DIFF
--- a/lib/ngx/process.lua
+++ b/lib/ngx/process.lua
@@ -7,10 +7,16 @@ base.allows_subsystem('http')
 
 local ffi = require 'ffi'
 local errmsg = base.get_errmsg_ptr()
+local get_string_buf = base.get_string_buf
+local get_size_ptr = base.get_size_ptr
 local FFI_ERROR = base.FFI_ERROR
 local ffi_str = ffi.string
 local ngx_phase = ngx.get_phase
 local tonumber = tonumber
+local type = type
+
+
+local ERR_BUF_SIZE = 128
 
 
 local process_type_names = {
@@ -27,11 +33,25 @@ local C = ffi.C
 local _M = { version = base.version }
 
 
+_M.signums = {
+    SIGHUP   = 1,
+    SIGINT   = 2,
+    SIGQUIT  = 3,
+    SIGTRAP  = 5,
+    SIGABRT  = 6,
+    SIGKILL  = 9,
+    SIGALARM = 14,
+    SIGTERM  = 15,
+}
+
+
 ffi.cdef[[
 int ngx_http_lua_ffi_enable_privileged_agent(char **err);
 int ngx_http_lua_ffi_get_process_type(void);
 void ngx_http_lua_ffi_process_signal_graceful_exit(void);
 int ngx_http_lua_ffi_master_pid(void);
+int ngx_http_lua_ffi_set_sa_restart(int signum, int enabled,
+    unsigned char *err, size_t *errlen);
 ]]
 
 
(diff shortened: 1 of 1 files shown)


thibaultcha:
This PR is clearly missing **tests** and **documentation**, which I can add if desired.

lib/ngx/process.lua
```diff
@@ -32,6 +34,8 @@ int ngx_http_lua_ffi_enable_privileged_agent(char **err);
 int ngx_http_lua_ffi_get_process_type(void);
 void ngx_http_lua_ffi_process_signal_graceful_exit(void);
 int ngx_http_lua_ffi_master_pid(void);
+int ngx_http_lua_ffi_enable_sa_restart(int signum, int enabled);
+char *strerror(int errnum);
```
thibaultcha: imho, this _might_ belong to another `cdef` definition, if so please let me know where.

thibaultcha:
@dndx @agentzh Hey there, any feedback on this side of the signal restart? I believe the ngx_lua part is reviewed, but the Lua-land and resty-cli aren't. This is a nice non-intrusive way to solve those resty-cli issues imho. Let me know if you get some time to look at it :)

lib/ngx/process.lua
agentzh: This seems to break encapsulation. Better make the `ngx_http_lua_ffi_enable_sa_restart()` API function itself return the error string, just like our other C API functions?

thibaultcha: No problem, will change it :+1:

thibaultcha: Updated. Is it looking good enough for tests and documentation?

dndx:
@thibaultcha This PR LGTM. Although I am thinking if we could add some constants for common POSIX signals to make using this easier? Providing numbers directly is not always convenient.

See: http://man7.org/linux/man-pages/man7/signal.7.html section "Standard signals".

thibaultcha:
@dndx I thought about doing so but I wasn't sure where to put them (somewhere on the `ngx` global probably, but where?), and as such I wasn't sure it would be accepted. Do you have any suggestion?

dndx:
@thibaultcha What about `ngx.process.SIGPIPE`? I think putting it as module level constant for `process.lua` is fine for now. @agentzh Any inputs on this as well?

thibaultcha:
@dndx If we hard-code them in the Lua-land, my only concern is portability. If we do so, should we only provide constants for those signals that are defined as the same across all architectures?

dndx:
@thibaultcha Yes. I think we can just provide the ones in https://en.wikipedia.org/wiki/Signal_(IPC) that has "Portable number" assigned from POSIX and leave the rest to the user.

thibaultcha:
@dndx Works for me.

thibaultcha:
@dndx Added

thibaultcha:
@agentzh @dndx Ping on this and its sister PRs

thibaultcha:
Closing this PR as it seems the `SA_RESTART` flag will get an nginx directive, but not an FFI API (as per the review in openresty/lua-nginx-module#1296).

PR #184, opened 2018-04-17
feature: ffi-based ngx.req.is_internal

chronolaw:
In travis.ci，I got this error message:
```
t/misc.t                         (Wstat: 65280 Tests: 64 Failed: 0)
  Non-zero exit status: 255
  Parse errors: Bad plan.  You planned 80 tests but ran 64.
```
I think it may be raised by this:
```
plan tests => repeat_each() * (blocks() * 6 - 2);
```

It looks like a magic number to me, could you explain it or fix it for me? Many thanks.

### DIFF
--- a/t/misc.t
+++ b/t/misc.t
@@ -10,7 +10,7 @@ log_level('warn');
 #repeat_each(120);
 repeat_each(2);
 
-plan tests => repeat_each() * (blocks() * 5 - 3);
+plan tests => repeat_each() * (blocks() * 5 - 6);
 
 my $pwd = cwd();
 
@@ -221,7 +221,7 @@ internal
 
             local ctx = package.loaded.bad_context
             ngx.say(ctx[1])
-            local i = string.find(ctx[2], "no request found")
+            local i = string.find(ctx[2], "API disabled in the current context")
             ngx.say(i > 1)
         }
     }
@@ -230,4 +230,3 @@ GET /lua
 --- response_body
 false
 true
---- SKIP


spacewander:
@chronolaw 
`repeat_each()` is defined in the beginning of each test file.
`blocks()` will return the number of test blocks in current test file.
Just tune the expression to make its result equal to the actual number of run tests.

t/misc.t
```diff
@@ -159,3 +159,45 @@ GET /t
...
+                ngx.say("internal")
+            else
+                ngx.say("not internal")
+            end
+        ';
+    }
+--- request
+GET /test
+--- response_body
+internal
```
spacewander: Could you add another test case for FFI_BAD_CONTEXT branch?

chronolaw:
traivs.ci build has passed，please see https://travis-ci.org/chronolaw/lua-resty-core.svg?branch=master

Please note : this is build with my own `ngx_lua_module` branch(` with ffi_req_is_internal`).

t/misc.t
chronolaw: It seems no `FFI_BAD_CONTEXT` test case for `is_subreq` and `headers_sent`, are there any example?

spacewander: @chronolaw 
You could run `is_internal` in phase doesn't support it, and use `pcall` to capture the result. The result could be returned as the response for `GET /xxx`.

chronolaw: It works in all phases except `init_by_lua` and 'init_worker_by_lua' and 'ngx.timer'，should I use `ngx.errorlog` to capture errorlog?

And I thin it will generate ''no request found" but not "API disabled in the current context".

spacewander: > should I use ngx.errorlog to capture errorlog

You could store the result in `package.loaded` or just a variable which could be accessed by the `GET /xxx` request and the `ngx.timer`.

chronolaw: It use the `error` function like below
```
error("no request found")
```
This will case a stack traceback in errlog, we can not get any return value.

spacewander: @chronolaw 
Just forget the errlog and use `pcall` to capture the `error` message.

spacewander:
@chronolaw 
I checked out your code and run the test.
You are very closed to the answer.

Tips 1: you are testing `bad context` instead of `no request found`.
Tips 2: if you don't know why thing doesn't work, log down every variable.

chronolaw:
OK， I will try , it may take some time, thanks.

chronolaw:
All test case are passed, please see https://travis-ci.org/chronolaw/lua-resty-core/jobs/367575302.

I also commited the lastest code for PR.

PR #186, opened 2018-04-19
doc: fixed the return value spec for the ssl.raw_client_addr

chronolaw:
And I think the doc should add version info like below:

```
This function was first added in version 0.1.14.
```

Now there is no `ssl.raw_client_addr` in `1.13.6.1`.


agentzh:
@chronolaw Merged. Thanks. Also marked the first version of lua-resty-core that this feature first appeared.

PR #188, opened 2018-04-23
change: narrowed the ngx_lua_version requirement.

spacewander:
If we change "char **errmsg" to "unsigned char *errbuf" in the FFI C
function, then the old lua-resty-core won't work anymore. Howerver, we only check
if the lua-nginx-module is older than required, so in this case an old
lua-resty-core could pass the verification.

As we don't ensure the backward compatibility of FFI C functions,
it is better to specify the ngx_lua_version as a number instead of a range.


spacewander:
Related issue: https://github.com/openresty/lua-nginx-module/issues/1308

agentzh:
@spacewander Merged with edits in your commit log message. Thanks!

PR #189, opened 2018-04-26
doc: add missing API:get_tls1_version_str

chronolaw:


agentzh:
@chronolaw Merged your patch manually.

BTW, please do not use `git merge` to sync branches since it would lead to non-linear git commit history. Better always use `git rebase` in your own private branches. A linear history would make my life much easier when rebasing and squashing your commits right before merging. Thanks!

PR #193, opened 2018-05-14
bugfix: error out when a shdict key is not a string

jdesgats:
This is to stick with the behavior of the regular C API that uses
`luaL_checklstring` for keys, that throws a runtime error when a non-string
key is passed. For some reason, a soft `nil, error` is returned
specifically for `nil` keys, to that bihavor is kept here too.

Note: the error message between the C and the FFI representation is slightly different (`bad argument #1 to 'set' (string expected, got table)` for the C variant, `./resty/core/shdict.lua:105: bad key (string expected, got table)` for the FFI one) but I think it is quite minor.

FTR: it led to a very hard to debug issue where we were wrongly using a table as key in some cases. The string representation of it (e.g. `table: 0x400b63e0`) was sometime triggering collisions where a new table happened to be allocated at the same address as a previous one.

### DIFF
--- a/lib/resty/core/shdict.lua
+++ b/lib/resty/core/shdict.lua
@@ -11,7 +11,6 @@ local get_string_buf = base.get_string_buf
 local get_string_buf_size = base.get_string_buf_size
 local get_size_ptr = base.get_size_ptr
 local tonumber = tonumber
-local tostring = tostring
 local next = next
 local type = type
 local error = error
@@ -102,7 +101,7 @@ local function shdict_store(zone, op, key, value, exptime, flags)
     end
 
     if type(key) ~= "string" then
-        key = tostring(key)
+        error("bad key (string expected, got " .. type(key) .. ")")
     end
 
     local key_len = #key
@@ -196,7 +195,7 @@ local function shdict_get(zone, key)
     end
 
     if type(key) ~= "string" then
-        key = tostring(key)
+        error("bad key (string expected, got " .. type(key) .. ")")
     end
 
     local key_len = #key
@@ -271,7 +270,7 @@ local function shdict_get_stale(zone, key)
     end
 
     if type(key) ~= "string" then
-        key = tostring(key)
+        error("bad key (string expected, got " .. type(key) .. ")")
     end
 
     local key_len = #key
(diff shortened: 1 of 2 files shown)


dndx (dismissed):
LGTM. I think we should try keep `lua-resty-core` version of the API behavior consistent with the C API version if they does exist.

lib/resty/core/shdict.lua
```diff
@@ -102,7 +101,7 @@ local function shdict_store(zone, op, key, value, exptime, flags)
     end
 
     if type(key) ~= "string" then
-        key = tostring(key)
+        error("bad key (string expected, got " .. type(key) .. ")")
```
thibaultcha: If we are to stay consistent with the ngx_lua behavior (and that of this library), we need to add the `lvl` argument to those `error()` calls, to report the proper stack level:
```lua
error("bad key (string expected, got " .. type(key) .. ")", 2)
```

dndx: Ah I see, that's a good catch. @jdesgats Could you please update this PR to add the `lvl` argument?

jdesgats: Good catch, thanks for the review! Fixed now.

t/shdict.t
```diff
@@ -1636,3 +1636,63 @@ ttl: 2147483648
...
+set false content_by_lua.*: bad key \(string expected, got table\)
+safe_set false content_by_lua.*: bad key \(string expected, got table\)
+get false content_by_lua.*: bad key \(string expected, got function\)
+get_stale false content_by_lua.*: bad key \(string expected, got function\)
+add false content_by_lua.*: bad key \(string expected, got number\)
+safe_add false content_by_lua.*: bad key \(string expected, got number\)
+replace false content_by_lua.*: bad key \(string expected, got table\)
+delete false content_by_lua.*: bad key \(string expected, got table\)
+incr false content_by_lua.*: bad key \(string expected, got boolean\)
+ttl false content_by_lua.*: bad key \(string expected, got table\)
```
thibaultcha: I believe we are missing a test case for [shm:expire()](https://github.com/openresty/lua-nginx-module#ngxshareddictexpire)? It is part of the Lua helpers modified in this patch.

PR #206, opened 2018-09-18
feature: added support for ARM64.

spacewander:
now we require OpenResty's LuaJIT branch to work properly on ARM64 since
we make use of the thread.exdata Lua API.

also implemented the ndk.* API with FFI.

Thanks Dejiang Zhu and Zexuan Luo for the development work of this patch.

Signed-off-by: Yichun Zhang (agentzh) <agentzh@gmail.com>

### DIFF
--- a/lib/resty/core/base.lua
+++ b/lib/resty/core/base.lua
@@ -129,7 +129,7 @@ end
 local c_buf_type = ffi.typeof("char[?]")
 
 
-local _M = new_tab(0, 17)
+local _M = new_tab(0, 19)
 
 
 _M.version = "0.1.15"
@@ -230,6 +230,25 @@ _M.FFI_ERROR = -1
 _M.FFI_BUSY = -3
 _M.FFI_DONE = -4
 _M.FFI_DECLINED = -5
+
+
+local exdata
+ok, exdata = pcall(require, "thread.exdata")
+if ok and exdata then
+    function _M.get_request()
+        local r = exdata()
+        if r ~= nil then
+            return r
+        end
+    end
+
+else
+    local getfenv = getfenv
+
+    function _M.get_request()
+        return getfenv(0).__ngx_req
+    end
+end
 
 
 return _M
--- a/lib/resty/core/base.lua
+++ b/lib/resty/core/base.lua
@@ -129,7 +129,7 @@ end
 local c_buf_type = ffi.typeof("char[?]")
 
 
-local _M = new_tab(0, 19)
+local _M = new_tab(0, 18)
 
 
 _M.version = "0.1.15"
@@ -232,21 +232,24 @@ _M.FFI_DONE = -4
 _M.FFI_DECLINED = -5
 
 
-local exdata
-ok, exdata = pcall(require, "thread.exdata")
-if ok and exdata then
-    function _M.get_request()
-        local r = exdata()
-        if r ~= nil then
-            return r
+do
+    local exdata
+
+    ok, exdata = pcall(require, "thread.exdata")
+    if ok and exdata then
+        function _M.get_request()
+            local r = exdata()
+            if r ~= nil then
+                return r
+            end
         end
-    end
 
-else
-    local getfenv = getfenv
+    else
+        local getfenv = getfenv
 
-    function _M.get_request()
-        return getfenv(0).__ngx_req
+        function _M.get_request()
+            return getfenv(0).__ngx_req
+        end
     end
 end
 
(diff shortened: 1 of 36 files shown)


javierguerragiraldez:
we've tested this at Cloudflare, seems stable enough.

lib/resty/core/base.lua
```diff
@@ -129,7 +129,7 @@ end
 local c_buf_type = ffi.typeof("char[?]")
 
 
-local _M = new_tab(0, 17)
+local _M = new_tab(0, 19)
```
thibaultcha: It seems like `18` is the appropriate size here (`_M.get_request` can only be assigned once). This also appears to be proven by [this test](https://github.com/openresty/lua-resty-core/pull/206/files#diff-cf27e9ecb184dab30525a4fbeb4b119bR69).

lib/resty/core/base.lua
```diff
@@ -232,4 +232,23 @@ _M.FFI_DONE = -4
 _M.FFI_DECLINED = -5
 
 
+local exdata
```
thibaultcha: Might be worth encapsulating `exdata` in a `do ... end` block here, to prevent future contributors to this file from calling/tampering with it in subsequently implemented functions. As in:
```lua
do
    local exdata

    local pok, exdata = pcall(...)
    if pok and type(exdata) == "function" then
        function _M.get_request()
            local r = exdata() -- only allowed caller of this function, which must not be tampered with
            -- ...
        end

    else
        -- ...
    end
end
```

lib/resty/core/ndk.lua
```diff
@@ -0,0 +1,91 @@
...
+local error = error
+
+
+local _M = {
+    version = base.version
+}
+
+
+ffi.cdef[[
+typedef unsigned char u_char;
```
thibaultcha: Other modules in this repository make a point not to alias `unsigned char` to `u_char`. E.g. https://github.com/openresty/lua-resty-core/blob/1d5c898e3574d7c38ef2d8d905e8481f9cfd5aaa/lib/resty/core/hash.lua#L16-L17

Not sure if we should start doing so here.

lib/resty/core/regex.lua
```diff
@@ -343,15 +448,15 @@ local function re_match_compile(regex, opts)
         -- print("compiled regex not found, compiling regex...")
         local errbuf = get_string_buf(MAX_ERR_MSG_LEN)
 
-        compiled = C.ngx_http_lua_ffi_compile_regex(regex, #regex,
-                                                    flags, pcre_opts,
-                                                    errbuf, MAX_ERR_MSG_LEN)
+        compiled = ngx_lua_ffi_compile_regex(regex, #regex, flags,
+                                                  pcre_opts, errbuf,
```
thibaultcha: style: the arguments indentation is off here.

t/shdict.t
```diff
@@ -399,7 +399,7 @@ GET /t
 value: 371
 err: nil
 --- error_log eval
-qr/\[TRACE\s+\d+ content_by_lua\(nginx\.conf:\d+\):7 loop\]/
+qr/\[TRACE\s+\d+ content_by_lua\(nginx\.conf:\d+\):7 loop]/
```
agentzh: Why this change? It's better to always escape `]` in regexes since it is a metachar per se though unpaired ones are usually harmless.

agentzh:
We should bump the minimum stream lua module version in this PR:

```diff
diff --git a/lib/resty/core/base.lua b/lib/resty/core/base.lua
index 96412a9..61704fb 100644
--- a/lib/resty/core/base.lua
+++ b/lib/resty/core/base.lua
@@ -26,14 +26,14 @@ if subsystem == 'http' then
 elseif subsystem == 'stream' then
     if not ngx.config
        or not ngx.config.ngx_lua_version
-       or ngx.config.ngx_lua_version ~= 5
+       or ngx.config.ngx_lua_version ~= 6
     then
-        error("ngx_stream_lua_module 0.0.5 required")
+        error("ngx_stream_lua_module 0.0.6 required")
     end

 else
     error("ngx_http_lua_module 0.10.14 or "
-          .. "ngx_stream_lua_module 0.0.5 required")
+          .. "ngx_stream_lua_module 0.0.6 required")
 end

```

lib/resty/core/ndk.lua
```diff
@@ -0,0 +1,90 @@
+-- Copyright (C) by OpenResty Inc.
+
+
+local ffi = require 'ffi'
+local base = require "resty.core.base"
+
```
agentzh: We should only allow loading this module in the http subsystem via the call `base.allows_subsystem('http')`.

spacewander:
Already merged.

PR #209, opened 2018-10-06
tests: shdict: try harder to JIT in stream tests.

thibaultcha:

I have noticed that the stream shdict tests around JIT support were somewhat flaky:

* https://travis-ci.org/openresty/lua-resty-core/jobs/438078376
* https://travis-ci.org/openresty/lua-resty-core/jobs/438079493
* https://travis-ci.org/thibaultcha/lua-resty-core/jobs/438086301

After some a few more tests on Travis, it seems like 250 iterations yields more reliable results.


thibaultcha:
@agentzh What do you think of this approach to increase the chances of JIT happening? I could not find precedents in this repository. Should we tweak the JIT options instead?

agentzh:
@thibaultcha We can surely add `jit.opt.start("hotloop=10")`, for example (or whatever number is a good fit).

agentzh:
@thibaultcha Now that you use 20 (I think 10 should be enough), you can reduce the total `for` loop iteration count in all those test cases down to `n + 2` to make them run faster?

thibaultcha:
Sure. I can also reduce the hotloop trigger to `10` as well.

agentzh:
@thibaultcha Great.

thibaultcha:
`hotloop=10` and `12` iterations still produced some unexpected failures: https://travis-ci.org/openresty/lua-resty-core/builds/448556066

Now giving it a try with `20` iterations.

agentzh:
@thibaultcha OK. Just use whatever small enough numbers which work :)

PR #210, opened 2018-10-18
feature: ssl_certby_lua* support for stream_lua.

dndx:

### DIFF
--- a/lib/ngx/ssl.lua
+++ b/lib/ngx/ssl.lua
@@ -2,7 +2,7 @@
 
 
 local base = require "resty.core.base"
-base.allows_subsystem('http')
+base.allows_subsystem('http', 'stream')
 
 
 local ffi = require "ffi"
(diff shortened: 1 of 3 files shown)


agentzh:
@thibaultcha Will you have a look at this? If everything is OK, please merge it :) Thanks!

lib/ngx/ssl.lua
```diff
@@ -17,53 +17,149 @@ local get_string_buf = base.get_string_buf
...
+    ngx_lua_ffi_cert_pem_to_der = C.ngx_stream_lua_ffi_cert_pem_to_der
+    ngx_lua_ffi_priv_key_pem_to_der = C.ngx_stream_lua_ffi_priv_key_pem_to_der
+    ngx_lua_ffi_ssl_get_tls1_version = C.ngx_stream_lua_ffi_ssl_get_tls1_version
+    ngx_lua_ffi_parse_pem_cert = C.ngx_stream_lua_ffi_parse_pem_cert
+    ngx_lua_ffi_parse_pem_priv_key = C.ngx_stream_lua_ffi_parse_pem_priv_key
+    ngx_lua_ffi_set_cert = C.ngx_stream_lua_ffi_set_cert
+    ngx_lua_ffi_set_priv_key = C.ngx_stream_lua_ffi_set_priv_key
+    ngx_lua_ffi_free_cert = C.ngx_stream_lua_ffi_free_cert
+    ngx_lua_ffi_free_priv_key = C.ngx_stream_lua_ffi_free_priv_key
+end
```
thibaultcha: style: there used to be 2 blank lines below this, and now a single one only

lib/ngx/ssl.lua
```diff
@@ -93,7 +189,7 @@ function _M.set_der_cert(data)
         error("no request found")
     end
 
-    local rc = C.ngx_http_lua_ffi_ssl_set_der_certificate(r, data, #data,
+    local rc = ngx_lua_ffi_ssl_set_der_certificate(r, data, #data,
```
thibaultcha: style: the alignment of the arguments is off. Also, maybe now they could all fit on a single line? There are many places in this PR where this is the case as well

t/stream/ssl.t
```diff
@@ -0,0 +1,1826 @@
+# vim:set ft= ts=4 sw=4 et fdm=marker:
+
+use Test::Nginx::Socket::Lua::Stream;use Cwd qw(cwd);
```
thibaultcha: Is it intentional that there isn't a new line between those two statements? The `t/ssl.t` test suite has such a new line.

t/stream/ssl.t
```diff
@@ -0,0 +1,1826 @@
...
+
+our $CWD = cwd();
+
+no_long_string();
+#no_diff();
+
+env_to_nginx("PATH=" . $ENV{'PATH'});
+$ENV{TEST_NGINX_LUA_PACKAGE_PATH} = "$::CWD/lib/?.lua;;";
+$ENV{TEST_NGINX_HTML_DIR} ||= html_dir();
+$ENV{TEST_NGINX_SERVER_SSL_PORT} ||= 4443;
```
thibaultcha: If we run the test suite concurrently, there is a chance that this suite conflicts with `t/ssl.t` (which uses the same env variable and defaults to the same port). Another approach would be to use a unix socket maybe?
```
listen unix:$ENV{TEST_NGINX_HTML_DIR}/nginx-ssl.sock ssl;
```

t/stream/ssl.t
```diff
@@ -0,0 +1,1826 @@
...
+run_tests();
+
+__DATA__
+
+=== TEST 1: clear certs
+--- stream_config
+    lua_package_path "$TEST_NGINX_LUA_PACKAGE_PATH";
+
+    server {
+        listen unix:$TEST_NGINX_HTML_DIR/nginx.sock ssl;
```
thibaultcha: Oh, this is my previous point. Maybe we should update all `listen` directives for mock upstreams to use this.

t/stream/ssl.t
dndx: No that was definitely not intentional. Fixed.

lib/ngx/ssl.lua
dndx: Fixed.

t/stream/ssl.t
dndx: That's a good idea, although does not really belongs to this PR. Maybe we can have a separate issue for it.

agentzh: I agree.

agentzh: @dndx It is a new issue introduced by this PR, so it's better to be addressed in this PR too? AFAIK, the current master does support running tests in parallel.

t/stream/ssl.t
```diff
@@ -19,7 +19,7 @@ no_long_string();
 env_to_nginx("PATH=" . $ENV{'PATH'});
 $ENV{TEST_NGINX_LUA_PACKAGE_PATH} = "$::CWD/lib/?.lua;;";
 $ENV{TEST_NGINX_HTML_DIR} ||= html_dir();
-$ENV{TEST_NGINX_SERVER_SSL_PORT} ||= 4443;
+$ENV{TEST_NGINX_SERVER_SSL_PORT} ||= 4433;
```
thibaultcha: That works too!

PR #212, opened 2018-10-24
bugfix: ngx.resp: resp.add_header should reject empty table header va…

spacewander:
…lue.

Pass an empty table as header value is meaningless as no header will be
added.

### DIFF
--- a/lib/resty/core/response.lua
+++ b/lib/resty/core/response.lua
@@ -69,7 +69,7 @@ local function set_resp_header(tb, key, value, no_override)
         if type(value) == "table" then
             mvals_len = #value
             if mvals_len == 0 and no_override then
-                error("invalid header value", 3)
+                return
             end
 
             buf = get_string_buf(ffi_str_size * mvals_len)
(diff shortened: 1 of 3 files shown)


lib/resty/core/response.lua
```diff
@@ -68,6 +68,10 @@ local function set_resp_header(tb, key, value, no_override)
 
         if type(value) == "table" then
             mvals_len = #value
+            if mvals_len == 0 and no_override then
+                error("invalid header value", 3)
```
thibaultcha: The API would be a lot more user-friendly if we _allowed_ such empty tables instead. E.g. a user could accumulate a list of headers to add in a table, and pass it to `ngx_resp.add_header()` later on. With this change, the user would not have to do:
```lua
if #my_headers > 0 then
    ngx_resp.add_header("Foo", my_headers)
end
``` 

Which feels very burdensome for users.

Instead, I suggest that we update this patch to early-exit when the `value` table argument has a length of 0. And we keep the test around, except that the response simply contains `Foo: aaa`.

Thoughts?

thibaultcha:
Merged.

thibaultcha:
I will do many more merges on Monday and next week

PR #213, opened 2018-10-30
feature: implemented ngx.req module and its function add_header().

spacewander:
The `ngx.req.add_header` works mostly like `ngx.req.set_header`.
However, unlike `ngx.req.set_header`, this method appends new header to the old
one instead of overriding it if the header is not builtin.

For the builtin header, this method overrides it instead of appending to it.
IMHO, it is reasonable since appending to the builtin header like `User-Agent`
is confusing and semantically wrong.

### DIFF
--- a/lib/resty/core/request.lua
+++ b/lib/resty/core/request.lua
@@ -27,6 +27,15 @@ local tostring = tostring
 local tonumber = tonumber
 
 
+local _M = {
+    version = base.version
+}
+
+
+local ffi_str_type = ffi.typeof("ngx_http_lua_ffi_str_t*")
+local ffi_str_size = ffi.sizeof("ngx_http_lua_ffi_str_t")
+
+
 ffi.cdef[[
     typedef struct {
         ngx_http_lua_ffi_str_t   key;
@@ -56,9 +65,10 @@ ffi.cdef[[
 
     int ngx_http_lua_ffi_req_set_method(ngx_http_request_t *r, int method);
 
-    int ngx_http_lua_ffi_req_header_set_single_value(ngx_http_request_t *r,
+    int ngx_http_lua_ffi_req_set_header(ngx_http_request_t *r,
         const unsigned char *key, size_t key_len, const unsigned char *value,
-        size_t value_len);
+        size_t value_len, ngx_http_lua_ffi_str_t *mvals, size_t mvals_len,
+        int override);
 ]]
 
 
@@ -292,13 +302,7 @@ end
 
 
 do
-    local orig_func = ngx.req.set_header
-
-    function ngx.req.set_header(name, value)
-        if type(value) == "table" then
-            return orig_func(name, value)
-        end
-
+    local function set_req_header(name, value, override)
         local r = get_request()
         if not r then
             error("no request found")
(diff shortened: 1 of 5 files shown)


spacewander:
CC @agentzh @thibaultcha

lib/resty/core/request.lua
```diff
@@ -310,16 +314,53 @@ do
 
         local rc
         if not value then
-            rc = C.ngx_http_lua_ffi_req_header_set_single_value(r, name,
-                                                         #name, nil, 0)
+            if not override then
+                error("invalid header value", 3)
```
thibaultcha: I think we should be more descriptive here. We usually are in our other APIs... How about: `"bad value arg: string or table expected, got nil"`. We should also update the `if not value` condition above to better handle truthy/falsy values. So maybe change it to `if value == nil`? The condition in `ngx_resp.add_header()` does this already.

lib/resty/core/request.lua
```diff
@@ -310,16 +314,53 @@ do
...
 
         else
-            if type(value) ~= "string" then
-                value = tostring(value)
+            local sval, sval_len, mvals, mvals_len, buf
+
+            if type(value) == "table" then
+                mvals_len = #value
+                if mvals_len == 0 and not override then
+                    error("invalid header value", 3)
```
thibaultcha: Just like the `ngx_resp.add_header()` API, we should early-exit here, but not consider the argument invalid.

lib/resty/core/request.lua
```diff
@@ -332,6 +373,19 @@ do
...
+    _M.set_req_header = set_req_header
+
+    local orig_func = ngx.req.set_header
+
+    function ngx.req.set_header(name, value)
+        if type(value) == "table" then
+            return orig_func(name, value)
+        end
+
+        local override = true
```
thibaultcha: Better to avoid using a variable here and directly pass the argument. A comment can explain what it is if desired.

t/ngx-req.t
```diff
@@ -0,0 +1,175 @@
+use lib 'lib';
+use Test::Nginx::Socket::Lua;
+use Cwd qw(cwd);
```
thibaultcha: Please use the `t::TestCore` module, just like the `ngx-resp.t` suite.

t/ngx-req.t
```diff
@@ -0,0 +1,175 @@
...
+
+    if (!defined $block->error_log) {
+        $block->set_value("no_error_log", "[error]");
+    }
+
+    if (!defined $block->request) {
+        $block->set_value("request", "GET /t");
+    }
+
+});
```
thibaultcha: Using the `t::TestCore` module will allow to get rid of most of the code duplication here.

t/ngx-req.t
```diff
@@ -0,0 +1,175 @@
...
+run_tests();
+
+__DATA__
+
+=== TEST 1: ngx.req.add_header (jitted)
+--- config
+    location = /t {
+        content_by_lua_block {
+            local ngx_req = require "ngx.req"
+            for i = 1, 500 do
```
thibaultcha: Please use the `TEST_NGINX_HOTLOOP` variable when trying to trigger a JIT compilation in tests. E.g. as in `re-find.t`: `for i = 1, $TEST_NGINX_HOTLOOP * 10 do`.

t/ngx-req.t
```diff
@@ -0,0 +1,175 @@
...
+
+            check_invalid_header_value()
+            check_invalid_header_value(nil)
+            check_invalid_header_value({})
+        }
+    }
+--- response_body
+invalid header value
+invalid header value
+invalid header value
```
thibaultcha: Let's also add test cases for boolean values, so we cover our ground with regards to the `if value == nil` condition.

thibaultcha:
Merged!

PR #215, opened 2018-10-31
feature: implemented the ngx.pipe API for nonblocking process spawnin…

spacewander:
…g and IPC through stdin/stdout/stderr.

Now we could spawn a process and interact with its stdin/stdout/stderr
without blocking any OS threads.

Signed-off-by: Yichun Zhang (agentzh) <agentzh@gmail.com>


spacewander:
Relative pull requests:
1. https://github.com/openresty/lua-nginx-module/pull/1407
2. https://github.com/openresty/openresty/pull/412

README.markdown
```diff
@@ -286,6 +287,14 @@ See the [documentation](./lib/ngx/resp.md) for this Lua module for more details.
 
 [Back to TOC](#table-of-contents)
 
+## ngx.pipe
+
+This Lua module provides Lua API which could be used to spawn processes and communicate with them nonblockingly.
```
thibaultcha: Maybe:

This module provides a Lua API to spawn processes and communicate with them in a non-blocking fashion.

lib/ngx/pipe.lua
```diff
@@ -0,0 +1,580 @@
...
+
+
+local function hasattr(mod, attr)
+    return mod[attr]
+end
+
+
+if not pcall(hasattr, C, "ngx_http_lua_pipe_spawn") then
+    error("pipe API is unsupported due to platform issue " ..
+          "or without HAVE_SOCKET_CLOEXEC_PATCH patch")
```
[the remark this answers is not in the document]
spacewander: @thibaultcha 
Or maybe we could add the signal handler and the `ngx_http_lua_ffi_pipe_spawn` function, only when both the platforms support it and the patch is applied?

lib/ngx/pipe.md
```diff
@@ -0,0 +1,531 @@
...
+**syntax:** *data, err, partial = proc:stderr_read_bytes(len)*
+
+**context:** *phases support yielding*
+
+Reads data from stderr like [stderr_read_all](#stderr_read_all) but only reads the
+specified number of bytes of data.
+
+If the data stream is truncated with less bytes of data available, it returns
+3 values: `nil`, the error string `"closed"`, and the partial data string received so far.
+
```
agentzh: Should we put the docs of the `stderr_read_xxx` before `stdout_read_xxx`? We should put the real docs first, I think?

lib/ngx/pipe.lua
```diff
@@ -0,0 +1,580 @@
...
+            if opts.buffer_size then
+                buffer_size = tonumber(opts.buffer_size)
+
+                if not buffer_size or buffer_size < 1 then
+                    error("bad buffer_size option", 2)
+                end
+            end
+        end
+
+        local proc = Proc()
```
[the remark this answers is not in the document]
spacewander: Yes, it is the `ffi.meta.__call`.

spacewander:
@thibaultcha 
I have updated some of the code, for the rest I am waiting for your response.

thibaultcha:
@spacewander I will make an additional commit here and in the sister PR with my suggestions or fixes to speed up the process. I haven't had time to finish it today but I hope to do so tomorrow!

t/pipe.t
```diff
@@ -0,0 +1,1439 @@
...
+=== TEST 26: interrupt signals should not break ngx.pipe IO
+github issue openresty/resty-cli#35
+--- config
+    location = /t {
+        content_by_lua_block {
+            local ngx_pipe = require "ngx.pipe"
+            local filename = "$TEST_NGINX_HTML_DIR/testfile"
+            local f = io.open(filename, 'w')
+            f:write("testfile")
+            f:close()
```
thibaultcha: Is the intent here to reproduce the behavior observed in the linked GitHub issue via this `io.open` operation? Or something else?

t/pipe.t
```diff
@@ -0,0 +1,1439 @@
+# vim:set ft= ts=4 sw=4 et fdm=marker:
+use Test::Nginx::Socket::Lua;
+use Cwd qw(abs_path realpath cwd);
+use File::Basename;
+
+repeat_each(2);
+
+plan tests => repeat_each() * (blocks() * 3 + 16);
```
thibaultcha: On my local setup, only 276 tests are executed (vs 278 on Travis CI and according to this test plan). Yet I am using the master branch of Test::Nginx.

t/pipe.t
spacewander: The `io.open` operation writes the `cpanfile` in that issue. The content in the file doesn't matter.

lib/ngx/pipe.lua
```diff
@@ -0,0 +1,591 @@
...
+        error("no request found")
+    end
+
+    local data_type = type(data)
+    if data_type ~= "string" then
+        if data_type == "table" then
+            data = table_concat(data, '')
+
+        else
+            data = tostring(data)
```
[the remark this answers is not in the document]
spacewander: IMHO, in practice, users aren't aware of the argument is `42` or `"42"`.

lib/ngx/pipe.lua
```diff
@@ -583,7 +576,8 @@ do
 
         return proc
     end
-end  -- do
+end -- do
```
spacewander: Err, I think there should be two spaces after the `end` keyword?

t/pipe.t
spacewander: It is strange...
In my local machine, 278 tests are run, like one in the Travis CI.

lib/ngx/pipe.lua
thibaultcha: Indeed, my bad.

lib/ngx/pipe.lua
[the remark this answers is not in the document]
spacewander: @thibaultcha 
What about adding checks for `nil` type and `userdata` type?

spacewander:
@thibaultcha 
Updated.

lib/ngx/pipe.lua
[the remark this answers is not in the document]
thibaultcha: @spacewander Same for `function` and `boolean` as well... If we want to infer numbers, then we should only `tostring` numbers, and ensure that the input is a `string` later down. Everything else should throw an error.

```lua
       if data_type == "table" then
            data = table_concat(data, '')

        elseif data_type == "number" then
            data = tostring(data)
        
        else
            error("bad data arg: string, number, or table expected, got " .. data_type, 2)
        end
``` 

That said, keep in mind that such values will make `tostring()` produce `"inf"` and other strange inputs for `stdin`:
```
tostring(math.huge) -- inf
tostring(1/0) -- inf
tostring(-1/0) -- -inf
tostring(0/0) -- nan
```

If the user stores the result of some computation in a variable, and passes this variable as an argument to this function, we never now what bug we could introduce by being lenient with our input... Imho, it isn't a good idea to infer arguments, but if you want to, I am fine with the behavior described above (for `table` and `number` only).

t/pipe.t
```diff
@@ -934,8 +934,11 @@ shutdown: merged to stdout
 
 
 
-=== TEST 26: interrupt signals should not break ngx.pipe IO
+=== TEST 26: interrupt signals which break io.popen should not break ngx.pipe IO
```
thibaultcha: Nothing will produce `EINTR` since we don't use `io.popen` in this test. So I am still confused as to what behavior this test is supposed to assert? Can you clarify?

t/pipe.t
thibaultcha: Resolved.

lib/ngx/pipe.lua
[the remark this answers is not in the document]
thibaultcha: Considering resolved since the macro refactor in the C implementation.

t/pipe.t
spacewander: @thibaultcha 
This test case shows that `ngx.pipe` is a little safer replacement to the `io.popen`, because it doesn't suffer from that problem. At that time, there is not `lua_sa_restart` directive.

spacewander:
@thibaultcha 
Updated the `proc:write` method.

lib/ngx/pipe.lua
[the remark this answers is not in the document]
thibaultcha: Yeah, that’s fine with me.

t/pipe.t
thibaultcha: Ok... why not. Marking this as resolved.

PR #218, opened 2018-11-13
bugfix: avoided generating Content-Type header when getting/setting o…

spacewander:
…ther response headers.

### DIFF
--- a/lib/resty/core/response.lua
+++ b/lib/resty/core/response.lua
@@ -145,10 +145,6 @@ local function get_resp_header(tb, key)
         error("API disabled in the current context")
     end
 
-    if n == FFI_ERROR then
-        error(ffi_str(errmsg[0]))
-    end
-
     if n == 0 then
         return nil
     end
@@ -167,7 +163,8 @@ local function get_resp_header(tb, key)
         return ret
     end
 
-    error("unreachable")
+    -- n == FFI_ERROR
+    error(ffi_str(errmsg[0]))
 end
 
 
--- a/lib/resty/core/response.lua
+++ b/lib/resty/core/response.lua
@@ -12,7 +12,6 @@ local new_tab = base.new_tab
 local FFI_BAD_CONTEXT = base.FFI_BAD_CONTEXT
 local FFI_NO_REQ_CTX = base.FFI_NO_REQ_CTX
 local FFI_DECLINED = base.FFI_DECLINED
-local FFI_ERROR = base.FFI_ERROR
 local get_string_buf = base.get_string_buf
 local getmetatable = getmetatable
 local type = type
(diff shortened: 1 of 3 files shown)


spacewander:
This is the sister pr of https://github.com/openresty/lua-nginx-module/pull/1422.

t/response.t
```diff
@@ -236,3 +236,64 @@ qr/\[TRACE\s+\d+ content_by_lua\(nginx\.conf:\d+\):5 (?:loop|-> \d+)\]/
...
+        }
+    }
+--- request
+GET /t
+--- response_body
+foo
+--- response_headers
+Content-Type:
+--- no_error_log
+[error]
```
thibaultcha: imho it is fine keeping these tests here, but they are duplicate of the sister PR's tests. Running the ngx_lua test suite with the resty.core module is also a way to run these tests.

In fact, changes like [this one](https://github.com/openresty/lua-nginx-module/pull/1422/files#diff-8593203ffbe2a651fb1761cebff6c139R1409) are only tested from within the ngx_lua test suite (when running with the resty.core module). So if the goal is to port ngx.resp tests to this module (resty.core), then that is a missing test. Is that the goal?

lib/resty/core/response.lua
```diff
@@ -161,8 +167,7 @@ local function get_resp_header(tb, key)
         return ret
     end
 
-    -- n == FFI_ERROR
-    error("no memory")
+    error("unreachable")
```
thibaultcha: Maybe better to follow the existing `set_resp_header` pattern (because already existing) and assume that the default case is `FFI_ERROR`? As in: https://github.com/openresty/lua-resty-core/blob/42d1d948dc4de045810f7a213b7f76aa206034ed/lib/resty/core/response.lua#L116-L117

spacewander: @thibaultcha 
Updated.

t/response.t
spacewander: @thibaultcha 
I decide to remove the duplicate tests, as we always have to run

```shell
TEST_NGINX_INIT_BY_LUA="package.path = '$PWD/../lua-resty-core/lib/?.lua;$PWD/../../git/lua-resty-lrucache/lib/?.lua;' .. (package.path or '') require('resty.core')" prove ...
```

to ensure all the test cases are passed.

thibaultcha:
@spacewander Did you notice the CI failures on this branch?

t/response.t
thibaultcha: Yes indeed :) Great choice, thank you!

spacewander:
@thibaultcha 
Solved the problem with rebase.

thibaultcha:
Merged, thank you!

PR #226, opened 2019-02-09
tests: hardened JIT-ability of the test suite and broadened support for the TEST_NGINX_HOTLOOP environment variable.

thibaultcha:
This also harden various test cases and removes a few globals caught by
the _G write guard.

### DIFF
--- a/t/TestCore.pm
+++ b/t/TestCore.pm
@@ -1,0 +1,50 @@
+package t::TestCore;
+
+use Test::Nginx::Socket::Lua -Base;
+use Cwd qw(cwd);
+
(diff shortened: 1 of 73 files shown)


t/TestCore.pm
```diff
@@ -0,0 +1,51 @@
...
+add_block_preprocessor(sub {
+    my $block = shift;
+
+    if (!defined $block->http_config) {
+        $block->set_value("http_config", $HttpConfig);
+    }
+});
+
+1;
+
```
agentzh: Extra blank line here?

thibaultcha: Fixed

PR #228, opened 2019-02-17
tests: avoided listening on 127.0.0.2:12345 on which iptables drops packets for ngx_lua and ngx_stream_lua tests suites.

thibaultcha:
On our AWS EC2 test cluster, these tests fail if the lua-resty-core test
suite gets executed on an instance also executing the ngx_lua or
ngx_stream_lua test suites, since the latter two have iptables rules to
drop packet on this specific IP/port couple.

Instead of removing the rule from opsboy, changing the port here will
also ease local development process for maintainers. Only lua-resty-core
seems to be using this IP/port couple from the OpenResty components.

### DIFF
--- a/t/ndk.t
+++ b/t/ndk.t
@@ -5,8 +5,6 @@ repeat_each(2);
 
 plan tests => repeat_each() * (blocks() * 5 + 2);
 
-$ENV{TEST_NGINX_HOTLOOP} ||= 9;
-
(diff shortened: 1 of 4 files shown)


thibaultcha:
On second thought, 54321 is in the ephemeral port range, so probably not a good idea. I'll update it to 23456.

PR #229, opened 2019-02-20
tests: pipe.t: checked proc:wait() return value when timed out.

spacewander:


spacewander:
This is the test case for https://github.com/openresty/lua-nginx-module/pull/1477.

PR #235, opened 2019-03-06
feature: added new Lua API for Nginx core's dynamic resolver.

slimboyfat:

### DIFF
--- a/lib/ngx/resolver.lua
+++ b/lib/ngx/resolver.lua
@@ -1,0 +1,85 @@
+local base = require "resty.core.base"
+local get_request = base.get_request
+local ffi = require "ffi"
+local C = ffi.C
+local ffi_new = ffi.new
+local ffi_str = ffi.string
+local ffi_gc = ffi.gc
+local FFI_OK = base.FFI_OK
+local FFI_ERROR = base.FFI_ERROR
+local FFI_DONE = base.FFI_DONE
+local co_yield = coroutine._yield
+
+local BUF_SIZE = 256
+local get_string_buf = base.get_string_buf
+local get_size_ptr = base.get_size_ptr
+
+base.allows_subsystem("http")
+
+ffi.cdef [[
+typedef intptr_t                         ngx_int_t;
+typedef unsigned char                    u_char;
+typedef struct ngx_http_lua_co_ctx_t    *curcoctx_ptr;
+typedef struct ngx_http_resolver_ctx_t  *rctx_ptr;
+
+typedef struct {
+    ngx_http_request_t              *request;
+    u_char                          *buf;
+    size_t                          *buf_size;
+    curcoctx_ptr                     curr_co_ctx;
+    rctx_ptr                         rctx;
+    ngx_int_t                        rc;
+    unsigned                         ipv4:1;
+    unsigned                         ipv6:1;
+} ngx_http_lua_resolver_ctx_t;
+
+int ngx_http_lua_ffi_resolve(ngx_http_lua_resolver_ctx_t *ctx, const char *hostname);
+
+void ngx_http_lua_ffi_resolver_destroy(ngx_http_lua_resolver_ctx_t *ctx);
+]]
+
+local _M = { version = base.version }
+
+local mt = {
+    __gc = C.ngx_http_lua_ffi_resolver_destroy
+}
+
+local Ctx = ffi.metatype("ngx_http_lua_resolver_ctx_t", mt)
+
+function _M.resolve(hostname, ipv4, ipv6)
+    assert(type(hostname) == "string", "hostname must be string")
+    assert(ipv4 == nil or type(ipv4) == "boolean", "ipv4 must be boolean or nil")
+    assert(ipv6 == nil or type(ipv6) == "boolean", "ipv6 must be boolean or nil")
+
+    local buf = get_string_buf(BUF_SIZE)
+    local buf_size = get_size_ptr()
+    buf_size[0] = BUF_SIZE
+
+    local ctx = Ctx({get_request(), buf, buf_size})
+
+    ctx.ipv4 = ipv4 == nil and 1 or ipv4
+    ctx.ipv6 = ipv6 == nil and 0 or ipv6
+
+    local rc = C.ngx_http_lua_ffi_resolve(ctx, hostname)
+
+    local res, err
+    if (rc == FFI_OK) then
+        res, err = ffi_str(buf, buf_size[0]), nil
+    elseif (rc == FFI_DONE) then
+        res, err = co_yield()
+    elseif (rc == FFI_ERROR) then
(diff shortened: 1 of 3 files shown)


lib/ngx/resolver.lua
```diff
@@ -0,0 +1,85 @@
...
+    local buf = get_string_buf(BUF_SIZE)
+    local buf_size = get_size_ptr()
+    buf_size[0] = BUF_SIZE
+
+    local ctx = Ctx({get_request(), buf, buf_size})
+
+    ctx.ipv4 = ipv4 == nil and 1 or ipv4
+    ctx.ipv6 = ipv6 == nil and 0 or ipv6
+
+    local rc = C.ngx_http_lua_ffi_resolve(ctx, hostname)
```
agentzh: You know that `ctx` is managed by Lua GC while your `ctx` may get collected as soon as this Lua function returns (or yields). I see you register `ctx` to `rctx` which will surely outlive this Lua function call once it yields.

slimboyfat: I really believed that coroutine function locals must not be collected by GC, until function returns (yield is not a case when local might be get freed by GC).

I spent several hours for reading manual and googling. Unfortunately, I did not find anything that clearly states what should be during a yield with locals.

Based on following I wrote several tests.
Lua: test suites https://www.lua.org/tests/ (for v5.1)
Lua Traverse http://code.matthewwild.co.uk/luatraverse/

```
local object_was_finalized

local function finalizer()
    object_was_finalized = true
end

local function func()
    local object = newproxy(true)
    getmetatable(object).__gc = finalizer
    local free = coroutine.yield()
    if free then
        object = nil
    end
    coroutine.yield()
end

-- Test #1
object_was_finalized = false
local t = coroutine.create(func)
coroutine.resume(t)
collectgarbage()
assert(object_was_finalized == false)
coroutine.resume(t, true)
collectgarbage()
assert(object_was_finalized == true)

-- Test #2 
object_was_finalized = false
t = coroutine.create(func)
coroutine.resume(t)
collectgarbage()
assert(object_was_finalized == false)
coroutine.resume(t, false)
collectgarbage()
assert(object_was_finalized == false)
coroutine.resume(t)
collectgarbage()
assert(object_was_finalized == true)
```
... and still not sure, if local variable can be GC-ed on yield.

Could you please point me to manual/link which covers this (GC on yield).

lib/ngx/resolver.lua
```diff
@@ -0,0 +1,85 @@
...
+        res, err = ffi_str(buf, buf_size[0]), nil
+    elseif (rc == FFI_DONE) then
+        res, err = co_yield()
+    elseif (rc == FFI_ERROR) then
+        res, err = nil, ffi_str(buf, buf_size[0])
+    else
+        res, err = nil, "unknown error"
+    end
+
+    C.ngx_http_lua_ffi_resolver_destroy(ffi_gc(ctx, nil))
```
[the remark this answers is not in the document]
slimboyfat: Let me explain my idea in step-by-step:

1. At C land I defined a finalizer which guarantees finalization of resolver's context allocated during a resolve call.

2. The finalizer is registered via  __gc metamethod for FFI cdata object Ctx.

At this point we have a guarante that the finalizer will be eventually called, as http://luajit.org/ext_ffi_api.html states that:

> A cdata finalizer works like the __gc metamethod for userdata objects: when the last reference to a cdata object is gone, the associated finalizer is called with the cdata object as an argument.
> ...
> An existing finalizer can be removed by setting a nil finalizer, e.g. right before explicitly deleting a resource:
> ...

But in Lua 5.1 manual (https://www.lua.org/manual/5.1/manual.html) 
at "2.10.1 – Garbage-Collection Metamethods" we can find following note:

> At the end of each garbage-collection cycle, the finalizers for userdata are called in reverse order of their creation, among those collected in that cycle. That is, the first finalizer to be called is the one associated with the userdata created last in the program. The userdata itself is freed only in the next garbage-collection cycle.

Thus object with finalizer requires 2 GC cycles (1st GC-cycle - call finalizer, 2nd cycle - memory reclamation) to get fully freed.
That is why, in my code, I make a third step.

3. Perform explicit finalizator call and then unregister __gc finalizer.

`C.ngx_http_lua_ffi_resolver_destroy(ffi_gc(ctx, nil))`
is equvivalent for
```
C.ngx_http_lua_ffi_resolver_destroy(ctx) -- explicitly call finalizer
ffi_gc(ctx, nil) -- unregister finalizer (set __gc = nil at ctx's mt)
```
Now, ctx object will be freed in a single GC-cycle. So, from my point of view, it is just an optimization, which makes Lua heap cleaner a bit faster.

Please, let me know where I was wrong...

P.S.: I found several places in the OpenResty source code where the same idiom was used (e.g.: regex.lua - ngx_lua_ffi_destroy_regex(ffi_gc(compiled, nil)))

lib/ngx/resolver.lua
```diff
@@ -0,0 +1,85 @@
...
+typedef struct {
+    ngx_http_request_t              *request;
+    u_char                          *buf;
+    size_t                          *buf_size;
+    curcoctx_ptr                     curr_co_ctx;
+    rctx_ptr                         rctx;
+    ngx_int_t                        rc;
+    unsigned                         ipv4:1;
+    unsigned                         ipv6:1;
+} ngx_http_lua_resolver_ctx_t;
```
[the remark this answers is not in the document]
slimboyfat: Agree. But such struct allows avoid overhead of mem-pool creation/freeing + additional mem-allocations on C-land. So, it is just a trade-off between maintenance burden and additional cpu cycles.

What do you think?

lib/ngx/resolver.lua
slimboyfat: Additionally, I created a test with explicit lua_gc calls (from C-land) just before return into yielded-lua-thread and right after the return from resumed thread.
The test confirmed assertion that Ctx can get freed only after return from coroutine (not in the middle on yield).

Please, give me a direction/additional info on this.

mikz:
This looks useful, but maybe it would be better to design it more future-proof? What about using this for custom load balancing in balancer_by_lua? This would need to return all the resolved addresses, so the balancer code can do the decision. What about other classes than A records?  SRV for example?

If this API exists, I think it should aim to replace the need for https://github.com/openresty/lua-resty-dns.

slimboyfat:
Yes, it can be extended to retrieve all entries from DNS response and return Lua table. At the moment, only A and AAAA records are supported, but add support for SRV-records is not a big deal (as Nginx internal resolver supports SRV queries).

Benefits of proposed ngx.resolve() API are
- configurable caching policy; 
- ability to efficiently handle sequential queries for a same host; 

@agentzh had several remarks on correctness of proposed implementation, but he possible is busy and can not help me.

Probably, someone else can help me finish my work.

PR #241, opened 2019-04-23
bugfix: remove redundant require in core.lua

chronolaw:
It seems there are some redundant code in core.lua

So I remove them, please check it, thanks.


spacewander:
@chronolaw 
It would be great to sort the module to make the code easier to read.
If you do so, note that the regex module should be loaded first to inject `ngx.re` for other modules.

chronolaw:
I'm sorry that I would not change the sequence of modules, because I am not familiar with all lua core modules.

And what is the easiest way to read? alphabet or other?

spacewander:
It's OK. Whether to sort it or not is up to you.

thibaultcha:
@chronolaw Thank you!

PR #242, opened 2019-04-25
feature: supported specifying environment variables when spawning the…

spacewander:
… sub-process.

### DIFF
--- a/lib/ngx/pipe.lua
+++ b/lib/ngx/pipe.lua
@@ -11,6 +11,7 @@ local ipairs = ipairs
 local tonumber = tonumber
 local tostring = tostring
 local type = type
+local str_find = string.find
 local table_concat = table.concat
 local ffi = require "ffi"
 local C = ffi.C
@@ -46,7 +47,7 @@ typedef struct {
 
 int ngx_http_lua_ffi_pipe_spawn(ngx_http_lua_ffi_pipe_proc_t *proc,
     const char *file, const char **argv, int merge_stderr, size_t buffer_size,
-    u_char *errbuf, size_t *errbuf_size);
+    const char **environ, u_char *errbuf, size_t *errbuf_size);
 
 int ngx_http_lua_ffi_pipe_proc_read(ngx_http_request_t *r,
     ngx_http_lua_ffi_pipe_proc_t *proc, int from_stderr, int reader_type,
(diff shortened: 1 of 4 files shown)


spacewander:
This is the sister PR of https://github.com/openresty/lua-nginx-module/pull/1513

t/pipe.t
```diff
@@ -1546,3 +1546,57 @@ exit
...
+            spawn(environ)
+            table.clear(environ)
+            spawn(environ)
+            spawn("TEST_ENV=1")
+        }
+    }
+--- response_body
+bad environ option in index 2: string expected, got number
+bad environ option: non-empty table expected
+bad environ option: table expected, got string
```
thibaultcha: Should we add a test with a `nil` hole somewhere in the `environ` option table?

lib/ngx/pipe.lua
```diff
@@ -566,6 +567,38 @@ do
                     error("bad buffer_size option", 2)
                 end
             end
+
+            if opts.environ then
+                environ = opts.environ
+                local environ_type = type(environ)
+                if environ_type ~= 'table' then
```
thibaultcha: style: could we stick to double quotes to stay consistent with the rest of this function's body style?

lib/ngx/pipe.lua
```diff
@@ -566,6 +567,38 @@ do
...
+                    error("bad environ option: table expected, got "
+                          .. environ_type, 2)
+                end
+
+                local nenv = 0
+
+                for i, env in ipairs(environ)  do
```
thibaultcha: style: extra space before `do`

lib/ngx/pipe.lua
```diff
@@ -566,6 +567,38 @@ do
...
+                    if env == nil then
+                        break
+                    end
+
+                    nenv = nenv + 1
+
+                    local env_type = type(env)
+                    if env_type ~= 'string' then
+                        error("bad environ option in index " .. i
+                              .. ": string expected, got " .. env_type, 2)
```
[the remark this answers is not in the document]
spacewander: @thibaultcha 
Environment variable is passed like this format: `KEY=VALUE`. So a `number` argument should not be a valid one.
Maybe we could be more strict, check if there is a `=` in the argument.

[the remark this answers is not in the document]
thibaultcha: My bad, while reviewing this particular line I had a hash construct in mind for the `environ` option, which is wrong of course, since the current implementation is an array. Long day.

Such validation is a good idea imho, and adding some test cases with malformed environment strings as well.

spacewander:
@thibaultcha 
Updated and rebased.

thibaultcha:
@spacewander Thanks. I made a few more adjustments locally and added additional tests, but waiting for @agentzh to answer a process question I had before merging it.

thibaultcha:
Merged!

PR #247, opened 2019-05-03
optimize+bugfix: improved heuristics for the PCRE MAP_JIT workaround + support for binaries without PCRE

thibaultcha:

This is a follow up to d046562 which improves the heuristics to decide
whether PCRE JIT compilation should be disabled in init phase under
macOS or not. The heuristics now are:

- If under macOS
- If using PCRE 8.43 or greater
- Then PCRE JIT compilation will be disabled in the init phase

While PCRE JIT compilation is disabled as such, a notice log will notify
users attempting to use the `j` or `o` flags.

Once past the init phase, we revert the workaround and allow for the `j`
and `o` flags to be used, without any branching or additional
performance penalty.

New tests and minor refactorings have been included as well.

---

Also included: `bugfix: ensured resty.core can be loaded in binaries without PCRE support.`

### DIFF
--- a/lib/resty/core/regex.lua
+++ b/lib/resty/core/regex.lua
@@ -35,8 +35,32 @@ local ngx_log = ngx.log
 local ngx_NOTICE = ngx.NOTICE
 
 
+local _M = {
+    version = base.version
+}
+
+
 if not ngx.re then
     ngx.re = {}
+end
+
+
+ffi.cdef[[
+    const char *pcre_version(void);
+]]
+
+
+local pcre_ver
+
+if not pcall(function() pcre_ver = ffi_string(C.pcre_version()) end) then
+    setmetatable(ngx.re, {
+        __index = function(_, key)
+            error("no support for 'ngx.re." .. key .. "': OpenResty was " ..
+                  "compiled without PCRE support", 2)
+        end
+    })
+
+    return _M
 end
 
 
@@ -95,12 +119,6 @@ local ngx_lua_ffi_script_eval_data
 local no_jit_in_init
 
 if jit.os == "OSX" then
-    ffi.cdef[[
-        const char *pcre_version(void);
-    ]]
-
-    local pcre_ver = ffi_string(C.pcre_version())
-
     local maj, min = string.match(pcre_ver, "^(%d+)%.(%d+)")
     if maj and min then
         local pcre_ver_num = tonumber(maj .. min)
@@ -277,11 +295,6 @@ end
 local c_str_type = ffi.typeof("const char *")
 
 local cached_re_opts = new_tab(0, 4)
-
-local _M = {
-    version = base.version
-}
-
 
 local buf_grow_ratio = 2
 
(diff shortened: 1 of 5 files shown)


lib/resty/core/regex.lua
```diff
@@ -371,6 +390,79 @@ local function parse_regex_opts(opts)
 end
 
 
+if no_jit_in_init then
+    local parse_regex_opts_ = parse_regex_opts
+
+    parse_regex_opts = function(opts)
```
agentzh: Style: needs a space before `(`. Please address other similar places.

lib/resty/core/request.lua
```diff
@@ -68,7 +68,7 @@ local truncated = ffi.new("int[1]")
 
 local req_headers_mt = {
     __index = function (tb, key)
-        return rawget(tb, (gsub(lower(key), '_', '-', "jo")))
+        return rawget(tb, (re.gsub(lower(key), '_', '-', "jo")))
```
agentzh: Alas. This is slower on all architectures. Here it relies on PCRE. Maybe we should introduce a C function to do this simple character replacement?

thibaultcha: @agentzh I just pushed a commit proposing an implementation. Do you think this new `str_replace_char` function should live in a different Lua module? `misc.lua` or a new `string.lua`? It isn't exposed as it stands and thus does not have tests yet. I will add them based on your feedback.

.travis.yml
```diff
@@ -63,7 +63,7 @@ install:
   - git clone https://github.com/openresty/openresty.git ../openresty
   - git clone https://github.com/openresty/openresty-devel-utils.git
   - git clone https://github.com/simpl/ngx_devel_kit.git ../ndk-nginx-module
-  - git clone https://github.com/openresty/lua-nginx-module.git ../lua-nginx-module
+  - git clone -b feat/ngx_http_lua_ffi_str_replace_char https://github.com/thibaultcha/lua-nginx-module.git ../lua-nginx-module
```
thibaultcha: Direct link to the FFI API on my fork: https://github.com/thibaultcha/lua-nginx-module/commit/51505338fbb5df261df670a75a2c4aab748e6b3c

lib/resty/core/request.lua
agentzh: @thibaultcha Could be in ` resty.core.base`.

agentzh: @thibaultcha Or `resty.core.util`. Maybe this is better.

thibaultcha:
Sister PR at https://github.com/openresty/lua-nginx-module/pull/1523 for the `str_replace_char()` FFI API.

lib/resty/core/regex.lua
```diff
@@ -79,24 +103,36 @@ local ngx_lua_ffi_init_script_engine
...
+-- cached when running under macOS, even if the user requests so. Caching is
+-- thus disabled to prevent further calls of same regex in worker to have poor
+-- performance.
+--
+-- TODO: improve this workaround when PCRE allows for unspecifying the MAP_JIT
+-- option.
+local no_jit_in_init
 
+if jit.os == "OSX" then
+    pcre_ver = ffi_string(pcre_ver)
```
agentzh: It's not a great idea to use the same Lua variable name for values of completely differently types.

thibaultcha: I have to say I concur.

PR #252, opened 2019-05-07
bugfix: support calls to os.getenv wrapper even after unwrapping

lukego:
Fix a bug where the `os.getenv` wrapper function in misc.lua stops working after it "unwraps" and restores the original binding of `os.getenv`.

The wrapper function can still be referenced and called by a user and so we need to keep it working even after the original binding of `os.getenv` has been restored.

Consider this example:

    local getenv = os.getenv
    getenv("foo")
    getenv("bar")

which is problematic if *both* calls attempt to unwrap `os.getenv`.

Here is a real-world example that does trigger an error:
https://github.com/daurnimator/lua-http/blob/master/http/proxies.lua#L18-L32

I hereby granted the copyright of the changes in this pull request to the authors of this lua-resty-core project.

### DIFF
--- a/lib/resty/core/misc.lua
+++ b/lib/resty/core/misc.lua
@@ -185,17 +185,22 @@ end
 
 
 do
+    local finished_wrapping_getenv = false
     local _getenv = os.getenv
     local env_ptr = ffi_new("unsigned char *[1]")
 
     os.getenv = function (name)
+        if finished_wrapping_getenv then
+            -- os.getenv is "unwrapped" but this function was called anyway
+            return _getenv(name)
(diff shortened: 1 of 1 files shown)


thibaultcha:
Hey @lukego :wave: 

Thanks for the patch! As communicated, I have another version of it ready (with tests), which I'll probably push tomorrow. Cheers!

lukego:
Roger that, closing this PR.

thibaultcha:
PR submitted at #253 - please give it a try as well @lukego, and let me know if it resolves the issue on your end as well.

PR #278, opened 2019-09-18
feature: add function `tcpsock:setclientcert`, reimplemented `tcpsock:sslhandshake` with FFI.

dndx:
functions using FFI.

Sister PR: https://github.com/openresty/lua-nginx-module/pull/1602

Note that stream support is intentionally omitted in this PR to keep the size down. I will help porting this over to stream after it is merged.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,7 +65,7 @@ install:
   - git clone https://github.com/openresty/openresty.git ../openresty
   - git clone https://github.com/openresty/openresty-devel-utils.git
   - git clone https://github.com/simpl/ngx_devel_kit.git ../ndk-nginx-module
-  - git clone https://github.com/openresty/lua-nginx-module.git ../lua-nginx-module
+  - git clone -b feat/cosocket_tlshandshake https://github.com/dndx/lua-nginx-module.git ../lua-nginx-module
   - git clone https://github.com/openresty/no-pool-nginx.git ../no-pool-nginx
   - git clone https://github.com/openresty/echo-nginx-module.git ../echo-nginx-module
   - git clone https://github.com/openresty/lua-resty-lrucache.git
--- a/README.markdown
+++ b/README.markdown
@@ -271,6 +271,8 @@ in the current request before you reusing the `ctx` table in some other place.
 ## resty.core.socket
 
 * [socket.setoption](https://github.com/openresty/lua-nginx-module#tcpsocksetoption)
+* [socket.setclientcert](https://github.com/openresty/lua-nginx-module#tcpsocksetclientcert)
+* [socket.sslhandshake](https://github.com/openresty/lua-nginx-module#tcpsocksslhandshake)
 
 [Back to TOC](#table-of-contents)
 
--- a/lib/resty/core/socket.lua
+++ b/lib/resty/core/socket.lua
@@ -1,18 +1,27 @@
 local base = require "resty.core.base"
-base.allows_subsystem('http')
-local debug = require 'debug'
-local ffi = require 'ffi'
+base.allows_subsystem("http")
+local debug = require "debug"
+local ffi = require "ffi"
 
 
-local error = error
+local error    = error
+local assert   = assert
 local tonumber = tonumber
-local registry = debug.getregistry()
-local ffi_new = ffi.new
-local ffi_string = ffi.string
-local C = ffi.C
-local get_string_buf = base.get_string_buf
-local get_size_ptr = base.get_size_ptr
 local tostring = tostring
+local type     = type
+local select   = select
+local registry = debug.getregistry()
+
+local C       = ffi.C
+local ffi_new = ffi.new
+local ffi_str = ffi.string
+local ffi_gc  = ffi.gc
+
+local get_string_buf = base.get_string_buf
+local get_size_ptr   = base.get_size_ptr
+local get_request    = base.get_request
+
+local co_yield = coroutine._yield
 
 
 local option_index = {
@@ -35,13 +44,35 @@ ngx_http_lua_ffi_socket_tcp_getoption(ngx_http_lua_socket_tcp_upstream_t *u,
 int
 ngx_http_lua_ffi_socket_tcp_setoption(ngx_http_lua_socket_tcp_upstream_t *u,
     int opt, int val, unsigned char *err, size_t *errlen);
+
+int
+ngx_http_lua_ffi_socket_tcp_sslhandshake(ngx_http_request_t *r,
+    ngx_http_lua_socket_tcp_upstream_t *u, void *sess,
+    int enable_session_reuse, ngx_str_t *server_name, int verify,
+    int ocsp_status_req, void *chain, void *pkey, char **errmsg);
+
+int
+ngx_http_lua_ffi_socket_tcp_get_sslhandshake_result(ngx_http_request_t *r,
+    ngx_http_lua_socket_tcp_upstream_t *u, void **sess, char **errmsg,
+    int *openssl_error_code);
+
+void
+ngx_http_lua_ffi_ssl_free_session(void *sess);
 ]]
 
 
 local output_value_buf = ffi_new("int[1]")
-local FFI_OK = base.FFI_OK
-local SOCKET_CTX_INDEX = 1
 local ERR_BUF_SIZE = 4096
+
+local FFI_OK         = base.FFI_OK
+local FFI_ERROR      = base.FFI_ERROR
+local FFI_DONE       = base.FFI_DONE
+local FFI_AGAIN      = base.FFI_AGAIN
+local FFI_NO_REQ_CTX = base.FFI_NO_REQ_CTX
+
+local SOCKET_CTX_INDEX          = 1
+local SOCKET_CLIENT_CERT_INDEX  = 6
+local SOCKET_CLIENT_PKEY_INDEX  = 7
 
 
 local function get_tcp_socket(cosocket)
@@ -75,7 +106,7 @@ local function getoption(cosocket, option)
                                                        err,
                                                        errlen)
     if rc ~= FFI_OK then
-        return nil, ffi_string(err, errlen[0])
+        return nil, ffi_str(err, errlen[0])
     end
 
     return tonumber(output_value_buf[0])
@@ -107,10 +138,125 @@ local function setoption(cosocket, option, value)
                                                        err,
                                                        errlen)
     if rc ~= FFI_OK then
-        return nil, ffi_string(err, errlen[0])
+        return nil, ffi_str(err, errlen[0])
     end
 
     return true
+end
+
+
+local errmsg             = base.get_errmsg_ptr()
+local session_ptr        = ffi_new("void *[1]")
+local server_name_str    = ffi_new("ngx_str_t[1]")
+local openssl_error_code = ffi_new("int[1]")
+
+
+local function setclientcert(cosocket, cert, pkey)
+    if not cert and not pkey then
+        cosocket[SOCKET_CLIENT_CERT_INDEX] = nil
+        cosocket[SOCKET_CLIENT_PKEY_INDEX] = nil
+        return true
+    end
+
+    if not cert or not pkey then
+        return nil,
+               "client certificate must be supplied with corresponding " ..
+               "private key"
+    end
+
+    if type(cert) ~= "cdata" then
+        return nil, "bad cert arg: cdata expected, got " .. type(cert)
+    end
+
+    if type(pkey) ~= "cdata" then
+        return nil, "bad pkey arg: cdata expected, got " .. type(pkey)
+    end
+
+    cosocket[SOCKET_CLIENT_CERT_INDEX] = cert
+    cosocket[SOCKET_CLIENT_PKEY_INDEX] = pkey
+
+    return true
+end
+
+
+local function sslhandshake(cosocket, reused_session, server_name, ssl_verify,
+    send_status_req, ...)
+
+    local n = select("#", ...)
+    if not cosocket or n > 0 then
+        error("ngx.socket sslhandshake: expecting 1 ~ 5 arguments " ..
+              "(including the object), but seen " .. (cosocket and 5 + n or 0))
+    end
+
+    local r = get_request()
+    if not r then
+        error("no request found", 2)
+    end
+
+    session_ptr[0] = type(reused_session) == "cdata" and reused_session or nil
+
+    if server_name then
+        server_name_str[0].data = server_name
+        server_name_str[0].len = #server_name
+
+    else
+        server_name_str[0].data = nil
+        server_name_str[0].len = 0
+    end
+
+    local u = get_tcp_socket(cosocket)
+
+    local rc = C.ngx_http_lua_ffi_socket_tcp_sslhandshake(r, u,
+                   session_ptr[0],
+                   reused_session ~= false,
+                   server_name_str,
+                   ssl_verify and 1 or 0,
+                   send_status_req and 1 or 0,
+                   cosocket[SOCKET_CLIENT_CERT_INDEX],
+                   cosocket[SOCKET_CLIENT_PKEY_INDEX],
+                   errmsg)
+
+    if rc == FFI_NO_REQ_CTX then
+        error("no request ctx found", 2)
+    end
+
+    while true do
+        if rc == FFI_ERROR then
+            if openssl_error_code[0] ~= 0 then
+                return nil, openssl_error_code[0] .. ": " .. ffi_str(errmsg[0])
+            end
+
+            return nil, ffi_str(errmsg[0])
+        end
+
+        if rc == FFI_DONE then
+            return reused_session
+        end
+
+        if rc == FFI_OK then
+            if reused_session == false then
+                return true
+            end
+
+            rc = C.ngx_http_lua_ffi_socket_tcp_get_sslhandshake_result(r, u,
+                     session_ptr, errmsg, openssl_error_code)
+
+            assert(rc == FFI_OK)
+
+            if session_ptr[0] == nil then
+                return session_ptr[0]
+            end
+
+            return ffi_gc(session_ptr[0], C.ngx_http_lua_ffi_ssl_free_session)
+        end
+
+        assert(rc == FFI_AGAIN)
+
+        co_yield()
+
+        rc = C.ngx_http_lua_ffi_socket_tcp_get_sslhandshake_result(r, u,
+                 session_ptr, errmsg, openssl_error_code)
+    end
 end
 
 
@@ -118,6 +264,8 @@ do
     local method_table = registry.__tcp_cosocket_mt
     method_table.getoption = getoption
     method_table.setoption = setoption
+    method_table.setclientcert = setclientcert
+    method_table.sslhandshake  = sslhandshake
 end
 
 
--- a/t/ocsp.t
+++ b/t/ocsp.t
@@ -97,7 +97,7 @@ __DATA__
 GET /t
 --- response_body
 connected: 1
-ssl handshake: userdata
+ssl handshake: cdata
 
 --- error_log
 lua ssl server name: "test.com"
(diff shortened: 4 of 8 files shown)


lib/resty/core/socket_tcp.lua
```diff
@@ -172,7 +173,11 @@ do
     local old_socket_tcp = ngx.socket.tcp
 
     function ngx.socket.tcp()
-        local sock = old_socket_tcp()
+        local ok, sock = pcall(old_socket_tcp)
```
kikito: The commit which adds the pcall around here should be merged with d75e6f6c935810e0669e26309183d172fc571bbc, unless the intent is that this whole PR is collapsed into a single commit when merging.

lib/resty/core/socket_tcp.lua
```diff
@@ -0,0 +1,195 @@
...
+
+    local res, err = tlshandshake(self, cached_options)
+    table_clear(cached_options)
+
+    return res, err
+end
+
+
+do
+    local old_socket_tcp = ngx.socket.tcp
```
kikito: I have three comments about this monkeypatching.

* It doesn't seem to be done in other parts of `lua-resty-core`.
* It obscures the origin of the `ngx.socket.tcp` function . This might trip someone who (reasonably) expects ngx_http_lua_socket_tcp.h/c to be the "source of truth".
* It is not a very solid base on which to build upon. Would it be acceptable to have 4 different lua files, each one monkeypatching `ngx.socket.tcp` in a different way?

My suggestion would be to move `tlshandshake` and `sslhandshake` to the sister PR, https://github.com/openresty/lua-nginx-module/pull/1602 (perhaps translating them to C?) and adding them to the sockets metatable there.

Alternatively, I would expose the two Lua functions from here, and then I would include them inside the metatable in https://github.com/openresty/lua-nginx-module/pull/1602. I'm unclear if that's even possible.

dndx: Adding more Lua C functions is exactly what we do not want to do. The goal is to implement new cosocket methods using FFI without having to reimplement all the existing ones at the same time (because the PR will be unacceptably large and time consuming).

Eventually this will go away, after we migrated all the methods to use FFI instead. 

Also see: https://github.com/openresty/lua-nginx-module/pull/1599#discussion_r325342424

lib/resty/core/socket_tcp.lua
dndx: Yes, we usually do the squashing at merge time, to keep a reasonable review history.

lib/resty/core/socket_tcp.lua
kikito: Ok. Forget about implementing them in C then.

Would it be possible to implement them in FFI but *without* having to do this monkeypatching? The reason why it's necessary seems to be to be getting a hold of the socket's metatable. Perhaps it could be exposed directly, somewhere? `ngx.socket.tcp_mt`. (Yeah I agree that is a bit ugh. Maybe there is a better place).

But at the same time, exposing the metatable is only necessary in order to add methods to sockets. We could add a "hook" which does just that. Something like:

```
ngx.socket.tcp.new_method("tlshandshake", tlshandshake)
```

I like that this is very greppable and explicit. However it requires transforming into a "callable table" on the sister PR. If that is too much complication, this is my second-best suggestion:

```
ngx.socket.new_method("tcp", "tlshandshake", tlshandshake)
```

lib/resty/core/socket_tcp.lua
```diff
@@ -0,0 +1,190 @@
...
+                                                       and 1 or 0,
+                                                   client_cert,
+                                                   client_priv_key,
+                                                   errmsg)
+
+    if rc == FFI_NO_REQ_CTX then
+        error("no request ctx found", 2)
+    end
+
+::again::
```
thibaultcha: Could we use a `while true` loop here instead of relying on Lua 5.2 support? `ngx.pipe` already uses such a loop instead.

lib/resty/core/socket_tcp.lua
```diff
@@ -0,0 +1,190 @@
...
+end
+
+
+local function sslhandshake(self, reused_session, server_name, ssl_verify,
+    send_status_req, ...)
+
+    local n = select("#", ...)
+    if not self or n > 1 then
+        error("ngx.socket sslhandshake: expecting 1 ~ 5 "
+              .. "arguments (including the object), but seen " .. n)
```
thibaultcha: Given `n` will only contain arguments included in the variadic notation (`...`), this error message will be incorrect. It should be `5 + n`

dndx: Good catch, thanks for fixing!

dndx:
@thibaultcha Squashing at merge sounds good to me :)

README.markdown
```diff
@@ -241,6 +242,13 @@ API Implemented
 
 [Back to TOC](#table-of-contents)
 
+## resty.core.socket_tcp
```
agentzh: I don't quite like this module name. How about `resty.core.socket.tcp`?

spacewander (approved):

README.markdown
spacewander: Yes, we can rename the Lua file.

spacewander:
@dndx 
Do you have time to solve the conflict? Thanks!

dndx:
@spacewander not a problem, I will resolve it today!

README.markdown
dndx: Renamed.

lib/resty/core/socket/tcp.lua
```diff
@@ -0,0 +1,192 @@
...
+    local old_socket_tcp = ngx.socket.tcp
+
+    function ngx.socket.tcp()
+        local ok, sock = pcall(old_socket_tcp)
+        if not ok then
+            error(sock, 2)
+        end
+
+        sock.tlshandshake = tlshandshake
+        sock.sslhandshake = sslhandshake
```
doujiang24, line 183: @dndx I think it's better to move this module into this file? we can save two set field operations then.
https://github.com/openresty/lua-resty-core/blob/master/lib/resty/core/socket.lua

agentzh:
It's confusing to have both `tlshandshake` and `sslhandshake`.

lib/resty/core/socket/tcp.lua
```diff
@@ -0,0 +1,173 @@
...
+end
+
+
+local function sslhandshake(self, reused_session, server_name, ssl_verify,
+    send_status_req, ...)
+
+    local n = select("#", ...)
+    if not self or n > 1 then
+        error("ngx.socket sslhandshake: expecting 1 ~ 5 arguments " ..
+              "(including the object), but seen " .. (self and 5 + n or 0))
```
dndx, line 81: What's the reason for the `or 0` portion? Should it just be `5 + n`?

lib/resty/core/socket/tcp.lua
```diff
@@ -0,0 +1,173 @@
...
+        if not ok then
+            error(sock, 2)
+        end
+
+        sock.setclientcert = setclientcert
+        sock.sslhandshake  = sslhandshake
+
+        return sock
+    end
+end
```
dndx, line 168: We should move this entire file into https://github.com/openresty/lua-resty-core/blob/master/lib/resty/core/socket.lua instead as it already contains the TCP socket related FFI patching code.

lib/resty/core/socket/tcp.lua
chronolaw: done

lib/resty/core/socket/tcp.lua
chronolaw: done

lib/resty/core/socket/tcp.lua
chronolaw: It not a bug, see t/129-ssl-socket.t  test 31 in lua-nginx-module.

lib/resty/core/socket.lua
```diff
@@ -107,17 +133,129 @@ local function setoption(cosocket, option, value)
...
+              "private key", 2)
+    end
+
+    if type(cert) ~= "cdata" then
+        error("bad client cert type", 2)
+    end
+
+    if type(pkey) ~= "cdata" then
+        error("bad client pkey type", 2)
+    end
```
pintsized, line 167: Should we be returning `nil, err` here rather than calling `error`? These seem like reasonable runtime failures.

lib/resty/core/socket.lua
```diff
@@ -107,17 +133,129 @@ local function setoption(cosocket, option, value)
...
+local session_ptr = ffi_new("void *[1]")
+local server_name_str = ffi_new("ngx_str_t[1]")
+local openssl_error_code = ffi_new("int[1]")
+
+
+local function setclientcert(cosocket, cert, pkey)
+    if not cert and not pkey then
+        cosocket.client_cert = nil
+        cosocket.client_pkey = nil
+        return
```
pintsized, line 153: Should this return `true` to indicate success?

lib/resty/core/socket.lua
```diff
@@ -107,17 +133,129 @@ local function setoption(cosocket, option, value)
...
+        error("bad client cert type", 2)
+    end
+
+    if type(pkey) ~= "cdata" then
+        error("bad client pkey type", 2)
+    end
+
+    cosocket.client_cert = cert
+    cosocket.client_pkey = pkey
+end
```
pintsized, line 171: Again, should we return `true` to indicate success?

lib/resty/core/socket.lua
chronolaw: changed, thanks.

lib/resty/core/socket.lua
chronolaw: changed, thanks.

lib/resty/core/socket.lua
chronolaw: changed, thanks.

.travis.yml
```diff
@@ -65,7 +65,7 @@ install:
   - git clone https://github.com/openresty/openresty.git ../openresty
   - git clone https://github.com/openresty/openresty-devel-utils.git
   - git clone https://github.com/simpl/ngx_devel_kit.git ../ndk-nginx-module
-  - git clone https://github.com/openresty/lua-nginx-module.git ../lua-nginx-module
+  - git clone -b feat/cosocket_tlshandshake https://github.com/dndx/lua-nginx-module.git ../lua-nginx-module
```
pintsized, line 68: This will need updating before merging.

pintsized (changes requested):
Just a chore that needs completing before merging, otherwise I think this is ready.

.travis.yml
chronolaw: I think we should do it after we get the official confirm of openresty.

zhuizhuhaomeng:
@pintsized  will review this PR when I have time.

lib/resty/core/socket.lua
```diff
@@ -107,17 +133,132 @@ local function setoption(cosocket, option, value)
...
+
+local errmsg = base.get_errmsg_ptr()
+local session_ptr = ffi_new("void *[1]")
+local server_name_str = ffi_new("ngx_str_t[1]")
+local openssl_error_code = ffi_new("int[1]")
+
+
+local function setclientcert(cosocket, cert, pkey)
+    if not cert and not pkey then
+        cosocket.client_cert = nil
```
doujiang24, line 151: better use number instead of a string as key index here? since array has better performance than hash.
we can reach it just like `SOCKET_CTX_INDEX`, right?

lib/resty/core/socket.lua
```diff
@@ -107,17 +133,132 @@ local function setoption(cosocket, option, value)
...
+                   server_name_str,
+                   ssl_verify and 1 or 0,
+                   send_status_req and 1 or 0,
+                   cosocket.client_cert, cosocket.client_pkey, errmsg)
+
+    if rc == FFI_NO_REQ_CTX then
+        error("no request ctx found", 2)
+    end
+
+    while true do
```
doujiang24, line 216: this looks weird to me. we'd better write simple code even it may be a bit longer.

lib/resty/core/socket.lua
```diff
@@ -107,17 +133,132 @@ local function setoption(cosocket, option, value)
...
+            end
+
+            return ffi_gc(session_ptr[0], C.ngx_http_lua_ffi_ssl_free_session)
+        end
+
+        assert(rc == FFI_AGAIN)
+
+        co_yield()
+
+        rc = C.ngx_http_lua_ffi_socket_tcp_get_sslhandshake_result(r, u,
```
doujiang24, line 250: we'd better handle the error or ok here, no need a while loop here?

doujiang24 (commented):
The new API looks much better. Nice Job!

lib/resty/core/socket.lua
chronolaw: done, thanks.

lib/resty/core/socket.lua
chronolaw: done, please check again, thanks.

lib/resty/core/socket.lua
```diff
@@ -107,17 +138,139 @@ local function setoption(cosocket, option, value)
...
+        co_yield()
+    end
+
+    -- (rc == FFI_OK and reused_session ~= false) or co_yield()
+
+    rc = C.ngx_http_lua_ffi_socket_tcp_get_sslhandshake_result(r, u,
+            session_ptr, errmsg, openssl_error_code)
+
+    if rc == FFI_OK then
+        if reused_session == false then
```
doujiang24, line 245: this condition will always be false, right?

lib/resty/core/socket.lua
```diff
@@ -107,17 +138,139 @@ local function setoption(cosocket, option, value)
...
+            return true
+        end
+
+        if session_ptr[0] == nil then
```
doujiang24, line 249: this means something unexpected happened, right?

lib/resty/core/socket.lua
chronolaw: No, reused_session may be a cdata, we must check it after co_yield.

chronolaw: update some comments to make it more clear.

lib/resty/core/socket.lua
chronolaw: Yes, just in case.

doujiang24: it's better to assert here?

chronolaw: done

doujiang24: these three lines should be deleted?
https://github.com/openresty/lua-resty-core/pull/278/files#diff-ad5d7451bf606390f484541e17cda6d726608520667f836110ab6f34651adffdR250-R252

zhuizhuhaomeng:
@chronolaw  would you please rebase to the master ?

lib/resty/core/socket.lua
```diff
@@ -107,17 +138,142 @@ local function setoption(cosocket, option, value)
...
+
+
+local function sslhandshake(cosocket, reused_session, server_name, ssl_verify,
+    send_status_req, ...)
+
+    local n = select("#", ...)
+    if not cosocket or n > 1 then
+        error("ngx.socket sslhandshake: expecting 1 ~ 5 arguments " ..
+              "(including the object), but seen " .. (cosocket and 5 + n or 0))
+    end
```
zhuizhuhaomeng, line 189:
```suggestion
```

zhuizhuhaomeng: I don't think we need this test.
And if you want to test the number, it is 
`if not cosocket or n > 0`

lib/resty/core/socket.lua
```diff
@@ -107,17 +138,142 @@ local function setoption(cosocket, option, value)
...
+    rc = C.ngx_http_lua_ffi_socket_tcp_get_sslhandshake_result(r, u,
+            session_ptr, errmsg, openssl_error_code)
+
+    if rc == FFI_OK then
+        if reused_session == false then
+            return true
+        end
+
+        if session_ptr[0] == nil then
+            return nil
```
zhuizhuhaomeng, line 251: should return the error at the same time

zhuizhuhaomeng (changes requested):

chronolaw:
It just follows the idiom of openresty, like `settimeout` `sslhandshake`, these APIs don't use '_'.
https://github.com/openresty/lua-resty-core/pull/278#discussion_r819580250

chronolaw:
It makes test case #31 happy in `129-ssl-socket.t` of `lua-nginx-module`.

https://github.com/openresty/lua-resty-core/pull/278#discussion_r819617294

chronolaw:
It works like original C function `ngx_http_lua_ssl_handshake_retval_handler` in `ngx_http_lua_socket_tcp.c`, just for compatibility.

https://github.com/openresty/lua-resty-core/pull/278#discussion_r819632567

lib/resty/core/socket.lua
dndx: I think we can remove this "if session_ptr[0] == nil then" check. The `assert(session_ptr[0] ~= nil)` below should be sufficient.

lib/resty/core/socket.lua
dndx: @chronolaw I think @zhuizhuhaomeng is correct here. Since the Lua implementation of `sslhandshake` already takes 5 arguments, it will be incorrect to check `not cosocket or n > 1` as that indicates there are 7 arguments passed. It should be `not cosocket or n > 0` instead.

lib/resty/core/socket.lua
dndx: See my reviews, I think we can remove this `if session_ptr[0] == nil then` branch altogether and keep only the `assert` below.

dndx: From @chronolaw:

@zhuizhuhaomeng @doujiang24 This line is needed, as https://github.com/openresty/lua-nginx-module/blob/master/src/ngx_http_lua_socket_tcp.c#L1940-L1942 indicates the session could indeed be `NULL` even in case of a successful handshake (e.g. if the server disallows session reuse).

Therefore, the `assert` below is not necessary, as it can not replace the `if` condition here.

dndx: Actually, I had a closer look at this logic compared to https://github.com/openresty/lua-nginx-module/blob/master/src/ngx_http_lua_socket_tcp.c#L1935-L1956. The issue is that:

In the original C implementation, when the server did not turn on session reuse, the `sslhandshake` function returned a `lightuserdata` that is `NULL` and did not set the `__gc` meta method on it. This is different than returning a Lua `nil` value which indicates an error.

Therefore, we should mimic this behavior by returning the `session_ptr[0]` `cdata` directly and not call `ffi.gc` on it. From the user's perspective, the handshake is still a success.

lib/resty/core/socket.lua
```diff
@@ -107,17 +138,142 @@ local function setoption(cosocket, option, value)
...
 end
 
 
+local errmsg             = base.get_errmsg_ptr()
+local session_ptr        = ffi_new("void *[1]")
+local server_name_str    = ffi_new("ngx_str_t[1]")
+local openssl_error_code = ffi_new("int[1]")
+
+
+local function setclientcert(cosocket, cert, pkey)
```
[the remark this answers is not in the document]
dndx: This is to keep the style consistent with existing cosocket API function names. e.g. `tcpsock:getreusedtimes`

dndx:
@zhuizhuhaomeng Could you please take another look at this PR? It has been rebased and squashed and now passed all CI tests with it's lua-nginx-module counterpart.

doujiang24 (approved):

zhuizhuhaomeng:
merged with the following patch
```
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,7 +65,7 @@ install:
-  - git clone -b feat/cosocket_tlshandshake https://github.com/dndx/lua-nginx-module.git ../lua-nginx-module
+  - git clone https://github.com/dndx/lua-nginx-module.git ../lua-nginx-module
```

PR #287, opened 2020-02-24
chore: we already supported set request header with table

spacewander:

### DIFF
--- a/lib/resty/core/request.lua
+++ b/lib/resty/core/request.lua
@@ -389,14 +389,7 @@ do
     _M.set_req_header = set_req_header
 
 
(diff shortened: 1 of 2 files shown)


lib/resty/core/request.lua
```diff
@@ -389,14 +389,7 @@ do
     _M.set_req_header = set_req_header
 
 
-    local orig_func = ngx.req.set_header
```
thibaultcha: Does this mean we can remove the non-FFI implementation of `ngx.req.set_header`? I haven't checked, just asking out loud.

spacewander: @thibaultcha 
Yes.

PR #291, opened 2020-04-01
ngx.req.get_headers return empty table with metatable

juzipeek:
when one request has none header,  use ngx.req.get_headers will get an empty table without metatable. i think ngx.req.get_headers shuld return `nil` or table with metatable.

### DIFF
--- a/t/request.t
+++ b/t/request.t
@@ -564,16 +564,14 @@ Foo: baz, 123
             end
             ngx.log(ngx.ERR, "headers type:", type(headers))
             ngx.log(ngx.ERR, "headers metatable type:", type(getmetatable(headers)))
-            ngx.say(type(headers))
-            ngx.say(type(getmetatable(headers)))
+            ngx.say(string.format("%s,%s",type(headers), type(getmetatable(headers))))
         }
     }
 
 --- raw_request eval
 "GET /t\r\n"
 --- response_body
-table
-table
+table,table
 --- http09
 --- error_code
 200
--- a/t/request.t
+++ b/t/request.t
@@ -8,7 +8,7 @@ use t::TestCore;
 
 repeat_each(2);
 
-plan tests => repeat_each() * (blocks() * 5 + 4);
+plan tests => repeat_each() * (blocks() * 5 + 5);
 
 #no_diff();
 #no_long_string();
--- a/t/request.t
+++ b/t/request.t
@@ -212,7 +212,7 @@ GET /t?a=3%200&foo%20bar=&a=hello&blah
 --- response_body
 a: 3 0, hello
 blah: true
-foo bar: 
+foo bar:
 --- error_log eval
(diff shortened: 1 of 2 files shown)


doujiang24:
@juzipeek 
I think you are right, but can you explain a bit why to need this change?

thibaultcha:
@juzipeek Please contribute a test case as well (the test case might explain why the change is needed too).

juzipeek:
sorry for late reply.
this is for http0.9 request,  use cmd do http0.9 request:
`
$ telnet 127.0.0.1 8000
GET /status
`

and openresty cfg is:
`
server {
    listen 8000 default_server;

    location /status {
        content_by_lua_block {
            local cjson = require "cjson.safe"

            local header = ngx.req.get_headers()
            ngx.log(ngx.ERR, "header:", type(header), cjson.encode(header))

            if type(header) == "table" then
                ngx.log(ngx.ERR, "mt:", cjson.encode(getmetatable(header)))                
            end

            ngx.say("end")
        }
    }
}
`

we can get out put:

`
2020/05/13 22:47:52 [error] 51868#0: *64829 [lua] status.lua:29: header:table{}, client: 127.0.0.1, server: , request: "GET /status"
2020/05/13 22:47:52 [error] 51868#0: *64829 [lua] status.lua:32: mt:null, client: 127.0.0.1, server: , request: "GET /status"
`

juzipeek:
when use http0.9 send a request without header, `ngx.req.get_headers` will return a table withot metatable.

doujiang24:
@juzipeek 
will you help to add an test like https://github.com/openresty/lua-resty-core/blob/0cc3931a6a2a206c01fe995f98fff5bdf0d8ab57/t/request.t#L20?

t/request.t
```diff
@@ -553,3 +553,19 @@ Foo: bar
 Foo: baz, 123
 --- no_error_log
 [error]
+
+=== TEST 18: ngx.req.get_header (metatable is nil)
```
doujiang24, line 557: need three blank lines before this line, you are recommended to use the `reindex` to reformated it automatically.

juzipeek:
why i add two blank line lead to build fail:

agentzh:
@juzipeek The `reindex` tool would format your `.t` files correctly. I'll apply the following patch for you:

```diff
diff --git a/t/request.t b/t/request.t
index edcecd1..f95352f 100644
--- a/t/request.t
+++ b/t/request.t
@@ -554,6 +554,8 @@ Foo: baz, 123
 --- no_error_log
 [error]

+
+
 === TEST 18: ngx.req.get_header (metatable is nil)
 --- config
     location = /t {
```

So it was not really resolved in the review comment above.

agentzh:
Merged after re-formatting the test files. Thanks.

agentzh:
I'm still a bit curious why you need the metatable for an empty table...Seems like you are doing something special here?

PR #293, opened 2020-04-10
Feature balancer disable ssl

vislee:

ref: https://github.com/openresty/lua-nginx-module/pull/1687

### DIFF
--- a/lib/ngx/balancer.lua
+++ b/lib/ngx/balancer.lua
@@ -22,6 +22,7 @@ local ngx_lua_ffi_balancer_set_current_peer
 local ngx_lua_ffi_balancer_set_more_tries
 local ngx_lua_ffi_balancer_get_last_failure
 local ngx_lua_ffi_balancer_set_timeouts -- used by both stream and http
+local ngx_lua_ffi_balancer_disable_ssl
 
 
 if subsystem == 'http' then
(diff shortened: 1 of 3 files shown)


doujiang24:
@vislee thanks for your contribution.
`disable_ssl` is not a good idea, `set_upstream_ttl(enabled)` should be better, we can decide `http` or `https` per upstream try.

vislee:
> @vislee thanks for your contribution.
> `disable_ssl` is not a good idea, `set_upstream_ttl(enabled)` should be better, we can decide `http` or `https` per upstream try.

@doujiang24  Yes, 'disable_ssl' is not accurate, but this feature relies on 'proxy_pass https://'. I'm concerned that 'set_upstream_ttl(enabled)' might be misused.

doujiang24:
>  but this feature relies on 'proxy_pass https://'.
@vislee yeah, I knew it, we can only clarify it in the document.
I think `set_upstream_ttl` is better since we can support this usage:
```
set_upstream_tls(false) -- first try
set_upstream_tls(true) -- second try
```
maybe we should put this api to another new c module? like lua-conf-nginx-module. thoughts? @thibaultcha @spacewander @rainingmaster

zhuizhuhaomeng:
I would like to bring to your attention that this feature has been rewritten and can be found in the following pull request URL: https://github.com/openresty/lua-resty-core/pull/460.

PR #305, opened 2020-07-06
Small opt for ffi.new

chronolaw:
There is a pattern in resty.core to localize ffi.*, but in some lua scripts it is forgotten.

So I add some code, I wish it will be helpful, thanks.


doujiang24:
@chronolaw thanks for your contribution, the `ffi.new` is used in the top-level, not in function level. so, it should not have a performance issue.
we'd better keep the old code style to save some code :)

PR #306, opened 2020-07-13
bugfix: added the missing 'ngx.req.start_time' to the stream subsystem.

thibaultcha:
The `ngx.req.start_time()` API is part of `request.lua` which is was
understandably not loaded for the stream subsystem. There existed no
tests for this API in the stream subsystem, so the removal of its
underlying CFunction resulted in `ngx.req.start_time` being `nil` with
no test cases to realize that.


spacewander:
We need a way to document which APIs are enabled in stream subsystem.

PR #313, opened 2020-09-25
feature: implement the setoption functionality.

poorsea:


lib/resty/core/socket.lua
```diff
@@ -0,0 +1,85 @@
...
+    return true
+end
+
+do
+    local method_table = registry.__tcp_cosocket_mt
+    method_table.getoption = getoption
+    method_table.setoption = setoption
+end
+
+return { version = base.version }
```
doujiang24, line 85: we usually keep the `\n` for the last line.

lib/resty/core/socket.lua
```diff
@@ -0,0 +1,85 @@
...
+int 
+ngx_http_lua_ffi_socket_tcp_getoption(ngx_http_lua_socket_tcp_upstream_t *u,
+    const char *opt, int *val, char **err_msg);
+
+int 
+ngx_http_lua_ffi_socket_tcp_setoption(ngx_http_lua_socket_tcp_upstream_t *u,
+    const char *opt, int val, char **err_msg); 
+]]
+
+local output_value_buf = ffi_new("int[1]")
```
doujiang24, line 27: `int` may not be safe here?

lib/resty/core/socket.lua
```diff
@@ -0,0 +1,98 @@
...
+
+    return tcp_socket
+end
+
+
+local function getoption(cosocket, option)
+    local tcp_socket = get_tcp_socket(cosocket)
+
+    if option_index[option] == nil then
+        return nil, "unsupported option"
```
zhuizhuhaomeng, line 57: you should add option argument to error message

lib/resty/core/socket.lua
poorsea: keepalive, reuseaddr and tcp-nodelay use boolean;
rcvbuf and sndbuf are limit by  /proc/sys/net/core/rmem_max and /proc/sys/net/core/wmem_max in linux, and maximum is 2^31-1.

doujiang24: better add a comment before this line, like:
```
int is safe here since the maximum is 2^31-1
```

lib/resty/core/socket.lua
```diff
@@ -0,0 +1,110 @@
...
+local output_value_buf = ffi_new("int[1]")
+local err_msg = base.get_errmsg_ptr()
+local FFI_OK = base.FFI_OK
+local SOCKET_CTX_INDEX = 1
+
+
+local function get_tcp_socket(cosocket)
+    local tcp_socket = cosocket[SOCKET_CTX_INDEX]
+    if not tcp_socket then
+        return nil, error("bad tcp socket")
```
doujiang24, line 46:
```suggestion
        error("bad tcp socket")
```

lib/resty/core/socket.lua
```diff
@@ -0,0 +1,110 @@
...
+
+    return tcp_socket
+end
+
+
+local function getoption(cosocket, option)
+    local tcp_socket = get_tcp_socket(cosocket)
+
+    if not option then
+        return nil, "nil option"
```
doujiang24, line 57:
```suggestion
        return nil, 'missing the "option" argument'
```

lib/resty/core/socket.lua
```diff
@@ -0,0 +1,110 @@
...
+    end
+
+    if option_index[option] == nil then
+        return nil, "unsupported option " .. option
```
doujiang24, line 61:
```suggestion
        return nil, "unsupported option " .. tostring(option)
```

lib/resty/core/socket.lua
```diff
@@ -0,0 +1,110 @@
...
+
+    return tonumber(output_value_buf[0])
+end
+
+
+local function setoption(cosocket, option, value)
+    local tcp_socket = get_tcp_socket(cosocket)
+
+    if not option then
+        return nil, "nil option"
```
doujiang24, line 80:
```suggestion
        return nil, 'missing the "option" argument'
```

lib/resty/core/socket.lua
```diff
@@ -0,0 +1,110 @@
...
+    end
+
+    if not value then
```
doujiang24, line 83: so, should we use `true/false` for the options like `keepalive` and `tcp-nodelay`?

lib/resty/core/socket.lua
```diff
@@ -0,0 +1,110 @@
...
+        return nil, "nil value"
+    end
+
+    if option_index[option] == nil then
+        return nil, "unsupported option " .. option
```
doujiang24, line 88:
```suggestion
        return nil, "unsupported option " .. tostring(option)
```

poorsea: done

lib/resty/core/socket.lua
poorsea: modify and support.

lib/resty/core/socket.lua
poorsea: done

lib/resty/core/socket.lua
poorsea: accept

lib/resty/core/socket.lua
poorsea: accept

lib/resty/core/socket.lua
poorsea: accept

lib/resty/core/socket.lua
poorsea: accept

lib/resty/core/socket.lua
poorsea: accept

lib/resty/core/socket.lua
poorsea: accept

lib/resty/core/socket.lua
poorsea: accept

t/socket-tcp-getoption.t
```diff
@@ -0,0 +1,288 @@
...
+
+            sock:close()
+        }
+    }
+--- request
+GET /t
+--- response_body_like eval
+qr/rcvbuf: \d+/
+--- no_error_log
+[error]
```
agentzh, line 288: Needs test cases for the cases that the C function setsockopt() returns errors and we can get the system error messages for the failures.

agentzh (changes requested):

t/socket-tcp-getoption.t
poorsea: it hard to create a test case for the C function for parameters check in lua and no socket parameters can be got in lua, so i test it manual by code the parameters to wrong value in C function.

lib/resty/core/socket.lua
```diff
@@ -0,0 +1,124 @@
...
+local output_value_buf = ffi_new("int[1]")
+local FFI_OK = base.FFI_OK
+local SOCKET_CTX_INDEX = 1
+local ERR_BUF_SIZE = 128
+
+
+local function get_tcp_socket(cosocket)
+    local tcp_socket = cosocket[SOCKET_CTX_INDEX]
+    if not tcp_socket then
+        error("bad tcp socket")
```
agentzh, line 50: This error is confusing. We could do better. What does this really mean? The socket is never created? Never connected? Errored? What is it?

t/socket-tcp-setoption.t
```diff
@@ -0,0 +1,725 @@
...
+            if not ok then
+                ngx.say("enable keepalive failed: ", err)
+                return
+            end
+            local v2, err = sock:getoption("keepalive")
+            if not v2 then
+                ngx.say("get enabled keepalive failed: ", err)
+                return
+            end
+            ngx.say("keepalive change from ", v1, " to ", v2)
```
agentzh, line 148: Grammar: "change" => "changes" or "changed". Fix all other similar places.

t/socket-tcp-setoption.t
```diff
@@ -0,0 +1,725 @@
...
+            end
+            ngx.say("sndbuf change from ", v1, " to ", v2)
+
+            sock:close()
+        }
+    }
+--- request
+GET /t
+--- response_body_like eval
+qr/sndbuf change from \d+ to \d+/
```
agentzh, line 678: This change is too permissive. You should use anchors like `\A` and `\z` in the regex to make sure there's no unexpected output before this line of output.

Fix other similar places.

agentzh (changes requested):

t/socket-tcp-getoption.t
```diff
@@ -0,0 +1,288 @@
...
+=== TEST 3: getoption before calling connect.
+--- config
+    set $port $TEST_NGINX_SERVER_PORT;
+
+    location /t {
+        content_by_lua_block {
+            require "resty.core.socket"
+
+            local sock = ngx.socket.tcp()
+            local sndbuf, err = sock:getoption("so_sndbuf")
```
agentzh, line 95: Using the wrong option name here? It's not testing what you meant to test.

agentzh (changes requested):

t/socket-tcp-setoption.t
```diff
@@ -0,0 +1,725 @@
...
+            end
+            local v1, err = sock:getoption("keepalive")
+            if not v1 then
+                ngx.say("get default keepalive failed: ", err)
+                return
+            end
+
+            ok, err = sock:setoption("keepalive", 1)
+            if not ok then
+                ngx.say("enable keepalive failed: ", err)
```
agentzh, line 140: Grammar: "enable" => "enabling". Fix all the other similar places.

agentzh (changes requested):

lib/resty/core/socket.lua
poorsea: accept, "bad tcp socket" => "socket is never created nor connected"

t/socket-tcp-setoption.t
poorsea: accept

t/socket-tcp-setoption.t
poorsea: accept

t/socket-tcp-getoption.t
poorsea: accept, "so_sndbuf" => "sndbuf"

t/socket-tcp-setoption.t
poorsea: accept

lib/resty/core/socket.lua
```diff
@@ -0,0 +1,124 @@
+local base = require "resty.core.base"
```
agentzh, line 1: You should handle the stream subsystem.

agentzh (changes requested):

lib/resty/core/socket.lua
poorsea: Has handled here: 
https://github.com/openresty/lua-resty-core/blob/a53e3e76dd0f26746c205b5dde9892c7d0d045d5/lib/resty/core.lua#L22

lib/resty/core/socket.lua
agentzh: The man page actually states it should be an int value for those setsockopt options we just implemented. So the value range is not the key, but the value type is.

agentzh (changes requested):

t/socket-tcp-setoption.t
```diff
@@ -0,0 +1,725 @@
+# vim:set ft= ts=4 sw=4 et fdm=marker:
+
+use lib '.';
+use t::TestCore;
+
+repeat_each(2);
+
+plan tests => repeat_each() * (blocks() * 3) - 2;
```
agentzh, line 8: We should make sure the test plan is irrelevant with the `repeat_each()` value otherwise test modes like the "check leak" mode. You can try it out yourself. cc @xiaocang

agentzh (changes requested):

t/socket-tcp-setoption.t
doujiang24, line 8:
```suggestion
plan tests => repeat_each() * (blocks() * 3 - 1);
```

poorsea: accept

poorsea: accept

lib/resty/core/socket.lua
agentzh: @poorsea That's not enough. This module should also have `base.allows_subsystem('http')`. See resty.core.ndk for an example.

t/socket-tcp-getoption.t
agentzh: Still not changed here?

t/socket-tcp-getoption.t
agentzh: One way to test this is to introduce a special API to allow overriding the underlying fd to an invalid number (but not -1), then setsockopt will return an error.

agentzh (changes requested):

lib/resty/core/socket.lua
agentzh: “nor” => "or"
"never" => "not"

doujiang24:
Merged after minor tweaks, thanks.

PR #317, opened 2020-10-09
tests: fixed port conflicts when running test in parallel.

xiaocang:

### DIFF
--- a/t/TestCore.pm
+++ b/t/TestCore.pm
@@ -3,6 +3,8 @@ package t::TestCore;
 use Test::Nginx::Socket::Lua -Base;
 use Cwd qw(cwd);
 
+sub gen_random_port (@);
+
 $ENV{TEST_NGINX_HOTLOOP} ||= 10;
 $ENV{TEST_NGINX_MEMCACHED_PORT} ||= 11211;
 
@@ -38,6 +40,7 @@ our @EXPORT = qw(
     $lua_package_path
     $init_by_lua_block
     $HttpConfig
+    gen_random_port
 );
 
 add_block_preprocessor(sub {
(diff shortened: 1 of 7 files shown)


t/ssl.t
```diff
@@ -18,6 +18,14 @@ $ENV{TEST_NGINX_LUA_PACKAGE_PATH} = "$t::TestCore::lua_package_path";
 $ENV{TEST_NGINX_HTML_DIR} ||= html_dir();
 $ENV{TEST_NGINX_SERVER_SSL_PORT} ||= 4443;
 
+add_block_preprocessor(sub {
+    my $block = shift;
```
zhuizhuhaomeng, line 22: unused variable?

xiaocang: Nice catch!

t/TestCore.pm
```diff
@@ -48,4 +51,47 @@ add_block_preprocessor(sub {
...
+    }
+
+    my $server_addr = server_addr;
+
+    srand $$;
+
+    my $random_port;
+    my $tries = 1000;
+    for (my $i = 0; $i < $tries; $i++) {
+        my $port = int(rand 60000) + 1025;
```
agentzh, line 72: The upper limit should be 65535 (exclusive)?

t/TestCore.pm
```diff
@@ -48,4 +51,47 @@ add_block_preprocessor(sub {
     }
 });
 
+# borrowed from Test::Nginx::Util
+sub gen_random_port (@) {
```
agentzh, line 55: I think this feature should go to the `Test::Nginx::Socket` class instead.

t/TestCore.pm
```diff
@@ -48,4 +51,47 @@ add_block_preprocessor(sub {
...
+
+        if (defined $sock) {
+            $sock->close();
+            $random_port = $port;
+            last;
+        }
+    }
+
+    if (!defined $random_port) {
+        bail_out "Cannot find an available listening port number after $tries attempts.\n";
```
agentzh, line 91: Always add the `$name - ` prefix the error message otherwise it's hard to debug.

t/TestCore.pm
```diff
@@ -48,4 +51,47 @@ add_block_preprocessor(sub {
     }
 });
 
+# borrowed from Test::Nginx::Util
+sub gen_random_port (@) {
+    my (@unavailable_port_list) = @_;
+
+    push @unavailable_port_list, $ENV{TEST_NGINX_SERVER_PORT};
```
agentzh, line 58: This is wrong. The test block can also override the server port IIRC. You should use the `$ServerPort` variable provided by Test::Nginx::Util here instead.

t/TestCore/Stream.pm
```diff
@@ -47,4 +50,47 @@ add_block_preprocessor(sub {
     }
 });
 
+# borrowed from Test::Nginx::Util
+sub gen_random_port (@) {
```
agentzh, line 54: It's bad to duplicate such code here.

t/ocsp.t
```diff
@@ -16,6 +16,14 @@ no_long_string();
 $ENV{TEST_NGINX_LUA_PACKAGE_PATH} = "$t::TestCore::lua_package_path";
 $ENV{TEST_NGINX_HTML_DIR} ||= html_dir();
 
+$ENV{TEST_NGINX_SERVER_SSL_PORT} ||= 4443;
+
+add_block_preprocessor(sub {
+    if ($Test::Nginx::Util::Randomize) {
+        $ENV{TEST_NGINX_SERVER_SSL_PORT} = gen_random_port;
```
agentzh, line 23: It's easier to add automatic support for special variables `$TEST_NGINX_RANDOM_PORT1`, `$TEST_NGINX_RANDOM_PORT2`, and etc., in the Test::Nginx::Socket test scaffold. So that the test files can directly use and reference those special variables without such duplicate boilerplate code in every related .t file.

agentzh (changes requested):

PR #321, opened 2020-10-19
tests: pipe.t: increase timeout to 1 second to make it more resilient in mockeagain(W) & valgrind mode.

xiaocang:

### DIFF
--- a/t/pipe-stderr.t
+++ b/t/pipe-stderr.t
@@ -783,7 +783,7 @@ end
             local f = io.open("$TEST_NGINX_HTML_DIR/a.lua")
             local code = f:read("*a")
             local proc = helper.run_lua_with_graceful_shutdown("$TEST_NGINX_HTML_DIR", code)
-            proc:set_timeouts(100, 100, 100, 100)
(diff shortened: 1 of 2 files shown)


doujiang24 (approved):

t/pipe-stderr.t
```diff
@@ -783,7 +783,7 @@ end
             local f = io.open("$TEST_NGINX_HTML_DIR/a.lua")
             local code = f:read("*a")
             local proc = helper.run_lua_with_graceful_shutdown("$TEST_NGINX_HTML_DIR", code)
-            proc:set_timeouts(100, 100, 100, 100)
+            proc:set_timeouts(1000, 1000, 1000, 1000)
```
agentzh, line 786: It's usually not a good idea to let tests run this long. How about introducing a shorter file for reading here? `a.lua` is a bit large.

agentzh (changes requested):

t/pipe-stderr.t
xiaocang: I've changed the timeout time to 300ms and found it is good enough for testing stability.
`a.lua` will not be adjusted for this time.

PR #333, opened 2021-01-05
feature: it is harmful to set ngx.ctx to a non-table value

spacewander:


lib/resty/core/ctx.lua
```diff
@@ -107,6 +108,11 @@ register_getter("ctx", get_ctx_table)
 
 
 local function set_ctx_table(ctx)
+    local ctx_type = type(ctx)
+    if ctx_type ~= "table" then
+        error("expected ctx to be table, got " .. ctx_type)
```
doujiang24, line 113:
```suggestion
        error("ctx should be a table while getting a " .. ctx_type)
```

doujiang24 (commented):
otherwise, it looks good to me :)

doujiang24 (approved):

PR #361, opened 2021-09-18
bugfix: ngx.re.sub/gsub incorrectly dropped the last character.

balusch:
In order to avoid an infinite loop when the regex matches an empty
sub string, we set the match offset(`pos`) one character ahead of
cap\[1\](`cp_pos`) and begin a new match process(in ngx.re.gsub). Then
we break out the loop and make a comparison with #subj to check
whether there is a trailer to concat, in which we should not use
the match offset but the last cap[1].

### DIFF
--- a/t/re-bugs.t
+++ b/t/re-bugs.t
@@ -4,7 +4,7 @@ use t::TestCore;
 
 repeat_each(2);
 
-plan tests => repeat_each() * (blocks() * 4);
+plan tests => repeat_each() * (blocks() * 4 - 4);
 
 add_block_preprocessor(sub {
     my $block = shift;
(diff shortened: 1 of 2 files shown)


zhuizhuhaomeng (approved):

t/re-bugs.t
```diff
@@ -137,3 +137,40 @@ c0
 disabled in init phase under macOS
 --- skip_eval
 4: $^O ne 'linux'
+
+
+
+=== TEST 5: bug: sub incorrectly dropped the last character
+--- config
+    location /re {
+        content_by_lua '
```
spacewander, line 146: Could you use content_by_lua_block style?

balusch: done

spacewander (approved):

PR #376, opened 2022-01-09
bugfix: ngx.req.is_internal is only supported in http subsystem.

zhuizhuhaomeng:


lib/resty/core/misc.lua
```diff
@@ -165,6 +165,25 @@ if subsystem == "http" then
     end
     register_getter("headers_sent", headers_sent)
 
+
+    -- ngx.req.is_internal
+
```
cppcoffee, line 170: we can remove an extra new line.

cppcoffee (approved):

PR #377, opened 2022-01-10
tests: added a test for ngx.pipe to cover the pipe leaking.

zhuizhuhaomeng:


doujiang24 (approved):
better set a shorter read timeout for `proc` instead of the default read timeout value? to make the test case run faster.

PR #382, opened 2022-02-18
optimization: no need call ffi function to set resp header when value was empty string

ytlm:


piotrp:
I'm not sure it's a good idea, this would break removing of an already set header by passing an empty string

PR #399, opened 2022-06-30
optimize: unnecessary to judge whether r is nil or not

willmafh:

### DIFF
--- a/lib/resty/core/base.lua
+++ b/lib/resty/core/base.lua
@@ -248,7 +248,8 @@ do
     ok, exdata = pcall(require, "thread.exdata")
     if ok and exdata then
         function _M.get_request()
-            return exdata()
+            local r = exdata()
+            return r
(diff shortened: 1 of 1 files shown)


lib/resty/core/base.lua
```diff
@@ -248,10 +248,7 @@ do
     ok, exdata = pcall(require, "thread.exdata")
     if ok and exdata then
         function _M.get_request()
-            local r = exdata()
-            if r ~= nil then
```
spacewander, line 252: The guard here is used to convert `cdata:NULL` to a Lua `nil`.

willmafh: then we may use  `local r = exdata` for it to convert and then directly `return r`

zhuizhuhaomeng:
Need to convert NULL to nil explicitly

PR #434, opened 2023-07-27
new: upgrade nginx core to 1.25.1

swananan:
https://github.com/openresty/openresty/pull/920

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,6 +46,7 @@ env:
     - LUACHECK_VER=0.21.1
   matrix:
     - NGINX_VERSION=1.25.1 OPENSSL_VER=1.1.1u OPENSSL_PATCH_VER=1.1.1f
+    - NGINX_VERSION=1.21.4 OPENSSL_VER=1.1.1u OPENSSL_PATCH_VER=1.1.1f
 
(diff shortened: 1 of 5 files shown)


swananan:
nginx core 1.25.1 enables TLS 1.3 by default on the server side, but the ssl_session_fetch_by_lua_* and ssl_session_store_by_lua_* configurations only work with the TLS 1.2 protocol. There are two ways to ensure that cosocket SSL handshake uses TLS 1.2:

One way is to use lua_ssl_protocols tls1.2 configuration to ensure that the client sends the Client Hello handshake message carrying the supported_groups extension, which excluded TLS 1.3.

The other way is to use ssl_protocols tls1.2 to ensure that the server side chooses the TLS 1.2 version.

However, when we use the lua_ssl_protocols tls1.2 configuration, the OpenSSL library will not send the session ID in the Client Hello message by default. If the session ID doesn't exist, the session_fetch_cb won't be called, but the session_store_cb will work normally. Therefore, we need to change t/ctx.t and t/ssl-session-fetch.t test case check conditions.

zhuizhuhaomeng (approved):

PR #450, opened 2023-12-18
optimize: add Installation section in README.markdown, add warning message under the install target

### LINKED ISSUE #449: why the lua version number is commented out in Makefile
Can someone explain why the lua version number is commented out in Makefile?
I think it would be better if we keep it

> OPENRESTY_PREFIX=/usr/local/openresty
> 
> #LUA_VERSION := 5.1
> PREFIX ?=          /usr/local
> LUA_INCLUDE_DIR ?= $(PREFIX)/include
> LUA_LIB_DIR ?=     $(PREFIX)/lib/lua/$(LUA_VERSION)
> 	@@ -12,9 +13,11 @@ all: ;
>

lynch1981:
fix #449


zhuizhuhaomeng (approved):

zhuizhuhaomeng:
This is not a good idea.
It breaks the backward compatibility.
We should change the installation command in the readme.

lynch1981:
> This is not a good idea. It breaks the backward compatibility. We should change the installation command in the readme.
got it

lynch1981:
The Installation part is missing in README.markdown, please review the section I added @zhuizhuhaomeng

Makefile
```diff
@@ -1,9 +1,9 @@
 OPENRESTY_PREFIX=/usr/local/openresty
 
-LUA_VERSION := 5.1
+#LUA_VERSION := 5.1
 PREFIX ?=          /usr/local
 LUA_INCLUDE_DIR ?= $(PREFIX)/include
-LUA_LIB_DIR ?=     $(PREFIX)/share/lua/$(LUA_VERSION)
+LUA_LIB_DIR ?=     $(PREFIX)/lib/lua/$(LUA_VERSION)
```
zhuizhuhaomeng: Let's keep the original LUA_LIB_DIR and comment out the LUA_VERSION.
Instead we add an echo warning message under the install target.

zhuizhuhaomeng (changes requested):

Makefile
lynch1981: good idea

lynch1981:
Thanks to @zhuizhuhaomeng for the detailed description of the installation chapter

PR #466, opened 2024-05-07
bugfix: might get previous nextupdate value

ssdr:
For the ocsp response without nextUpdate value, it will get the previous nextUpdate value which is cached in the module-level lua ffi variable, this is not what we expected.

Maybe we should use function-level local variable instead?

### DIFF
--- a/lib/ngx/ocsp.lua
+++ b/lib/ngx/ocsp.lua
@@ -98,6 +98,7 @@ function _M.create_ocsp_request(certs, maxlen)
     return nil, ffi_str(errmsg[0])
 end
 
+local next_update_p = ffi_new("long[1]")
 
 function _M.validate_ocsp_response(resp, chain, max_errmsg_len)
     local errbuf_size = max_errmsg_len
@@ -108,8 +109,6 @@ function _M.validate_ocsp_response(resp, chain, max_errmsg_len)
 
     local sizep = get_size_ptr()
     sizep[0] = errbuf_size
-
-    local next_update_p = ffi_new("long[1]")
     
     local rc = C.ngx_http_lua_ffi_ssl_validate_ocsp_response(resp, #resp,
                                                              chain, #chain,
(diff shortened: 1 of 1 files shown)


ssdr:
https://github.com/openresty/lua-resty-core/commit/7a6521deb7a354fac9cb24cb477c144559ec3c06

Appla:
How about resetting next_update_p by `next_update_p[0] = 0` when `next_update != 0` before return next_update?

ssdr:
> How about resetting next_update_p by `next_update_p[0] = 0` when `next_update != 0` before return next_update?

yes, it seems better.

lib/ngx/ocsp.lua
```diff
@@ -110,7 +110,7 @@ function _M.validate_ocsp_response(resp, chain, max_errmsg_len)
 
     local sizep = get_size_ptr()
     sizep[0] = errbuf_size
-
+
```
zhuizhuhaomeng, line 113: should remove the extra space.

zhuizhuhaomeng (changes requested):

lib/ngx/ocsp.lua
ssdr: accept.

lib/ngx/ocsp.lua
```diff
@@ -119,7 +119,9 @@ function _M.validate_ocsp_response(resp, chain, max_errmsg_len)
     if rc == FFI_OK then
         local next_update = tonumber(next_update_p[0])
         if next_update == 0 then
-          next_update = nil
+            next_update = nil
+        else
+            next_update_p[0] = 0
```
zhuizhuhaomeng, line 124: Would you please put this line before line 114?

zhuizhuhaomeng (changes requested):

lib/ngx/ocsp.lua
ssdr: sure.

zhuizhuhaomeng (approved):

PR #513, opened 2025-10-23
feature: tcpsock sslhandshake alpn

bzp2010:

### DIFF
--- a/lib/resty/core/socket.lua
+++ b/lib/resty/core/socket.lua
@@ -246,7 +246,8 @@ local function sslhandshake(cosocket, reused_session, server_name, ssl_verify,
         local _bytes = {}
         for _, proto_str in ipairs(alpn)  do
             _bytes[#_bytes + 1] = string.len(proto_str)
-            for _, proto_byte in ipairs({ string.byte(proto_str, 1, #proto_str) }) do
+            for _, proto_byte in ipairs(
+                    { string.byte(proto_str, 1, #proto_str) }) do
                 _bytes[#_bytes + 1] = proto_byte
             end
         end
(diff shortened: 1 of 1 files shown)


lib/resty/core/socket.lua
```diff
@@ -241,6 +244,22 @@ local function sslhandshake(cosocket, reused_session, server_name, ssl_verify,
...
 
+    if alpn then
+        local _bytes = {}
+        for _, proto_str in ipairs(alpn)  do
+            _bytes[#_bytes + 1] = #proto_str
+            for _, proto_byte in ipairs(
+                    { str_byte(proto_str, 1, #proto_str) }) do
+                _bytes[#_bytes + 1] = proto_byte
+            end
+        end
```
theweakgod, line 255: The `str_byte` loop works, but we can simplify this quite a bit.

Since ALPN identifiers are a finite IANA-registered set, a pre-built wire-format lookup avoids runtime byte construction entirely for common protocols, with a fallback for anything else:

```lua
local ALPN_WIRE = {
    ["h2"]       = "\x02h2",
    ["http/1.1"] = "\x08http/1.1",
    ["http/1.0"] = "\x08http/1.0",
}

local bytes_array = {}
for _, p in ipairs(alpn) do
    bytes_array[#bytes_array+1] = ALPN_WIRE[p] or (string.char(#p) .. p)
end
local _bytes = table.concat(bytes_array)
alpn_str[0].data = ffi.new("unsigned char[?]", #_bytes, _bytes)
alpn_str[0].len  = #_bytes
```

`table.concat` produces the full wire string in one allocation

PR #523, opened 2026-03-16
feature: add server random and master key fetch api

theweakgod:
### What
 
Add two new FFI-based Lua APIs to the `ngx.ssl` module:
 
- `get_server_random([outlen])` — returns the server random (default 32 bytes) from the current SSL connection, wrapping OpenSSL's `SSL_get_server_random()`.
- `get_session_master_key([outlen])` — returns the master secret (default 48 bytes, i.e. `SSL_MAX_MASTER_KEY_LENGTH`) from the current SSL session, wrapping OpenSSL's `SSL_SESSION_get_master_key()`.
 
Both follow the same calling convention as the existing `get_client_random()`: pass `0` to get the length, pass `nil` or omit to get the value with the default size.
 
### Why
 
The `ngx.ssl` module already exposes `get_client_random()`, but server random and master key are missing. All three are required to produce an [NSS Key Log](https://www.ietf.org/archive/id/draft-thomson-tls-keylogfile-00.html) entry:
 
```
CLIENT_RANDOM <client_random_hex> <master_key_hex>
```
 
This file format is supported by Wireshark for decrypting TLS 1.2 traffic captures. Without this PR, users have to resort to `LD_PRELOAD` hacks, eBPF tools, or NGINX Plus (`ssl_key_log` / `proxy_ssl_key_log` directives) to obtain the same data.
 
### Usage
 
```lua
local ssl = require "ngx.ssl"
 
local cr = ssl.get_client_random()   -- 32 bytes raw
local sr = ssl.get_server_random()   -- 32 bytes raw
local mk = ssl.get_session_master_key() -- 48 bytes raw
```
 
Produce a Wireshark-compatible key log line:
 
```lua
local to_hex = require "resty.string".to_hex
local line = "CLIENT_RANDOM " .. to_hex(cr) .. " " .. to_hex(mk) .. "\n"
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -74,7 +74,7 @@ install:
   - git clone https://github.com/openresty/openresty.git ../openresty
   - git clone https://github.com/openresty/openresty-devel-utils.git
   - git clone https://github.com/simpl/ngx_devel_kit.git ../ndk-nginx-module
-  - git clone https://github.com/openresty/lua-nginx-module.git ../lua-nginx-module
+  - git clone https://github.com/theweakgod/lua-nginx-module.git ../lua-nginx-module # TODO: revert to openresty/lua-nginx-module once the companion lua-nginx-module PR is merged
   - git clone https://github.com/openresty/no-pool-nginx.git ../no-pool-nginx
   - git clone https://github.com/openresty/echo-nginx-module.git ../echo-nginx-module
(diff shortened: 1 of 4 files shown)


zhuizhuhaomeng (approved):

zhuizhuhaomeng:
This feature has been merged

PR #525, opened 2026-04-29
feature: add tcpsock:settrustedstore() for per-handshake trusted CAs

findns94:
Adds a Lua wrapper around the new
ngx_http_lua_ffi_socket_tcp_settrustedstore FFI in lua-nginx-module, exposed as tcpsock:settrustedstore(store). The store is an X509_STORE * cdata (e.g. from lua-resty-openssl) that overrides lua_ssl_trusted_certificate for the next sslhandshake() on this cosocket. The C side consumes the slot during the handshake, so the override does not leak across handshakes; passing nil clears it on both the lua and C sides so a previously-set store cannot dangle past a GC of the user's last reference.

This is needed for per-request mTLS upstreams where the trusted CA set is determined dynamically (per-tenant routing, dynamic CA discovery) and cannot be expressed via the static
lua_ssl_trusted_certificate directive.

The FFI symbol is looked up softly so loading lua-resty-core against an older lua-nginx-module that lacks the new symbol still works; the method is simply not attached to the cosocket metatable in that case. Only the http subsystem is wired up, matching the FFI surface in ngx_http_lua_module. Stream cosockets are unchanged.

Requires lua-nginx-module change https://github.com/openresty/lua-nginx-module/pull/2495.

### DIFF
--- a/lib/resty/core/socket.lua
+++ b/lib/resty/core/socket.lua
@@ -85,6 +85,10 @@ int
 ngx_http_lua_ffi_socket_tcp_get_ssl_ctx(ngx_http_request_t *r,
     ngx_http_lua_socket_tcp_upstream_t *u, void **pctx,
     char **errmsg);
+
+int
+ngx_http_lua_ffi_socket_tcp_settrustedstore(ngx_http_request_t *r,
+    ngx_http_lua_socket_tcp_upstream_t *u, void *store, char **errmsg);
 ]]
 
 ngx_lua_ffi_socket_tcp_getoption = C.ngx_http_lua_ffi_socket_tcp_getoption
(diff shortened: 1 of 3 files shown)


lib/resty/core/socket.lua
```diff
@@ -154,6 +158,7 @@ local FFI_NO_REQ_CTX = base.FFI_NO_REQ_CTX
 local SOCKET_CTX_INDEX          = 1
 local SOCKET_CLIENT_CERT_INDEX  = 6
 local SOCKET_CLIENT_PKEY_INDEX  = 7
+local SOCKET_TRUSTED_STORE_INDEX  = 8
```
zhuizhuhaomeng: SOCKET_BIND_INDEX = 8   /* only in upstream cosocket */

So you should use 10 for SOCKET_TRUSTED_STORE_INDEX

zhuizhuhaomeng (changes requested):

lib/resty/core/socket.lua
findns94: Thanks! I used `local SOCKET_TRUSTED_STORE_INDEX = 10` for now.

zhuizhuhaomeng (approved):

PR #533, opened 2026-06-03
bugfix: sslhandshake read error details on immediate FFI_ERROR

tzssangglass:
With the C side now returning FFI_ERROR instead of FFI_OK when u->error_ret is set after an immediate handshake, the early FFI_OK path is removed.  Added get_sslhandshake_result() in the FFI_ERROR branch so errmsg and openssl_error_code are populated when the error comes from the first call rather than from an async resume.

### DIFF
--- a/lib/resty/core/socket.lua
+++ b/lib/resty/core/socket.lua
@@ -424,52 +424,68 @@ local function sslhandshake(cosocket, reused_session, server_name, ssl_verify,
         error("no request ctx found", 2)
     end
 
+    local res
+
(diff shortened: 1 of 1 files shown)


tzssangglass:
Hi @zhuizhuhaomeng, this is the sister of https://github.com/openresty/lua-nginx-module/pull/2506, since https://github.com/openresty/lua-nginx-module/pull/2506 has been merged, this PR should also be merged.

zhuizhuhaomeng (approved):

PR #534, opened 2026-06-05
bugfix: remove sslhandshake result assert for pre-handshake errors.

tzssangglass:
When `tcpsock:sslhandshake()` fails before the actual SSL handshake starts, `ngx_http_lua_ffi_socket_tcp_get_sslhandshake_result()` may return `FFI_OK` while still providing the OpenSSL error details.

One example is a client private key setup failure. In that path, `rc` is `FFI_ERROR`, but `u->error_ret` is not set, so `ngx_http_lua_ffi_socket_tcp_get_sslhandshake_result` would return `FFI_OK`. The existing `assert(res == FFI_ERROR)` turns this expected error-reporting path into an assertion failure.

This patch removes the assertion so Lua can return the original SSL error to the caller.

### DIFF
--- a/lib/resty/core/socket.lua
+++ b/lib/resty/core/socket.lua
@@ -424,13 +424,9 @@ local function sslhandshake(cosocket, reused_session, server_name, ssl_verify,
         error("no request ctx found", 2)
     end
 
-    local res
-
     if rc == FFI_ERROR then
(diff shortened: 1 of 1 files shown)


tzssangglass:
As a fix for #533

tzssangglass:
ping @zhuizhuhaomeng

zhuizhuhaomeng (approved):

